PR #7, opened 2022-07-09
move TODO item to correct heading

nikomatsakis:
I think this TODO was misplaced, under integers instead of pointers. This was confusing to me for a while!


RalfJung:
Ah you are probably right, thanks!

PR #14, opened 2022-07-14
Experimental sum-like union representation.

alercah:
This is a counterexample, I hope, to @RalfJung's claim that the
representation model of union values needs to be much more elaborate to
take this approach to unions, e.g. in
rust-lang/unsafe-code-guidelines#73.

In this representation, a union is treated as a conceptual sum type,
except that it may actually have one value for each field due to the
lack of discriminant.

When decoding, we attempt to decode all fields. If at least one
succeeds, we consider it a success. This captures the idea that a union
value is valid iff it is valid for at least one field.

When encoding, for every field present, we encode it, and then we
combine those into a single encoding. There is a unique encoding that
will ensure that round-tripping a decode to an encode and back is valid.

The actual definition is quite simple, but it does require relaxing the
round-tripping invariant in the other direction: going to bytes and back
may add information to a union value, because it forces us to compute
which other fields' values can be deduced from the stored value.

I believe that if we get rid of the checks that a union value always has
at least one valid field, we actually recover something quite close to
the current model, except that the field Values are properly encoded in
the union. This might actually be an advantage, because it will make
code dealing with unions and tuples much more similar, for instance in
codegen, rather than being two very different beasts.

Incidentally extends tuple encode/decode to arbitrary sized tuples, in
order to factor out the common code.

Rendered:
[Values](https://github.com/alercah/minirust/blob/value-unions/lang/values.md),
[Types](https://github.com/alercah/minirust/blob/value-unions/lang/types.md)

### DIFF
--- a/lang/types.md
+++ b/lang/types.md
@@ -56,13 +56,8 @@ enum Type {
     Union {
         /// Fields *may* overlap. Fields only exist for field access place projections,
         /// they are irrelevant for the representation relation.
-        fields: Fields,
-        /// A union can be split into multiple "chunks", where only the data inside those chunks is
-        /// preserved, and data between chunks is lost (like padding in a struct).
-        /// This is necessary to model the behavior of some `repr(C)` unions, see
-        /// <https://github.com/rust-lang/unsafe-code-guidelines/issues/156> for details.
-        chunks: List<(Size, Size)>, // (offset, length) for each chunk.
-        /// The total size of the union, can indicate padding after the last chunk.
+        fields: Fields
+        /// The total size of the union, can indicate padding after all fields.
         size: Size,
     },
     Enum {
@@ -126,7 +121,7 @@ impl Type {
             Ref { pointee, .. } | Box { pointee } => pointee.inhabited,
             Tuple { fields, .. } => fields.iter().all(|type| type.inhabited()),
             Array { elem, count } => count == 0 || elem.inhabited(),
-            Union { .. } => true,
+            Union { .. } => fields.iter().any(|type| type.inhabited()),
             Enum { variants, .. } => variants.iter().any(|type| type.inhabited()),
         }
     }
--- a/lang/values.md
+++ b/lang/values.md
@@ -28,8 +28,8 @@ enum Value {
         idx: BigInt,
         data: Value,
     },
-    /// Unions are represented as "lists of chunks", where each chunk is just a raw list of bytes.
-    Union(List<List<AbstractByte>>),
+    /// A union value contains values for one or more of its fields.
+    Union(List<Option<Value>>),
 }
 ```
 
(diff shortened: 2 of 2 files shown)


RalfJung:
Thanks for the PR!

However, (without having looked at the code yet; not sure when I will have the time), I don't think we should do this. I am impressed it is *possible*, but even if it is, that doesn't mean it is a good idea. Unions should be simple. We shouldn't contort ourselves to copy some bad ideas other languages are stuck with.

Also, the following should be legal:
```rust
fn main() {
    union U {
        a: (u8, bool),
        b: (bool, u8),
    }
    let mut u = U { a: (25, false) };
    u.b.1 = 26;
    let _u2 = u; // typed copy
}
```
There's not even any unsafe code here. And a naive implementation of "unions as untagged sum types" would say the typed copy is UB. What does yours say?

alercah:
> Unions should be simple.

Maybe we just have different views of what is "simple" here. I personally think it's easier to reason about `List<Option<Value>>` than `List<List<AbstractByte>>`. `AbstractByte` is scary and subtle, and the less it comes up when trying to, say, review someone's code for UB, the better.

> And a naive implementation of "unions as untagged sum types" would say the typed copy is UB. What does yours say?

The current code says this is UB, and deliberately so. But the actual repr isn't as an untagged sum, it's as a product where every field is optional. The only reason it's UB is because of explicit assertions that at least one field is populated. If you remove those, then it would be well-defined. Just not with the semantics you expect.

RalfJung:
> Maybe we just have different views of what is "simple" here. I personally think it's easier to reason about `List<Option<Value>>` than `List<List<AbstractByte>>`. AbstractByte is scary and subtle, and the less it comes up when trying to, say, review someone's code for UB, the better.

For all intents and purposes, it's `List<AbstractByte>`. And that's a type everyone using, e.g., `alloc` has to be comfortable with. `AbstractByte` is an important part of the Rust AM, and it would be very strange to not have a way to represent it in the language itself.

But since the model you are proposing has strictly less UB than mine, then if you can justify a piece of code to be UB-free under that model, it will also be UB-free under mine! It is only some cases that you declare as having UB, that I think must be allowed.

> Just not with the semantics you expect.

Uh.oh. To be clear, the following code should be UB-free and the assertion always pass:
```rust
fn main() {
    union U {
        a: (u8, bool),
        b: (bool, u8),
    }
    let mut u = U { a: (25, false) };
    u.b.1 = 26;
    let u2 = u; // typed copy
    assert_eq!(unsafe { u2.a.0 }, 25);
    assert_eq!(unsafe { u2.b.1 }, 26);
}
```
If we make those writes safe, we can't just say "oh but the data will be lost on a copy", that's way too subtle. I certainly could not review code under those semantics.

Also, here's some other code that I assume you are ruling out, that I strongly think should work:
```rust
use std::mem::MaybeUninit;
use std::ptr::addr_of_mut;

fn main() { unsafe {
    let mut x: MaybeUninit<(u8, u8)> = MaybeUninit::uninit();
    // Field-by-field init, and we copy the entire thing to a different place
    // in the middle.
    addr_of_mut!((*x.as_mut_ptr()).0).write(15);
    let mut y = x;
    addr_of_mut!((*y.as_mut_ptr()).1).write(42);
    let pair = y.assume_init();
    assert_eq!(pair, (15, 42));
} }
```
If you don't do the move in the middle, then this must work, it is part of what `MaybeUninit` is for. And I don't think it is reasonable to have the move destroy the data here. Ergo this should also work.

RalfJung:
Also, looking at your semantics, I am surprised -- given how much you seem to like the C++ model, why do you allow code that is forbidden under C++? I expected the following to be UB:
```rust
#[repr(C)]
union U { f1: bool, f2: u8 }

let u = U { f1: false };
let int = unsafe { u.f2 }; // accessing union not via the active variant
```

To be clear, I think the above code should be allowed, but you cannot use "C++ is doing great" as justification and then have a model that is different from the one of C++.

PR #19, opened 2022-08-09
Fix more typos

sanxiyn:
I finished reading it once through. Will file more substantial comments separately.


RalfJung:
Looks good, thanks!

PR #22, opened 2022-08-17
Update README.md

hkBst:
Some small fixes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ All types except for mutable references are `Copy` (let's just imagine we implic
 
 We use `panic!` (and `unwrap` and slice indexing and similar standard Rust operations) to indicate conditions that should always hold; if execution ever panics, that is a bug in the specification.
 
-Our functions are generally pure; they can write to mutable reference but we can consider this to be implemented via explicit state passing.
+Our functions are generally pure; they can write to mutable references but we can consider this to be implemented via explicit state passing.
 When we do need other effects, we make them explicit in the return type.
 The next sections describe the effects used in the MiniRust interpreter.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -148,8 +148,8 @@ Right now, to my knowledge the Ferrocene Spec does not go into a lot of detail o
 
 ### What about Miri?
 
-MiniRust is in, at least conceptually, very similar to [Miri](https://github.com/rust-lang/miri).
-You can think of it as "idealized Miri": if Miri didn't have to care about working with all the rustc data structures that represent MIR and types, and didn't care about performance nor diagnostics, then it would be implemented like this specification.
+MiniRust is, at least conceptually, very similar to [Miri](https://github.com/rust-lang/miri).
+You can think of it as an "idealized Miri": if Miri didn't have to care about working with all the rustc data structures that represent MIR and types, and cared about neither performance nor diagnostics, then it would be implemented like this specification.
```
RalfJung, line 152: Is the "an" here required"? To me, the sentence "MiniRust is idealized Miri" sounds better than "MiniRust is an idealized Miri" -- an indefinite article sounds somewhat odd here IMO, "being a Miri" is not a thing.

RalfJung (commented):
Thanks for the PR! I have just one question about one of the changes.

hkBst:
Hi Ralf,

I think the article is not strictly required here, but I think the sentence
is better when it is included. To address your example: "being a Rust" is
also not really a thing, but I think my point was that a thing can be
idealized in many different ways. Nonsense examples: "Carbon is an
idealized Rust (because all its unicorns are pink), and Zig is also an
idealized Rust because it does not have trolls.".

But in case of doubt I recommend asking on the english stackexchange, so I
have done that.[1]

[1]:
https://english.stackexchange.com/questions/593761/should-you-include-an-article-when-using-idealized-with-a-unique-thing

RalfJung:
Hm yeah that's a pretty good example, thanks! From your SE post I think I would use "Python is idealized Ruby" but "Saturn is an idealized Mars", and I cannot explain why, it just 'feels' better... :shrug:

PR #25, opened 2022-09-03
Fix some typos

memoryleak47:


lang/values.md
```diff
@@ -321,9 +321,9 @@ impl PartialOrd for Value {
                 p1 <= p2,
             (Tuple(vals1), Tuple(vals2)) =>
                 vals1 <= vals2,
-            (Variant { idx: idx1, data: data1 }, Variant { idx: idx2, data: data2 }) =>
-                idx == idx1 && data1 <= data2,
-            (Union(chunks1), Union(chunks2)) => chunks1 <= chunks2,
+            (Value::Variant { idx: idx1, data: data1 }, Value::Variant { idx: idx2, data: data2 }) =>
```
RalfJung, line 324: Just above we use the other `Value` variants so this doesn't seem right.

Can we do `use Value::*;` at the top of the function instead?

RalfJung (commented):
Thanks! Looks good aside from one point.

lang/values.md
RalfJung: This should probably not be in this PR anyway, it conceptually seems to belong to your other PR.

memoryleak47: Yep, this shouldn't be in this PR. good catch!

memoryleak47:
I fixed the commit, and then rebased ontop of RalfJung/master.

RalfJung:
Thanks! Please squash this into a single commit.

memoryleak47:
done!

PR #27, opened 2022-09-12
Support arbitrary-sized tuples and arrays

### LINKED ISSUE #9: Support arbitrary-sized tuples
So far we only support pairs, not tuples with an arbitrary number of fields. That's just laziness though, and should not be hard to fix.

frank-king:
Fixes #9.

I think array types should be treated differently with tuples (or structs), because there are no paddings between two adjacent elements (at least for now, according to [the reference](https://doc.rust-lang.org/reference/type-layout.html#array-layout)). But I'm not sure whether it will change in the future, so I left it with a TODO.

### DIFF
--- a/lang/values.md
+++ b/lang/values.md
@@ -185,27 +185,25 @@ Note that types like `&!` have no valid value: when the pointee type is uninhabi
   (The latter would not even be possible with the current structure of MiniRust.)
   Also see [this discussion](https://github.com/rust-lang/unsafe-code-guidelines/issues/77).
 
-### Tuples (and arrays, structs, ...)
-
-For simplicity, we only define pairs for now.
+### Tuples (and structs, ...)
 
 ```rust
 impl Type {
-    fn decode(Tuple { fields: [field1, field2], size }: Self, bytes: List<AbstractByte>) -> Option<Value> {
+    fn decode(Tuple { fields, size }: Self, bytes: List<AbstractByte>) -> Option<Value> {
         if bytes.len() != size { throw!(); }
-        let (offset1, type1) = field1;
-        let val1 = type1.decode(bytes[offset1..][..type1.size()]);
-        let (offset2, type2) = field2;
-        let val2 = type2.decode(bytes[offset2..][..type2.size()]);
-        Value::Tuple([val1, val2])
+        Value::Tuple(
+            fields.iter().map(|(offset, ty)| {
+                ty.decode(bytes[offset..][..ty.size()])
+            }).collect()?,
+        )
     }
-    fn encode(Tuple { fields: [field1, field2], size }: Self, val: Value) -> List<AbstractByte> {
-        let Value::Tuple([val1, val2]) = val else { panic!() };
-        let mut bytes = [AbstractByte::Uninit; size];
-        let (offset1, type1) = field1;
-        bytes[offset1..][..type1.size()] = type1.encode(val1);
-        let (offset2, type2) = field2;
-        bytes[offset2..][..type2.size()] = type2.encode(val2);
+    fn encode(Tuple { fields, size }: Self, val: Value) -> List<AbstractByte> {
+        let Value::Tuple(values) = val else { panic!() };
+        assert_eq!(chunk_data.len(), chunks.len());
+        let mut bytes = list![AbstractByte::Uninit; size];
+        for ((offset, ty), value) in fields.iter().zip(values) {
+            bytes[offset..][..ty.size()].copy_from_slice(ty.encode(value));
+        }
         bytes
     }
 }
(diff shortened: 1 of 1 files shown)


RalfJung:
That is not a good argument - tuples can have padding but they don't require padding. This makes arrays (of a fixed length) a special case of tuples.

So I am not convinced we should split arrays and tuples. Maybe this helps for place expressions (indexing vs field projection) but it does not help for (monomorphic) type representations.

frank-king:
I didn't mean to treat arrays and tuples differently when talking about the `Type::encode` and `Type::decode` functions. I mean syntactically in MiniRust, `Type::Array` and `Type::Tuple` are different enum variants, so we should write two individual `match` arms (or fallible patterns) for them.

I just added the `Type::Array` pattern because it was missing in the original text.

To deal with `Type::Tuple` and `Type::Array` in the same match arm, we need a helper function `.fields()` which returns an iterator of offsets and field types `(Size, Type)`, and with this helper function, `Type::Tuple` and `Type::Array` can be easily processed in a single `decode` or `encode` function.

RalfJung:
Sorry, I misunderstood (and on mobile the diff is hard to read). Will take a closer look after my vacation (in about a week).

lang/values.md
```diff
@@ -185,27 +185,24 @@ Note that types like `&!` have no valid value: when the pointee type is uninhabi
...
-    fn decode(Tuple { fields: [field1, field2], size }: Self, bytes: List<AbstractByte>) -> Option<Value> {
+    fn decode(Tuple { fields, size }: Self, bytes: List<AbstractByte>) -> Option<Value> {
         if bytes.len() != size { throw!(); }
-        let (offset1, type1) = field1;
-        let val1 = type1.decode(bytes[offset1..][..type1.size()]);
-        let (offset2, type2) = field2;
-        let val2 = type2.decode(bytes[offset2..][..type2.size()]);
-        Value::Tuple([val1, val2])
+        Value::Tuple(
+            fields.into_iter().map(|(offset, ty)| {
```
RalfJung, line 195:
```suggestion
            fields.iter().map(|(offset, ty)| {
```

lang/values.md
```diff
@@ -185,27 +185,24 @@ Note that types like `&!` have no valid value: when the pointee type is uninhabi
...
-        let Value::Tuple([val1, val2]) = val else { panic!() };
-        let mut bytes = [AbstractByte::Uninit; size];
-        let (offset1, type1) = field1;
-        bytes[offset1..][..type1.size()] = type1.encode(val1);
-        let (offset2, type2) = field2;
-        bytes[offset2..][..type2.size()] = type2.encode(val2);
+    fn encode(Tuple { fields, size }: Self, val: Value) -> List<AbstractByte> {
+        let Value::Tuple(values) = val else { panic!() };
+        let mut bytes = list![AbstractByte::Uninit; size];
+        for ((offset, ty), value) in fields.into_iter().zip(values) {
```
RalfJung, line 203:
```suggestion
        for ((offset, ty), value) in fields.iter().zip(values) {
```
also please add an assertion ensuring they have the same length

lang/values.md
```diff
@@ -215,6 +212,32 @@ Note in particular that `decode` ignores the bytes which are before, between, or
...
+        Value::Tuple(
+            bytes.chunks(elem.size())
+                .map(|elem_bytes| elem.decode(elem_bytes))
+                .try_collect()?,
+        )
+    }
+    fn encode(Array { elem, count }: Self, val: Value) -> List<AbstractByte> {
+        let Value::Tuple(values) = val else { panic!() };
+        assert_eq!(values.len(), count);
+        values.into_iter().flat_map(|value| {
```
RalfJung, line 230:
```suggestion
        values.iter().flat_map(|value| {
```

lang/values.md
```diff
@@ -215,6 +212,32 @@ Note in particular that `decode` ignores the bytes which are before, between, or
...
+            let bytes = elem.encode(value);
+            assert_eq!(bytes.len(), elem.size());
+            bytes
+        }).collect()
+    }
+}
+```
+
+- TODO: Should we consider paddings between two adjacent elements in arrays?
```
RalfJung, line 239: I don't think we need this remark, arrays are pretty tied down in how they are laid out. There is no room for padding here.

lang/values.md
```diff
@@ -185,27 +185,24 @@ Note that types like `&!` have no valid value: when the pointee type is uninhabi
...
         if bytes.len() != size { throw!(); }
-        let (offset1, type1) = field1;
-        let val1 = type1.decode(bytes[offset1..][..type1.size()]);
-        let (offset2, type2) = field2;
-        let val2 = type2.decode(bytes[offset2..][..type2.size()]);
-        Value::Tuple([val1, val2])
+        Value::Tuple(
+            fields.into_iter().map(|(offset, ty)| {
+                ty.decode(bytes[offset..][..ty.size()])
+            }).try_collect()?,
```
RalfJung, line 197: Can't `collect` also do this?

RalfJung (commented):
Looks good, apart from a few small nits! The code duplication is small enough that a helper function is probably not worth it.

lang/values.md
RalfJung: Yeah for `Result` `collect` also works, so given that `try_collect` is unstable IMO we should use `collect` for now.

RalfJung:
Thanks, that looks very nice!

PR #28, opened 2022-09-12
Minor fix: type check of unions

frank-king:
Union fields must fit into one of its chunks.

### DIFF
--- a/lang/well-formed.md
+++ b/lang/well-formed.md
@@ -67,8 +67,14 @@ impl Type {
                 for (offset, type) in fields {
                     type.check()?;
                     ensure(size >= offset.checked_add(type.size())?)?;
+
+                    // And it must fit into one of the chunks.
+                    ensure(chunks.into_iter().any(|(chunk_offset, chunk_size)| {
+                        chunk_offset <= offset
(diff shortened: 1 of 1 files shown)


RalfJung:
Thanks! I am on vacation; will take a closer look when I am back.

RalfJung:
Makes sense, thanks! I am not sure if we want to insist on chunks being sorted but indeed that is currently what the check does.

lang/well-formed.md
```diff
@@ -67,8 +67,14 @@ impl Type {
...
                     ensure(size >= offset.checked_add(type.size())?)?;
+
+                    // And it must fit into one of the chunks.
+                    ensure(chunks.into_iter().any(|(chunk_offset, chunk_size)| {
+                        chunk_offset <= offset
+                            && offset + type.size() <= chunk_offset + chunk_size
+                    }))?;
                 }
-                // The chunks must be disjoint.
+                // The chunks must be sorted in their offsets and disjoint.
```
RalfJung, line 77:
```suggestion
                // The chunks must be sorted in their offsets and disjoint.
                // FIXME: should we relax this and allow arbitrary chunk order?
```

Issue #1, opened 2022-05-25
Support unsized types


RalfJung:
One of the things I am least sure about is how to best add support for unsized types to MiniRust. Clearly `Pointer` should optionally carry some `metadata`, but what do we do about all the places in the spec where we make use of the size (or alignment) of a type? There will be quite a few decisions to be made there, I think. Unsized arguments will also be interesting; unsized locals in general I think are out-of-scope for now due to [their various problems](https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/Status.20of.20unsized.20locals).

RalfJung:
This is mostly implemented by now, except for unsized structs which are tracked at https://github.com/minirust/minirust/issues/258.

PR #29, opened 2022-09-13
Support slice type and pointer-unsize cast

frank-king:
This PR implements part of #1.

It adds optional metadata in the pointer value and the place representation.

Slices' subslice is not supported yet. In MIRI, it seems to be a function call of `Index::<impl RangeBounds<usize>>::index`. I'm not sure how should mini-rust deal with it.

TODO: need to check the files manually one by one.
- [ ] `prelude.md`
- [ ] `types.md`
- [ ] `values.md`
- [ ] `well-formed.md`
- [ ] `syntax.md`
- [ ] `machine.md`
- [ ] `step.md`
- [ ] `operator.md`

### DIFF
--- a/lang/operator.md
+++ b/lang/operator.md
@@ -50,6 +50,20 @@ impl Machine {
 }
 ```
 
+### Pointer unsize cast
+
+```rust
+impl Machine {
+    fn eval_un_op(&mut self, PtrUnsize(pointee): UnOp, operand: Value) -> NdResult<Value> {
+        let Value::Ptr(ptr) = operand else { panic!("non-pointer input to ptr-unsize cast") };
+        assert!(ptr.meta.is_none(), "fat-pointer input to ptr-unsize cast") };
+        let Type::Array { elem, count } = pointee else { panic!("non-array pointee type to ptr-unsize cast") };
+
+        Value::Ptr(PointerRepr { meta: Some(count), ..ptr })
+    }
+}
+```
+
 ## Binary operators
 
 ```rust
--- a/lang/step.md
+++ b/lang/step.md
@@ -94,7 +94,7 @@ impl Machine {
         if !check_safe_ptr(p, Layout { align, ..ptype.layout() }) {
             throw_ub!("creating reference to invalid (null/unaligned/uninhabited) place");
         }
-        Value::Ptr(p)
+        Value::Ptr(PointerRepr { ptr: Pointer(p, None), meta: None })
     }
 
 }
@@ -121,11 +121,14 @@ impl Machine {
 ## Place Expressions
 
 Place expressions evaluate to places.
-For now, that is just a pointer (but this might have to change).
+It is a pointer of its address, and an optional meta of pointer size.
+For the slice type, the meta must be present and contains the length of it.
+It may be extended to support other unsized types (such as trait objects) in the future.
+
 Place evaluation ensures that this pointer is always dereferenceable (for the type of the place expression).
 
 ```rust
-type Place = Pointer;
+type Place = PointerRepr;
 
 impl Machine {
     fn eval_place(&mut self, place: PlaceExpr) -> NdResult<Place>;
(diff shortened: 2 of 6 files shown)


RalfJung:
Thanks for the PR! Unsized types will need some careful design I think, so it probably makes sense to discuss this a bit before committing to and implementing one design.

frank-king:
# Types
## Size

To support unsized types, I make `Type`'s `.size()` to return an `Option<Size>`, and this can affect the value's encoding/decoding.

For most sized types or compound types where only sized types are allowed in their fields (such as unions, tuples, enums, etc.), it is trivial to replace `.size()` with `.size().unwrap()`.

## Layout
References and boxes (or even the raw pointers) should be carefully taken care of, because their pointee types may be unsized. I changed the `size` in `Layout` to be optional because the only difference between a sized type and an unsized type I think, is that the sized type has a known `size` but the unsized type doesn't. They both have an `align` and an `inhabited` property (I'm not sure of it).

# Values
## Metadata
When encoding/decoding the unsized types, we need extra information to determine how many bytes should we decode or how many values should we encode. In the surficial Rust language, fat pointers carry metadata containing this information. But in MiniRust, there is no metadata involved yet in the `encode`/`decode` method, so I think here it might need further design,

For slice types, it is possible to calculate their lengths from the list of abstract bytes or the list of values, since the length of bytes must be a multiple of the element type's size, and the number of encoded values must equal the length of the slice. But for trait objects, it is not possible to do that.

## PointerRepr (The May-be-Fat Pointer)
Fortunately, because `encode` and `decode` are only used in the typed load or store, it is possible to pass in the metadata from the pointer to read or write. I simply added a `PointerRepr` type including a pointer address and the optional metadata, where the metadata needs further design if we would like to support more unsized types such as trait objects.

For pointer families, there is a minor change in `encode`/`decode`: we need to encode or decode the metadata if it is a fat pointer.

BTW, `PointerRepr` may not be a good name.

## The "More Defined" Partial Order
For thin pointers, it remains the same case as the previous `Pointer`.

For fat pointers, I don't think there is a partial order of the "more defined" relationship because there are no choices of changing the value of metadata when doing an encode and then decode. Thus, I defined its partial order as, a may-be-fat pointer is more defined than another if and only if its address is more defined than the other, and they have the same metadata.

# Places
I cannot tell the subtle differences between `Place` and `Pointer` (I just think `Place` of a semantical correspondence of `Pointer`), so I simply changed the type alias of `Place` from the thin `Pointer` to the may-be-fat `PointerRepr`, to enable places of unsized types.

# New Operators for Slice Types
Finally, I added two operators which are commonly used for the slice types.

## `Len`
Similar to MIRI, `len` supports both array types and slice types:
- for the array type, the length comes directly from its type's definition;
- for the slice type, the length comes from the metadata of the place value.

## `PtrUnsize`
This converts a sized pointer to an unsized pointer. I put it in the `UnaryOp` but I'm not sure if it is the right place.

---

Questions that I'm not sure about:
- [ ] Should we make the size in `Layout` an `Option<Size>` or use other ways to represent fat-pointer types?
- [ ] What type should the metadata in fat pointers be?
- [ ] How should we support other unsized types such as trait objects? (We may need to define function pointers though)
- [ ] What name should we pick up for the may-be-fat pointer? (Currently `PointerRepr`)
- [ ] What type should the `Place` be?
- [ ] Are we going to add more operators for unsized types?

RalfJung:
@frank-king are you on Zulip? For design decisions I think a hood strategy is to have a design document on https://hackmd.io/ and then discuss [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/136281-t-lang.2Fwg-unsafe-code-guidelines).

frank-king:
Yes, my ID on Zulip is "Frank King".

RalfJung:
Great, I [created a topic](https://rust-lang.zulipchat.com/#narrow/stream/136281-t-lang.2Fwg-unsafe-code-guidelines/topic/MiniRust.20support.20for.20unsized.20types) :)

RalfJung:
Thanks for the proposal! However, I don't think a MiniRust PR (with concrete code suggestions) is a good place for such a design discussion, so I am going to close it. Please either open a new issue with the design draft, or put it into a [hackmd](https://hackmd.io/) so that we can discuss it on Zulip.

PR #30, opened 2022-09-25
Support arbitrary-sized integers

### LINKED ISSUE #8: Support arbitrary-sized integers
So far we only support 16-bit integers. That's just laziness, I wanted to ensure that we handle endianess but not write the code for arbitrary-width handling.

frank-king:
Fixes #8.

In `decode`, I use a big `match` instead of `bytes.iter().map(...).collect()` because `Endianess::decode` accepts a fixed-sized array `[u8; N]` instead of a slice `&[u8]`. It looks kind of verbose. I'm not sure if we should add a helper macro here.

EDIT: `encode` has a similar problem.

### DIFF
--- a/lang/prelude.md
+++ b/lang/prelude.md
@@ -14,11 +14,11 @@ type AbstractByte = Memory::AbstractByte;
 // The endianess, which defines how integers are encoded and decoded.
 trait Endianess {
     /// If `signed == Signed`, the data is interpreted as two's complement.
-    fn decode<const N: usize>(self, signed: Signedness, bytes: [u8; N]) -> BigInt;
+    fn decode(self, signed: Signedness, bytes: List<u8>) -> BigInt;
 
-    /// This can fail (return `None`) if the `int` does not fit into `N` bytes,
+    /// This can fail (return `None`) if the `int` does not fit into `size` bytes,
     /// or if it is negative and `signed == Unsigned`.
-    fn encode<const N: usize>(self, signed: Signedness, int: BigInt) -> Option<[u8; N]>;
+    fn encode(self, signed: Signedness, size: Size, int: BigInt) -> Option<List<u8>>;
 }
 const ENDIANESS: impl Endianess;
 
--- a/lang/values.md
+++ b/lang/values.md
@@ -91,18 +91,23 @@ Note, in particular, that `bool` just entirely ignored provenance; we discuss th
 
 ### Integers
 
-For now we only define `u16` and `i16`.
-
 ```rust
 impl Type {
-    fn decode(Type::Int(IntType { signed, size: Size::from_bits(16) }): Self, bytes: List<AbstractByte>) -> Option<Value> {
-        let [AbstractByte::Init(b0, _), AbstractByte::Init(b1, _)] = *bytes else { throw!() };
-        Value::Int(ENDIANESS.decode(signed, [b0, b1]))
+    fn decode(Type::Int(IntType { signed, size }): Self, bytes: List<AbstractByte>) -> Option<Value> {
+        if bytes.len() != size.bytes() {
+            throw!();
+        }
+        // Fails if any byte is `Uninit`.
+        let bytes_data = bytes.iter().map(|b| b.data()).collect()?;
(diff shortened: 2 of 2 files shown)


frank-king:
Would it be better to let `Endianess::decode` accept `List<u8>` and let `Endianess::encode` return `List<u8>`?

RalfJung:
Ah, good point.

We could make `decode` take a `List` and `encode` take a `Size`, maybe?

frank-king:
Sounds good. That would be neater.

lang/prelude.md
```diff
@@ -11,11 +11,11 @@ use Memory::{Provenance, Pointer, AbstractByte};
...
 trait Endianess {
     /// If `signed == Signed`, the data is interpreted as two's complement.
-    fn decode<const N: usize>(self, signed: Signedness, bytes: [u8; N]) -> BigInt;
+    fn decode(self, signed: Signedness, bytes: List<u8>) -> BigInt;
 
-    /// This can fail (return `None`) if the `int` does not fit into `N` bytes,
+    /// This can fail (return `None`) if the `int` does not fit into `size` bytes,
     /// or if it is negative and `signed == Unsigned`.
-    fn encode<const N: usize>(self, signed: Signedness, int: BigInt) -> Option<[u8; N]>;
+    fn encode(self, signed: Signedness, size: Size, int: BigInt) -> Option<List<u8>>;
```
RalfJung, line 18: This will also need adjustments in all existing users of these functions (in particular, pointer encode/decode).

frank-king: Oh, I forgot that.

lang/values.md
```diff
@@ -91,18 +91,22 @@ Note, in particular, that `bool` just entirely ignored provenance; we discuss th
...
 ```rust
 impl Type {
-    fn decode(Type::Int(IntType { signed, size: Size::from_bits(16) }): Self, bytes: List<AbstractByte>) -> Option<Value> {
-        let [AbstractByte::Init(b0, _), AbstractByte::Init(b1, _)] = *bytes else { throw!() };
-        Value::Int(ENDIANESS.decode(signed, [b0, b1]))
+    fn decode(Type::Int(IntType { signed, size }): Self, bytes: List<AbstractByte>) -> Option<Value> {
+        if bytes.len() != size.bytes() {
+            throw!();
+        }
+        let bytes_data = bytes.iter().map(|b| b.data()).collect()?;
```
RalfJung, line 100:
```suggestion
        // Fails if any byte is `Uninit`.
        let bytes_data = bytes.iter().map(|b| b.data()).collect()?;
```

lang/well-formed.md
```diff
@@ -19,6 +19,8 @@ fn ensure(b: bool) -> Option<()> {
 impl IntType {
     fn check(self) -> Option<()> {
         ensure(self.size.bytes().is_power_of_two())?;
+        // We only support int types that are no larger than 128 bits.
```
RalfJung, line 22: This is what rustc currently does, but is there anything that would break in Minirust if things would become bigger?

For now I am inclined to permit arbitrary integer sizes here; that doesn't mean the rustc frontend has to use them all.

frank-king: Considering only the integer types, there's no problem with supporting arbitrary-sized integers. But if we support arbitrary-sized integers, should we support arbitrary-sized floats? (Though MiniRust doesn't support floats yet)

RalfJung: Rust doesn't have `f16` or `f128` so I think the answer to your question is 'no'.

But anyway MiniRust should treat floats very abstractly I think, via some softloat type provided as an external library. floats are pretty low -priority right now though...

frank-king: Indeed, floats are much more complicated than integers. It lets me think of the `long double` in C++, which might be `f80` in Rust (the size is even not the power of two).

If we treat floats completely differently from integers (i.e., not by encoding/decoding from raw bytes and reinterpreting the bits), I agree not to limit the size of integer types. It would be a superset of the integer types that rustc supports.

RalfJung:
Looks great, thanks!

PR #33, opened 2022-11-02
Make Machine and Value generic over the Memory

memoryleak47:


lang/syntax.md
```diff
@@ -93,10 +93,13 @@ enum Terminator {
 And finally, the syntax of expressions:
 
 ```rust
+/// Constants are Values, but are not allowed to be Value::Ptr | Value::Union.
```
RalfJung, line 96: The comment is not accurate -- they can still be a pointer with `None` provenance.

lang/step.md
```diff
@@ -39,19 +39,35 @@ impl Machine {
...
+impl<M: Memory> Machine<M> {
+    fn eval_value(&mut self, val: ValueExpr) -> NdResult<Value<M>>;
 }
 ```
 
 ### Constants
 
-Constants are trivial, as one would hope.
-
 ```rust
```
RalfJung, line 49:
```suggestion
Convert from constant to value -- converts provenance from `!` to `M::Provenance`.

```rust
```

lang/well-formed.md
```diff
@@ -44,15 +44,15 @@ impl PtrType {
...
+    fn check<Info: MemoryInfo>(self) -> Option<()> {
         use Type::*;
         match self {
             Int(int_type) => {
                 int_type.check()?;
             }
             Bool => (),
             Pointer(ptr_type) {
-                int_type.check()?;
+                ptr_type.check()?;
```
RalfJung, line 55: *oops* good catch

lang/step.md
```diff
@@ -39,19 +39,35 @@ impl Machine {
...
+                let args = args.into_iter()
+                    .map(|arg| self.eval_constant(arg))
+                    .collect();
+                Value::Tuple(args)
+            },
+            Constant::Variant { idx, data } => {
+                let data = self.eval_constant(data);
+                Value::Variant { idx, data }
+            },
+            Constant::Ptr(..) | Constant::Union(..) => panic!("invalid constant encountered!"),
```
RalfJung, line 65: Please add a comment in well-formed noting that if we allow these consts, we have to also support them here.

mem/interface.md
```diff
@@ -30,7 +30,7 @@ As far as the interface is concerned, this is some opaque extra data that we car
 The unit of communication between the memory model and the rest of the program is a *byte*.
 To distinguish our MiniRust bytes from `u8`, we will call them "abstract bytes".
 Abstract bytes differ from `u8` to support representing uninitialized Memory and to support maintaining pointer provenance when pointers are stored in memory.
-We define the `AbstractByte` type as follows, where `Provenance` will later be instantiated with the `Memory::Provenance` associated type.
+We define the `AbstractByte` type as follows, where `Provenance` will later be instantiated with the `M::Provenance` associated type.
```
RalfJung, line 33: The trait is still called `Memory`; `M` is not in scope here.

mem/interface.md
```diff
@@ -78,11 +78,16 @@ struct Pointer<Provenance> {
     provenance: Option<Provenance>,
 }
 
+trait MemoryInfo {
```
RalfJung, line 81: Please add a doc comment explaining what this trait is for.

prelude.md
```diff
@@ -57,6 +54,20 @@ enum Mutability {
 }
 pub use Mutability::*;
 
+// The endianness, which defines how integers are encoded and decoded.
```
RalfJung, line 57:
```suggestion
/// The endianness, which defines how integers and pointers are encoded and decoded.
```

lang/syntax.md
RalfJung:
```suggestion
/// Constants are Values, but cannot have provenance.
/// (We actually entirely disallow Value::Ptr | Value::Union in the well-formedness checker.)
```

lang/values.md
```diff
@@ -122,38 +122,38 @@ This is required to achieve a "monotonicity" with respect to provenance (as disc
 
 ### Raw pointers
 
-Decoding pointers is a bit inconvenient since we do not know `PTR_SIZE`.
+Decoding pointers is a bit inconvenient since we do not know `M::PTR_SIZE`.
```
RalfJung, line 126:
```suggestion
```
Outdated comment, now that we support arbitrarily sized integers.

lang/values.md
```diff
@@ -165,21 +165,21 @@ impl Type {
...
     // References (and `Box`) need to be non-null, aligned, and not point to an uninhabited type.
     // (Think: uninhabited types have impossible alignment.)
     ptr.addr != 0 && ptr.addr % pointee.align.bytes() == 0 && pointee.inhabited
 }
 
 impl Type {
-    fn decode(Type::Ref { pointee, .. } | Type::Box { pointee }: Self, bytes: List<AbstractByte>) -> Option<Value> {
-        let ptr = decode_ptr(bytes)?;
-        if !check_safe_ptr(ptr, pointee) { throw!(); }
+    fn decode<M: Memory>(Type::Ref { pointee, .. } | Type::Box { pointee }: Self, bytes: List<M::AbstractByte>) -> Option<Value<M>> {
```
RalfJung, line 175: Oh I forgot to update this for the unified pointer type. Will do that after your MR.
```suggestion
    // FIXME: this is not yet updated to the unified pointer type.
    fn decode<M: Memory>(Type::Ref { pointee, .. } | Type::Box { pointee }: Self, bytes: List<M::AbstractByte>) -> Option<Value<M>> {
```

lang/syntax.md
```diff
@@ -90,12 +90,24 @@ enum Terminator {
 }
 ```
 
-And finally, the syntax of expressions:
+In order to define expressions, we first need to define constants.
```
RalfJung, line 93:
```suggestion
We also need to define constants (a strict subset of `Value`).
```

mem/interface.md
```diff
@@ -78,19 +78,18 @@ struct Pointer<Provenance> {
...
 /// We also let read operations potentially mutate memory (they actually can
 /// change the current state in concurrent memory models and in Stacked Borrows).
-trait Memory: MemoryInfo {
+trait Memory {
     /// The type of pointer provenance.
     type Provenance: Eq;
 
+    /// The size of a pointer.
+    const PTR_SIZE: Size;
+    const ENDIANNESS: Endianness;
```
RalfJung, line 91:
```suggestion

    /// The endianess used for encoding multi-byte integer values (and pointers).
    const ENDIANNESS: Endianness;
```

PR #34, opened 2022-11-04
add Constant::Variant

memoryleak47:


memoryleak47:
I replaced `variants[idx]` with `variants.get(idx)?`, hence the force-push!

PR #35, opened 2022-11-04
define `BasicMemory::{ENDIANNESS, PTR_SIZE}`

memoryleak47:


mem/basic.md
```diff
@@ -54,6 +46,15 @@ struct BasicMemory {
 }
 ```
 
+The model represents a 64-bit little-endian machine.
+
+```rust
+impl Memory for BasicMemory {
+    const PTR_SIZE: Size = Size::from_bytes(8).unwrap();
```
RalfJung, line 53:
```suggestion
    const PTR_SIZE: Size = Size::from_bits(64).unwrap();
```
seems to better match the text. ;)

PR #38, opened 2022-11-07
Small fixes

memoryleak47:


mem/basic.md
```diff
@@ -152,7 +152,7 @@ impl BasicMemory {
     /// Check if the given pointer is dereferenceable for an access of the given
     /// length and alignment. For dereferenceable, return the allocation ID and
     /// offset; this can be missing for invalid pointers and accesses of size 0.
-    fn check_ptr(&self, ptr: Pointer<Self::Provenance>, len: Size, align: Align) -> Result<Option<(AllocId, Size)>> {
+    fn check_ptr(&self, ptr: Pointer<<Self as Memory>::Provenance>, len: Size, align: Align) -> Result<Option<(AllocId, Size)>> {
```
RalfJung, line 155: Why is this needed?

memoryleak47: Good question, I think it should be unambiguous, but apparently Rust doesn't detect it.

Minimal example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=49b7b4bf69efa8703eeeaa987b4b333b

RalfJung: Oh strange, this only happens for associated types of concrete types.

I think we might want to refactor this later but for now it's fine.

RalfJung, line 155:
```suggestion
    fn check_ptr(&self, ptr: Pointer<AllocId>, len: Size, align: Align) -> Result<Option<(AllocId, Size)>> {
```

memoryleak47: Note that this change is only necessary once though, as in all other occurences we have `<M: Memory>`, and thus `M::Provenance` is really unambiguous

RalfJung: Yeah it's only an issue on a concrete type. Reported as https://github.com/rust-lang/rust/issues/104119.

PR #40, opened 2022-11-07
replace `PartialOrd` with `DefinedRelation`

memoryleak47:


lang/values.md
```diff
@@ -271,10 +271,18 @@ Note that none of the definitions in this section are needed to define the seman
 They only serve as internal consistency requirements of the semantics.
 It would be a specification bug if the representation relations defined above violated these properties.
 
+```rust
+trait DefinedRelation {
+  /// returns whether `self` is less or as defined as `other`
+  fn le_defined(&self, other: &Self) -> bool;
```
RalfJung, line 277: We've mostly been using 4-space indent I think, let's stick to that.

memoryleak47: whoops!

PR #42, opened 2022-11-07
ignore prelude.md code. It will be replaced by the baselib implementation

memoryleak47:


prelude.md
```diff
@@ -2,7 +2,7 @@
 
 Across all files in this repository, we assume some definitions to always be in scope.
 
-```rust
+```rust,ignore
```
RalfJung, line 5: Do we really want to ignore all of them?
`Mutability` and `TerminationInfo` seem like things we can define locally, no?

memoryleak47: They can be defined locally.
I just think it's simpler to mirror the whole module, and not add an exception for those two cases.

What do you think?

RalfJung: I'm beginning to feel like maybe we actually want to locally define most of these things?

Basically the deciding factor should be: is this a type we want to independently implement for the Rust and Coq translations, or would we prefer a single shared definition for both? And long-term we probably want to share everything except for `BigInt` and `NdResult`.

RalfJung: That makes the prelude rather big so it'd probably make sense to have a directory with files `prelude/size.md`, `prelude/align.md`, `prelude/endianess.md`, ...

memoryleak47: sounds good!
I'll open a new PR for the new prelude.

Also:
> (If we have the attributes in specr:: it might make sense to als rename baselib:: to specr::?)

Note that currently `baselib` is not user-facing, as `type BigInt;` is used for opaque types.
Do we want to change that?

RalfJung: Ah right. Well now that we have the attributes it makes sense to use the `specr` namespace also for other things, I'd say?

memoryleak47: Yep, why not.
I kind of like the magic of `type _;`, but I don't mind changing it.

RalfJung: Sure, let's keep it as is for now. :)

PR #43, opened 2022-11-08
New prelude

memoryleak47:
This reworks how baselib (now called specr) and the prelude work together.

Remarks:
- `std::result::Result` is mandatory in `Align` because of `type Result<T> = ...` from the prelude/main.md, or do we want to return `TerminationInfo::Ub` incase of errors?
- `Size::from_bytes` has been implemented as infallible, as we intended for the future.


prelude/align.md
```diff
@@ -0,0 +1,47 @@
+This is basically a copy of the `Align` type in the Rust compiler.
+See [Align](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_target/abi/struct.Align.html).
+
+`Align` is always a power of two.
+
+```rust
+pub struct Align { pow2: BigInt }
```
RalfJung, line 7: I don't think we want to use the `pow2` representation that the Rust compiler uses. This will be just unnecessarily painful in Coq.

prelude/align.md
```diff
@@ -0,0 +1,47 @@
...
+
+impl Align {
+    pub const ONE: Align = Align { pow2: BigInt::from(0) };
+
+    pub fn from_bits(bits: impl Into<BigInt>) -> std::result::Result<Align, String> {
```
RalfJung, line 12: I don't think we want from_bits and bits on `Align`.

prelude/align.md
```diff
@@ -0,0 +1,47 @@
...
+        let bytes = Size::from_bits(bits).bytes();
+        Align::from_bytes(bytes)
+    }
+
+    pub fn from_bytes(align: impl Into<BigInt>) -> std::result::Result<Align, String> {
```
RalfJung, line 17: Might be nicer to make this infallible by always using the next power of two. (`BigInt` should have a `next_power_of_two` method for this.)

lang/prelude.md
```diff
@@ -13,5 +13,5 @@ use crate::mem::{Memory, AbstractByte, Pointer, IntPtrCast};
 // to use and whether to do adjustments like sign extension.
 // `Call` does not even check that caller and callee agree on the size
 // (and indeed for register passing, mismatching size might be okay).
-type ArgAbi: PartialEq;
+use specr::ArgAbi;
```
RalfJung, line 16:
```suggestion
// FIXME: This is incomplete.
#[derive(PartialEq)]
#[non_exhaustive]
enum ArgAbi {
    Register,
    Stack(Size, Align),
}
```
This is something MiniRust must define, I think... hiding it in specr won't help, anyway.^^

lang/syntax.md
```diff
@@ -6,7 +6,7 @@ First, the general structure of programs and functions:
 ```rust
 /// Some opaque type of function names.
 /// The details of this this is represented to not matter.
-type FnName;
+use specr::FnName;
```
RalfJung, line 9: Ah those will be tricky. I don't think we want specr to know which kinds of names we have. But in Coq, finding a good representation of names will be a key design choice.

I think we should either write this as
```rust
struct FnName(specr::Name);
```
or as something like
```rust
#[specr::name]
type FnName;
```
so that the tool do whatever magic it needs to do for variable names.

prelude/main.md
```diff
@@ -0,0 +1,71 @@
...
+/// means the program caused UB or put the machine to a halt.
+pub type Result<T=()> = std::result::Result<T, TerminationInfo>;
+
+#[non_exhaustive]
+pub enum TerminationInfo {
+  Ub(String),
+  MachineStop(String),
+}
+
+/// Some macros for convenient yeeting (yes this is valid syntax on nightly Rust).
```
RalfJung, line 16:
```suggestion
/// Some macros for convenient yeeting, i.e., return an error from a
/// `Option`/`Result`-returning function.
```
But also, we almost certainly don't want these macros in Coq like that.

Can you define a `yeet!` macro in specr so that these here can become `specr::yeet!(())` and `specr::yeet!(TerminationInfo::Ub(...))` etc? This will probably have to invoke the [FromResidual](https://doc.rust-lang.org/nightly/std/ops/trait.FromResidual.html) trait.

prelude/main.md
```diff
@@ -0,0 +1,71 @@
...
+    Mutable,
+    Immutable,
+}
+pub use Mutability::*;
+
+/// The type of mathematical integers.
+/// We assume all the usual arithmetic operations to be defined.
+pub use specr::BigInt;
+
+impl BigInt {
```
RalfJung, line 58:
````suggestion
```

We document some relevant parts of the BigInt API here.

```rust,ignore
impl BigInt {
````

File: prelude/main.md [hunk 8]
```diff
@@ -0,0 +1,71 @@
...
+    /// Returns the unique value that is equal to `self` modulo `2^size.bits()`.
+    /// If `signed == Unsigned`, the result is in the interval `0..2^size.bits()`,
+    /// else it is in the interval `-2^(size.bits()-1) .. 2^(size.bits()-1)`.
+    ///
+    /// `size` must not be zero.
+    pub fn modulo(self, signed: Signedness, size: Size) -> BigInt { todo!() }
```
RalfJung, line 64:
```suggestion
    pub fn modulo(self, signed: Signedness, size: Size) -> BigInt;
```

mem/basic.md
```diff
@@ -227,3 +227,14 @@ impl Memory for BasicMemory {
...
 }
 ```
+
+A size is valid, whenever it is non-negative and in-bounds for signed `PTR_SIZE`.
+
+```rust
+impl Memory for BasicMemory {
+    fn valid_size(size: Size) -> bool {
+        size.in_bounds(Signed, Self::PTR_SIZE)
+        && size >= 0
```
RalfJung, line 237: Please fix the formatting here (single line should suffice)

lang/syntax.md
```diff
@@ -6,7 +6,7 @@ First, the general structure of programs and functions:
 ```rust
 /// Some opaque type of function names.
 /// The details of this this is represented to not matter.
-type FnName;
+struct FnName(specr::Name);
```
memoryleak47, line 9: I preferred this representation for now, but we can always change it to `#[specr::name] type FnName` or something shall we need it.

prelude/endianness.md
```diff
@@ -0,0 +1,19 @@
...
+pub enum Endianness {
+    LittleEndian,
+    BigEndian,
+}
+
+pub use Endianness::*;
+
+impl Endianness {
+    /// If `signed == Signed`, the data is interpreted as two's complement.
+    pub fn decode(self, signed: Signedness, bytes: List<u8>) -> BigInt { todo!() }
```
memoryleak47, line 13: This is written as `todo!()` instead of `;` for now, so that things still compile.

memoryleak47:
Should be ready to review now! :)

prelude/bigint.md
```diff
@@ -0,0 +1,52 @@
...
+        let m = BigInt::from(2).pow(size.bits());
+
+        // n is in range `-(m-1)..m`.
+        let n = self % m;
+
+        match signed {
+            // if `Unsigned`, output needs to be in range `0..m`:
+            Unsigned if n < 0 => n + m,
+            // if `Signed`, output needs to be in range `-m/2 .. m/2`:
+            Signed if n >= m/2 => n - m,
```
RalfJung, line 42: What is if started out with `n < m/2`?

prelude/bigint.md
```diff
@@ -0,0 +1,52 @@
+BigInt is the type of mathematical integers.
+
+We assume all the usual arithmetic operations to be defined.
+Additionally, BigInt provides a few utility functions.
+
+```rust
+pub use specr::BigInt;
+
+impl BigInt {
```
RalfJung, line 9: This won't work if `BigInt` is in a different crate, right?

I think long-term we'll want to move all of this to the specr crate. But for now this is fine.

prelude/bigint.md
memoryleak47: whoops, is fixed now.

PR #44, opened 2022-11-08
some fixes

memoryleak47:
Remark:
`Memory: Sized` is apparently necessary due to the `Value<Self>` argument in `fn typed_store`

### DIFF
--- a/lang/step.md
+++ b/lang/step.md
@@ -56,13 +56,13 @@ impl<M: Memory> Machine<M> {
             Constant::Int(i) => Value::Int(i),
             Constant::Bool(b) => Value::Bool(b),
             Constant::Tuple(args) => {
-                let args = args.into_iter()
-                    .map(|arg| self.eval_constant(arg))
-                    .collect();
-                Value::Tuple(args)
+                let vals = args.into_iter()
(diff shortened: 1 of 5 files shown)


lang/step.md
```diff
@@ -56,13 +56,15 @@ impl<M: Memory> Machine<M> {
...
             Constant::Tuple(args) => {
-                let args = args.into_iter()
-                    .map(|arg| self.eval_constant(arg))
-                    .collect();
-                Value::Tuple(args)
+                let mut vals = Vec::new();
+                for arg in args {
+                    let val = self.eval_constant(arg)?;
+                    vals.push(val);
+                }
```
RalfJung, line 63: I'd rather use a functional `collect` here than an imperative `push` loop. Can't we get that to work?

memoryleak47: agreed, we cannot do `.map(|c| self.eval_constant(c)?)` though, as the `?` will try to return from the closure, not our outer function.
Anything else I've come up with, just wasn't easier than this imperative solution.

memoryleak47: nvm, `try_collect` is better

PR #45, opened 2022-11-09
crate specr

memoryleak47:
The commit makes `BigInt::{modulo, in_bounds}` a free function instead of a method.
So that `specr` can become a crate instead of a module.

Btw, we have two crates now called `specr` (ugh!)
I think we should rename specr (the transpiler) to `specr-transpile` or something.

### DIFF
--- a/prelude/bigint.md
+++ b/prelude/bigint.md
@@ -18,36 +18,36 @@ impl BigInt {
 
     /// Computes `self / other`, returns None if `other` is zero.
     pub fn checked_div(other: impl Into<BigInt>) -> Option<BigInt>;
+}
 
-    /// Returns the unique value that is equal to `self` modulo `2^size.bits()`.
-    /// If `signed == Unsigned`, the result is in the interval `0..2^size.bits()`,
-    /// else it is in the interval `-2^(size.bits()-1) .. 2^(size.bits()-1)`.
-    ///
-    /// `size` must not be zero.
-    pub fn modulo(self, signed: Signedness, size: Size) -> BigInt {
-        if size == 0 {
-            panic!("BigInt::modulo received invalid size zero!");
-        }
-
-        // the modulus.
-        let m = BigInt::from(2).pow(size.bits());
-
-        // n is in range `-(m-1)..m`.
-        let n = self % m;
-
-        match signed {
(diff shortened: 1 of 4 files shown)


prelude/bigint.md
```diff
@@ -18,36 +18,36 @@ impl BigInt {
...
-            Signed if n >= m/2 => n - m,
-            Signed if n < -m/2 => n + m,
-            _ => n,
-        }
+/// Returns the unique value that is equal to `i` modulo `2^size.bits()`.
+/// If `signed == Unsigned`, the result is in the interval `0..2^size.bits()`,
+/// else it is in the interval `-2^(size.bits()-1) .. 2^(size.bits()-1)`.
+///
+/// `size` must not be zero.
+pub fn modulo(i: BigInt, signed: Signedness, size: Size) -> BigInt {
```
RalfJung, line 28: My proposal was to move these functions to specr entirely. Here in Minirust we just want something like
````
For documentation purposes, some of the key BigInt functions are listed here, but they are all defined by specr.

```rust,ignore
impl BigInt {
    /// Returns the unique value that is equal to `i` modulo `2^size.bits()`.
    /// If `signed == Unsigned`, the result is in the interval `0..2^size.bits()`,
    /// else it is in the interval `-2^(size.bits()-1) .. 2^(size.bits()-1)`.
    ///
    /// `size` must not be zero.
    pub fn modulo(i: BigInt, signed: Signedness, size: Size) -> BigInt
}
```
````

RalfJung:
> Btw, we have two crates now called specr (ugh!)
> I think we should rename specr (the transpiler) to specr-transpile or something.

Ah hm... or we rename the thing imported here into `specrlib`?

prelude/bigint.md
memoryleak47: I think we cannot define it in specr, if we want specr to be a crate.
modulo uses `Size` and `Signedness` after all, right?

RalfJung: Right, so those types would have to move to specrlib. Hm...

In Coq we have `Z.modulo` which already always returns a positive number, not like the strange `%` operation Rust has. But we'd still have to duplicate signedness...

PR #46, opened 2022-11-10
fixes

memoryleak47:


lang/operator.md
```diff
@@ -126,7 +126,7 @@ impl<M: Memory> Machine<M> {
         // `offset.abs()` will fit into a `Size` since we did the overflow check above.
         // FIXME: actually, it could be isize::MIN and then everything breaks? Is that
         // a valid offset?
-        self.mem.dereferenceable(min_ptr, Size::from_bytes(offset.abs()), Align::ONE)?;
+        self.mem.dereferenceable(min_ptr, Size::from_bytes(offset.abs()), Align::from_bytes(1))?;
```
memoryleak47, line 129: There is no `const fn` BigInt constructor, as it's heap-allocated.
Another option would be `Align::one()`.

lang/step.md
```diff
@@ -69,7 +69,7 @@ impl<M: Memory> Machine<M> {
     }
 
     fn eval_value(&mut self, ValueExpr::Constant(constant, _ty): ValueExpr) -> NdResult<Value<M>> {
-        self.eval_constant(constant)
```
memoryleak47, line 72: This code *seemed* correct, but when assembled into a match block it caused an error.
The other match arms return `Value`, which is expected to be converted to `NdResult` by `specr::hidden::ret`.

lang/step.md
```diff
@@ -200,7 +201,7 @@ impl<M: Memory> Machine<M> {
         let offset = match ty {
             Type::Array { elem, count } => {
                 if index < count {
-                    index * elem.size::<M>()
+                    elem.size::<M>() * index
```
memoryleak47, line 204: the `Size` operator overload only works this way around

lang/values.md
```diff
@@ -97,7 +97,7 @@ Note, in particular, that `bool` just entirely ignored provenance; we discuss th
 ```rust
 impl Type {
     fn decode<M: Memory>(Type::Int(IntType { signed, size }): Self, bytes: List<AbstractByte<M::Provenance>>) -> Option<Value<M>> {
-        if bytes.len() != size.bytes() {
+        if size.bytes() != bytes.len() {
```
memoryleak47, line 100: again, operator overload ordering. `BigInt * usize` is defined, but `usize * BigInt` is not

prelude/size.md
```diff
@@ -28,5 +26,42 @@ impl Size {
 
     pub fn bytes(self) -> BigInt { self.raw }
     pub fn bits(self) -> BigInt { self.raw * 8 }
+    pub fn is_zero(&self) -> bool { self.raw == 0 }
+}
+
+// We implement a few operators for size.
```
memoryleak47, line 32: Soo, I decided to implement a few operators for `Size`.
It doesn't feel great having that in the spec really.

But: It feels even worse writing `Size::from_bytes(a.bytes() + b.bytes())` instead of `a+b` over and over again.
and if `specr` is a crate, it cannot secretly implement `Size`s operators.

lang/values.md
```diff
@@ -181,19 +181,19 @@ Note that types like `&!` have no valid value: when the pointee type is uninhabi
...
         )
     }
     fn encode<M: Memory>(Type::Tuple { fields, size }: Self, val: Value<M>) -> List<AbstractByte<M::Provenance>> {
         let Value::Tuple(values) = val else { panic!() };
         assert_eq!(values.len(), fields.len());
-        let mut bytes = list![AbstractByte::Uninit; size];
+        let mut bytes = list![AbstractByte::Uninit; size.bytes()];
         for ((offset, ty), value) in fields.iter().zip(values) {
-            bytes[offset..][..ty.size::<M>()].copy_from_slice(ty.encode::<M>(value));
+            bytes[offset.bytes()..][..ty.size::<M>().bytes()].copy_from_slice(ty.encode::<M>(value));
```
memoryleak47, line 196: I actually don't like how much `.bytes()` we need to spam sometimes.
But I cannot implement the `List` in a way that understands `Size`,
except if we implement `Deref<Target=BigInt>` or some other trait.
But that would kinda defeat the purpose of the Size newtype.

prelude/size.md
memoryleak47: Also, I can extract this into another PR if you want. Doesn't really qualify as "fixes" ^^

lang/operator.md
RalfJung: Which type are you using for BigInt?
In principle a non-heap case would be possible for small numbers. But I can also live with `Align::one()`.

lang/values.md
RalfJung: Can `prelude/size.md` implement `Index<RangeFrom<Size>> for List<T>` and `Index<RangeTo<Size>> for List<T>`?

lang/well-formed.md
```diff
@@ -61,26 +61,25 @@ impl Type {
                 // The fields must not overlap.
                 // We check fields in the order of their (absolute) offsets.
                 fields.sort_by_key(|(offset, _ty)| offset);
-                let mut last_end = Size::ZERO;
+                let mut last_end = Size::from_bytes(0);
```
RalfJung, line 64:
```suggestion
                let mut last_end = Size::zero();
```

lang/well-formed.md
```diff
@@ -61,26 +61,25 @@ impl Type {
...
                 for (offset, ty) in fields {
                     // Recursively check the field type.
                     ty.check_wf::<M>()?;
                     // And ensure it fits after the one we previously checked.
                     ensure(offset >= last_end)?;
-                    last_end = offset.checked_add(ty.size::<M>())?;
+                    last_end = offset + ty.size::<M>();
                 }
                 // And they must all fit into the size.
                 ensure(size >= last_end)?;
```
RalfJung, line 73:
```suggestion
                // And they must all fit into the size.
                // The size is in turn checked to be valid for `M`, and hence all offsets are valid, too.
                ensure(size >= last_end)?;
```

prelude/size.md
RalfJung: Yeah this will definitely not translate well to Coq...

... I think you convinced me that Size and Align should be defined in specr, not in MiniRust.

lang/values.md
RalfJung: Ah but that would assume all lists are byte-indexed, not bit-indexed... :/

lang/operator.md
memoryleak47: I'm currently using the crate `num-bigint`.
Yeah I also thought whether we made some kind of
`enum Int { Big(BigInt), Small(i64) }`, would be kinda overkill though I think ^^

memoryleak47:
most changes are in the smaller PR #48, the rest requires new design choices.
Mostly around whether we allow `Size` handling to be implicitly as bytes sometimes, or whether we decide to write it out explicitly.

PR #47, opened 2022-11-10
add missing Hash & Clone trait bounds

memoryleak47:


mem/interface.md
```diff
@@ -34,14 +34,14 @@ We define the `AbstractByte` type as follows, where `Provenance` will later be i
 
 ```rust
 #[derive(PartialEq, Eq)]
-pub enum AbstractByte<Provenance> {
+pub enum AbstractByte<Provenance: Eq + Hash + Clone> {
```
RalfJung, line 37: In Rust we usually avoid trait bounds in the types and only have them on the `impl` blocks that need them.

lang/values.md
```diff
@@ -284,7 +284,7 @@ trait DefinedRelation {
 Starting with `AbstractByte`, we define `b1 <= b2` ("`b1` is less-or-equally-defined as `b2`") as follows:
 
 ```rust
-impl<Provenance: Eq> DefinedRelation for AbstractByte<Provenance> {
+impl<Provenance: Eq + Hash + Clone> DefinedRelation for AbstractByte<Provenance> {
```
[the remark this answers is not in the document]
memoryleak47: `AbstractByte` needs Provenance as generic param, or do you mean that the trait bounds are unnecessary? I guess that might be true.

mem/interface.md
memoryleak47: Ah right, I've heard that before. What was the rationale for that again?

RalfJung: I think the ideas is to generally only state assumptions where we actually need them.

lang/values.md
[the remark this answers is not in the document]
RalfJung: Yes I was asking about the trait bounds.

memoryleak47:
conflicts with other PR are fixed :+1:

lang/values.md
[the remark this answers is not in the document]
RalfJung: This is still open though. What happens if you remove these extra bounds, in particular `Hash`?

memoryleak47:
`try_insert` also requires `Debug` on a few more traits, I'll quickly add this.

lang/values.md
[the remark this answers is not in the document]
memoryleak47: Sorry, didn't see the comment. Fixed now.

lang/syntax.md
```diff
@@ -6,6 +6,7 @@ First, the general structure of programs and functions:
 ```rust
 /// Some opaque type of function names.
 /// The details of this this is represented to not matter.
+#[derive(Hash, PartialEq, Eq, Debug)]
```
RalfJung, line 9: Not `Clone`?

memoryleak47: Yeah, `#[derive(Clone)]` is added by specr as of now.
We should probably have some document explaining what specr actually does to the original code ^^'

RalfJung: Oh I see.

That'd be the advantage of a `#[specr::name]` attribute -- it could hide all these impls. For Coq we don't want that `Hash` or `Debug` stuff I am pretty sure...

memoryleak47: So `#[specr::name]` would implement Clone, Copy and maybe GcCompat?

I'm wondering, would there be enums/structs which we don't want to give `#[specr::name]`?

RalfJung: `#[specr::name]` is meant as a marker specifically for types that represent 'names' and are used as keys in a `Map`. We don't want to use this for all types.

PR #49, opened 2022-11-11
more fixes

memoryleak47:


lang/machine.md
```diff
@@ -37,7 +37,7 @@ struct StackFrame<M: Memory> {
     /// The first component identifies the basic block,
     /// the second the statement inside that basic block.
     /// If the index is len+1, it refers to the terminator.
-    next: (BbName, u64),
+    next: (BbName, BigInt),
```
memoryleak47, line 40: I presume we want that?

mem/basic.md
```diff
@@ -166,7 +166,7 @@ impl BasicMemory {
         // FIXME: This is *not* what rustc does, since from this it follows that
         // `offset(0)` is allowed on all non-null pointers which does not match
         // the LLVM IR rustc generates.
-        if len == 0 {
+        if len.is_zero() {
```
memoryleak47, line 169: I didn't want to implement `PartialEq<i32>` for `Size`, due to the bits v. bytes ambiguity.
Hence we need `ìs_zero`

lang/machine.md
RalfJung: Yeah I think we never want a bounded type in minirust. Good catch!

PR #51, opened 2022-11-14
Fixes

memoryleak47:


prelude/endianness.md
```diff
@@ -10,10 +10,10 @@ pub use Endianness::*;
 
 impl Endianness {
     /// If `signed == Signed`, the data is interpreted as two's complement.
-    pub fn decode(self, signed: Signedness, bytes: List<u8>) -> BigInt { todo!() }
+    pub fn decode(self, signed: Signedness, bytes: List<u8>) -> BigInt;
```
RalfJung, line 13: Shouldn't the entire block be `rust,ignore`?

memoryleak47: So far, I preferred to remove these "forward declarations", (i.e. method impls with a `;` block) in specr.
Otherwise one would need to cut code like this into two separate code blocks.  (which I presumed looks ugly?)
One code block for `enum Endianness` which shall not be ignored, and another block for ignored methods.

RalfJung: `Endianess` has to live in specrlib anyway doesn't it?

memoryleak47: Interesting, I didn't expect to define `Endianness` in libspecr, but yeah that works as well I guess.
I'll close this for now, and open another PR as `subslice_with_length` need to change aswell due to our `im` List backend ideas.

PR #52, opened 2022-11-18
List changes

memoryleak47:


lang/machine.md
```diff
@@ -45,12 +45,17 @@ We also define some helper functions that will be useful later.
 
 ```rust
 impl<M: Memory> Machine<M> {
-    fn cur_frame(&self) -> &StackFrame<M> {
+    fn cur_frame(&self) -> StackFrame<M> {
```
memoryleak47, line 48: This is again necessary, as we can't return refs into `GC_STATE`, and also it's not too costly as `StackFrame` contains just a bunch of usizes.

lang/step.md
```diff
@@ -28,7 +28,9 @@ impl<M: Memory> Machine<M> {
         } else {
             // Bump up PC, evaluate this statement.
             let stmt = block.statements[next_stmt];
-            self.cur_frame_mut().next_stmt = next_stmt + 1;
```
memoryleak47, line 31: `frame.next_stmt` does not exist anymore, it's `next: (BbName, BigInt)` now apparently.

lang/step.md
```diff
@@ -275,13 +277,17 @@ impl<M: Memory> Machine<M> {
...
+        });
     }
 
     fn eval_statement(&mut self, Statement::StorageDead(local): Statement) -> NdResult {
         // Here we make it a spec bug to ever mark an already dead local as dead.
         let layout = self.cur_frame().func.locals[local].layout::<M>();
-        let p = self.cur_frame_mut().locals.remove(local).unwrap();
+        let p = self.mutate_cur_frame(|frame| {
+            frame.locals.remove(local).unwrap()
+        });
```
memoryleak47, line 290: This is the reason for the generic `O` argument of `mutate_cur_frame`.

lang/step.md
RalfJung: Ah right my last commit here was buggy.

lang/step.md
```diff
@@ -28,7 +28,9 @@ impl<M: Memory> Machine<M> {
         } else {
             // Bump up PC, evaluate this statement.
             let stmt = block.statements[next_stmt];
-            self.cur_frame_mut().next_stmt = next_stmt + 1;
+            self.mutate_cur_frame(|frame| {
+                frame.next.1 = next_stmt + 1;
```
RalfJung, line 32: I'll refactor this a bit more after merging, I don't like the assignment to an anonymous tuple field here.

lang/values.md
```diff
@@ -193,7 +197,7 @@ impl Type {
         assert_eq!(values.len(), fields.len());
         let mut bytes = list![AbstractByte::Uninit; size.bytes()];
         for ((offset, ty), value) in fields.iter().zip(values) {
-            bytes[offset.bytes()..][..ty.size::<M>().bytes()].copy_from_slice(ty.encode::<M>(value));
+            bytes.write_subslice_with_length(offset.bytes(), ty.size::<M>().bytes(), ty.encode::<M>(value));
```
RalfJung, line 200: `ty.size::<M>()` is redundant here, right? The length of `ty.encode::<M>(value)` already says all we need.

lang/step.md
RalfJung: Or maybe you can do that... basically replace `next` by `next_block` and `next_stmt`.

lang/step.md
```diff
@@ -384,7 +394,10 @@ impl<M: Memory> Machine<M> {
         }
 
         // Advance the PC for this stack frame.
-        self.cur_frame_mut().next = (next_block, BigInt::zero());
+        self.mutate_cur_frame(|frame| {
+            frame.next = (next_block, BigInt::zero());
```
RalfJung, line 398: Might be worth factoring this into a `frame.jump_to_block()` so that one cannot forget to also update the next_stmt to 0.

memoryleak47: Yeah, that sounds good.
I think as `jump_to_block` would only be used as `self.mutate_cur_frame(|frame|) frame.jump_to_block(_))`
it might make sense to make it `machine.jump_to_block(_)`?

lang/machine.md
```diff
@@ -33,11 +33,12 @@ struct StackFrame<M: Memory> {
     /// The place where the caller wants to see the return value.
     caller_ret_place: Place<M>,
 
-    /// The next statement/terminator to execute (the "program counter").
-    /// The first component identifies the basic block,
-    /// the second the statement inside that basic block.
-    /// If the index is len+1, it refers to the terminator.
```
memoryleak47, line 39: This `len+1` was one-off, right?

lang/step.md
RalfJung: Hm, conceptually it is a frame action... I think defining it on the frame is better. It exposes structurally that that is the only piece that changes.

lang/machine.md
RalfJung: Ah, good catch.

lang/machine.md
```diff
@@ -33,11 +33,12 @@ struct StackFrame<M: Memory> {
...
-    next: (BbName, BigInt),
+    /// `next_block` and `next_stmt` describe the next statement/terminator to execute (the "program counter").
+    /// `next_block` identifies the basic block,
+    next_block: BbName,
+    /// `next_stmt` specifies the statement inside that basic block.
+    /// If `next_stmt` is len, it refers to the terminator.
```
RalfJung, line 40:
```suggestion
    /// If `next_stmt` is equal to the number of statements in this block (an
    /// out-of-bounds index in the statement list), it refers to the terminator.
```

lang/values.md
```diff
@@ -197,7 +197,7 @@ impl Type {
         assert_eq!(values.len(), fields.len());
         let mut bytes = list![AbstractByte::Uninit; size.bytes()];
         for ((offset, ty), value) in fields.iter().zip(values) {
-            bytes.write_subslice_with_length(offset.bytes(), ty.size::<M>().bytes(), ty.encode::<M>(value));
+            bytes.write_subslice_at_offset(offset.bytes(), ty.encode::<M>(value));
```
RalfJung, line 200: In a list I think we usually say "index", not "offset", right?

lang/step.md
memoryleak47: okay, is changed now.

PR #53, opened 2022-11-21
Fixes

memoryleak47:


lang/values.md
```diff
@@ -80,7 +80,7 @@ impl Type {
         if bytes.len() != 1 {
             throw!();
         }
-        match bytes[0] {
+        match bytes[BigInt::zero()] {
```
memoryleak47, line 83: or do we want `impl Into<BigInt>`?

lang/values.md
```diff
@@ -327,7 +327,7 @@ impl<Provenance: Eq + Clone> DefinedRelation for Pointer<Provenance> {
 The order on `List<AbstractByte>` is assumed to be such that `bytes1 <= bytes2` if and only if they have the same length and are bytewise related by `<=`.
 In fact, we define this to be in general how lists are partially ordered (based on the order of their element type):
 ```rust
-impl<T: DefinedRelation> DefinedRelation for List<T> {
+impl<T: DefinedRelation + specr::Obj> DefinedRelation for List<T> {
```
memoryleak47, line 330: I think this will come up more and more, you're probably right that we want some automated solution for those at some point.

lang/well-formed.md
```diff
@@ -217,7 +217,7 @@ impl PlaceExpr {
     fn check_wf<M: Memory>(self, locals: Map<LocalName, PlaceType>) -> Option<PlaceType> {
         use PlaceExpr::*;
         match self {
-            Local(name) => locals.get(name),
+            Local(name) => locals.get(name)?,
```
memoryleak47, line 220: This is again required as the other arms return `PlaceType` directly, so this arm cannot return `Option<PlaceType>`.

lang/well-formed.md
```diff
@@ -373,14 +373,14 @@ impl Function {
                 } else {
                     // A new block.
                     bb_live_at_entry.insert(block_name, live_locals);
-                    todo.push_back(block_name);
+                    todo.push(block_name);
```
memoryleak47, line 376: Or do you prefer `push_back` over `push`?
Generally, I tried to be as close to `Vec` as possible API-wise.

lang/values.md
RalfJung: For indexing? I mean if it's not too much to ask... yeah indexing with any `impl Into<BigInt>` would be nice. :)

lang/values.md
RalfJung: Yeah, we probably just want this added to all generic parameters.

lang/well-formed.md
RalfJung: `push` is fine.

mem/interface.md
```diff
@@ -82,9 +82,9 @@ pub struct Pointer<Provenance> {
 /// executing the same operation on the same memory can have different results.
 /// We also let read operations potentially mutate memory (they actually can
 /// change the current state in concurrent memory models and in Stacked Borrows).
-pub trait Memory: Sized {
+pub trait Memory: specr::Obj {
     /// The type of pointer provenance.
-    type Provenance: Eq + Hash + Clone + Debug;
+    type Provenance: specr::Obj + Eq + Hash;
```
RalfJung, line 87: `Eq` and `Hash` should not be needed any more, after our discussion? :)

(Again would be nice to just add the bounds automatically. They will look pretty different in Coq.)

lang/values.md
memoryleak47: then again we have the `FnOnce` problem. `mutate_cur_frame`s generic function argument is not `GcCompat`.

memoryleak47: Do we want to add some kind of opt-out, maybe?
Maybe `?specr::Obj`?

RalfJung: Oh right.

Can you detect `FnOnce` and then not add the bound? (The Coq backend will also have to treat `impl FnOnce` special.)

RalfJung:
(Github doesn't always send notifications for force-pushes, so please make sure to also add a comment when things are ready for another round of review.)

prelude/endianness.md
```diff
@@ -1,6 +1,6 @@
 The endianness defines how integers and pointers are encoded and decoded.
```
RalfJung, line 1:
```suggestion
The endianness defines how integers and pointers are encoded and decoded.
This is defined by specr, we just repeat the signature here for reference.
```

PR #54, opened 2022-11-24
Fixes

memoryleak47:


lang/types.md
```diff
@@ -22,9 +22,9 @@ In the future, we might want to separate a type from its layout, and consider th
 ```rust
 /// A "layout" describes the shape of data in memory.
 struct Layout {
-    size: Size,
-    align: Align,
-    inhabited: bool,
+    pub size: Size,
```
memoryleak47, line 25: These fields are used in `BasicMemory::retag_ptr`.
We could also keep it private and add `.size()` etc. methods for Layout, but I think thats a bit verbose for a spec.

mem/basic.md
```diff
@@ -235,7 +236,7 @@ A size is valid, whenever it is non-negative and in-bounds for signed `PTR_SIZE`
 ```rust
 impl Memory for BasicMemory {
     fn valid_size(size: Size) -> bool {
-        size.in_bounds(Signed, Self::PTR_SIZE) && size >= 0
```
memoryleak47, line 238: I think it makes more sense to check that a Size is non-negative when constructing `Size`.

prelude/size.md
```diff
@@ -19,12 +19,18 @@ impl Size {
         let bits = bits.into();
 
         // round up `bits / 8`
-        let raw = bits / 8 + ((bits % 8) + 7) / 8;
-        Size { raw }
+        let bytes = bits / 8 + ((bits % 8) + 7) / 8;
```
memoryleak47, line 22: This calls `Size::from_bytes` to not repeat the `bytes < 0` check.

mem/basic.md
RalfJung: I like that -- Size as 'unsigned BigInt', basically. Or as `BigNat`.

Can you update the `Size` docs accordingly?

memoryleak47: @RalfJung done!

RalfJung: Please also add doc comments to `from_bits` and `from_bytes` saying that they will panic then the argument is negative.

memoryleak47: uh, as we are rounding up an input from `[-7, -1]` wouldn't even panic from_bits, I'll change that.

memoryleak47:
@RalfJung ready for review

mem/basic.md
RalfJung: Oh though got rounded up to 0? Ouch, good catch.

PR #55, opened 2022-11-24
implement Align::restrict_for_offset and related fn

memoryleak47:


prelude/bigint.md
```diff
@@ -16,6 +19,9 @@ impl BigInt {
     /// Checks whether `self` is a power of two.
     pub fn is_power_of_two(&self) -> bool;
 
+    /// Returns the absolute value of `self`.
+    pub fn abs(self) -> BigInt;
+
```
memoryleak47, line 24: whoops, forgot to mention a few BigInt methods..

prelude/align.md
```diff
@@ -23,5 +23,26 @@ impl Align {
     pub fn bytes(self) -> BigInt {
         self.raw
     }
+
+    /// Computes the best alignment possible for the given offset
+    /// (the largest power of two that the offset is a multiple of).
+    ///
+    /// N.B., for an offset of `0`, this returns `2^64`.
```
memoryleak47, line 30: Do you prefer another handling of the `offset == zero` case?
This is how rustc does it, but `2^64` is somewhat arbitrary for us - as we are working with BigInt.

RalfJung: This should probably return `Option<Align>`, and return `None` if the offset is zero.

prelude/align.md
```diff
@@ -23,5 +23,26 @@ impl Align {
...
+            None => BigInt::from(64),
+        };
+
+        Align::from_bytes(BigInt::one() << exp)
+    }
+
+    /// Lower the alignment, if necessary, such that the given offset
+    /// is aligned to it (the offset is a multiple of the alignment).
+    pub fn restrict_for_offset(self, offset: Size) -> Align {
+        self.min(Align::max_for_offset(offset))
```
RalfJung, line 45: Where is `min` coming from?

prelude/bigint.md
```diff
@@ -7,6 +7,9 @@ Additionally, BigInt provides a few utility functions.
 pub use specr::BigInt;
 
 impl BigInt {
+    pub fn zero() -> BigInt;
+    pub fn one() -> BigInt;
```
RalfJung, line 11: This overlaps with the other PR, doesn't it?

prelude/bigint.md
```diff
@@ -30,7 +36,7 @@ impl BigInt {
         }
 
         // the modulus.
-        let m = BigInt::from(2).pow(size.bits());
+        let m = BigInt::one() << size.bits();
```
RalfJung, line 39: Please don't use shifts to encode exponentiation. That's a low-level bit manipulation trick, nothing we want to use in a specification.

prelude/align.md
memoryleak47: Align impls `Ord`, I've added a comment for it.

memoryleak47:
@RalfJung ready for review

prelude/align.md
```diff
@@ -3,6 +3,8 @@ See [Align](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_target/abi/str
 
 `Align` is always a power of two.
 
+For convenience, Align implements `Ord`.
```
RalfJung, line 6: Yeah but where does that come from? I would have expected a `derive` or so?

PR #56, opened 2022-11-25
add Size::ZERO, Align::ONE constants

memoryleak47:
I did this `const BigInt` thing, which allows for `Align::ONE`, `Size::ZERO` etc.
```rust
pub enum BigInt {
    Big(GcCow<ExtBigInt>),
    Small(i128),
}
```
(Btw, do we want to rename `BigInt` to `Int`?)

### DIFF
--- a/prelude/size.md
+++ b/prelude/size.md
@@ -11,9 +11,7 @@ Users needs check whether a given `Size` is too large for their Machine themselv
 pub struct Size { raw: BigInt }
 
 impl Size {
-    pub fn zero() -> Size {
-        Size { raw: BigInt::from(0) }
-    }
+    pub const ZERO: Size = Size { raw: BigInt::ZERO };
 
     /// Rounds `bits` up to the next-higher byte boundary, if `bits` is
     /// not a multiple of 8.
--- a/prelude/size.md
+++ b/prelude/size.md
@@ -28,6 +28,14 @@ impl Size {
         Size { raw }
     }
 
+    /// variation of `from_bits` for const contexts.
+    /// Cannot fail since the input is unsigned.
+    pub const fn from_bits_const(bits: u64) -> Size {
+        let bytes = bits / 8 + ((bits % 8) + 7) / 8;
+        let raw = Int::from(bytes);
+        Size { raw }
+    }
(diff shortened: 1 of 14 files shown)


prelude/size.md
```diff
@@ -11,9 +11,7 @@ Users needs check whether a given `Size` is too large for their Machine themselv
 pub struct Size { raw: BigInt }
 
 impl Size {
-    pub fn zero() -> Size {
-        Size { raw: BigInt::from(0) }
-    }
+    pub const ZERO: Size = Size::ZERO;
```
RalfJung, line 14:
```suggestion
    pub const ZERO: Size = Size::ZERO;
```
This looks recursive? :D

memoryleak47: whoops, missed that one ^^

memoryleak47:
@RalfJung ready for review

prelude/size.md
```diff
@@ -30,8 +28,15 @@ impl Size {
         Size { raw }
     }
 
+    /// variation of `from_bits` for const contexts.
```
RalfJung, line 31:
```suggestion
    /// variation of `from_bits` for const contexts.
    /// Cannot fail since the input is unsigned.
```

RalfJung, line 31: Please also add a `from_bytes_const` for completeness' sake.

memoryleak47:
@RalfJung suggestions are implemented

prelude/size.md
```diff
@@ -30,8 +28,16 @@ impl Size {
...
+    /// Cannot fail since the input is unsigned.
+    pub const fn from_bits_const(bits: u64) -> Size {
+        let bytes = bits / 8 + ((bits % 8) + 7) / 8;
```
RalfJung, line 34: This is not exactly readable, we should find a way to express this more nicely...

In fact [div_ceil](https://doc.rust-lang.org/nightly/std/primitive.u64.html#method.div_ceil) seems like exactly the right thing? Can you use that here, and also add it to `Int`? I'd like to keep MiniRust clean of terrible things like "how to do integer division rounding up"^^

PR #58, opened 2022-11-27
Fixes

memoryleak47:


lang/step.md
```diff
@@ -132,7 +132,7 @@ For now, that is just a pointer (but this might have to change).
 Place evaluation ensures that this pointer is always dereferenceable (for the type of the place expression).
 
 ```rust
-type Place<M: Memory> = Pointer<M::Provenance>;
```
memoryleak47, line 135: Rusts type aliases don't support bounds as of now.
We could implement something in specr, but I think this change is fine too.

lang/types.md
```diff
@@ -21,7 +21,7 @@ In the future, we might want to separate a type from its layout, and consider th
 
 ```rust
 /// A "layout" describes the shape of data in memory.
-struct Layout {
+pub struct Layout {
```
memoryleak47, line 24: some modules from mem use `Layout`

lang/well-formed.md
```diff
@@ -186,7 +187,10 @@ impl ValueExpr {
                         Type::Int(IntType { signed: Unsigned, size: M::PTR_SIZE })
                     }
                     Int2Ptr(ptr_ty) => {
-                        ensure(matches!(operand, Type::Int(IntType { signed: Unsigned, size: M::PTR_SIZE })))?;
```
memoryleak47, line 189: Sadly one cannot match like this `size: M::PTR_SIZE`.

lang/step.md
RalfJung: This will probably make the Coq translator unhappy, but yeah I guess it works for now.

lang/well-formed.md
RalfJung: Does `Type` have equality? Then we could just use `==` instead of `matches!`?

memoryleak47: Yep, thats better!

PR #64, opened 2022-12-02
add basic CI

memoryleak47:


.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
...
+    name: Check whether minirust succeeds to build.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: download minirust-tooling
+        run: |
+          git clone "https://github.com/memoryleak47/minirust-tooling" ../minirust-tooling
+          cd ../minirust-tooling
+          git checkout 19f3392 # Fixed commit, we need to bump this occasionally.
```
memoryleak47, line 16: This will probably be part a config file at some point.

RalfJung: I think instead of having a ton of tiny actions with shell scripts in them, I'd prefer one action that just calls `bash ci.sh` and then that script does the rest.

.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
+name: Build
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    name: Check whether minirust succeeds to build.
```
RalfJung, line 7:
```suggestion
    name: specr build
```
The current name doesn't even fit the column in the actions view^^

RalfJung:
Strangely I cannot tell github to run your jobs... so I pushed this to my own branch which then seems to run: https://github.com/RalfJung/minirust/actions/runs/3608243738/jobs/6080739020.

.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
...
+          cd ../minirust-tooling
+          git checkout 19f3392 # Fixed commit, we need to bump this occasionally.
+
+      - name: link
+        run: |
+          ln -s ${{github.workspace}} ../minirust-tooling/minirust
+
+      - name: setup nightly
+        run: |
+          rustup install nightly
```
RalfJung, line 24: We also need to pin the nightly, since bumping the nightly might change/break some features we use.

.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
...
+jobs:
+  build:
+    name: Check whether minirust succeeds to build.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: download minirust-tooling
+        run: |
+          git clone "https://github.com/memoryleak47/minirust-tooling" ../minirust-tooling
```
RalfJung, line 14: So `cargo install specr --git https://github.com/memoryleak47/minirust-tooling#19f3392` or something like that does not work? I guess the problem is that it needs to know where to find libspecr?

memoryleak47: As of now `minirust-tooling` requires a particular directory structure to function.
`minirust-tooling/minirust` needs to be present, it can then transpile this to `minirust-tooling/gen-minirust`, which is consumed by eg. `minimize`.

I want this to become more flexible later on, I think.
But for now it's probably easiest to `clone` it, and then soft-link `minirust` to the correct spot.

.github/workflows/build.yml
memoryleak47: Yeah, thats a good point.
I pinned it in `minirust-tooling` now - wanted to do that anyways.

.github/workflows/build.yml
memoryleak47: Are you fine with having just one large action?
I kinda fear someone might actually run `ci.sh` on their computer.

.github/workflows/build.yml
RalfJung: Okay, fair.

.github/workflows/build.yml
RalfJung: I have such `ci.sh` scripts in most of my projects and never had people run into problems. I think it's better to keep shell script logic outside these awfully verbose yml files with their strange escaping and indentation rules.

.github/workflows/build.yml
memoryleak47: `minirust-tooling/specr-transpile` transpiles into `minirust-tooling/gen-minirust`, which is affected by the `minirust-tooling/rust-toolchain.toml`.

.github/workflows/build.yml
memoryleak47: @RalfJung alright, should be ready to review now

ci.sh
```diff
@@ -0,0 +1,13 @@
...
+# Fixed `minirust-tooling` commit, we need to bump this occasionally.
+TOOLING_COMMIT="0441a98"
+
+git clone "https://github.com/memoryleak47/minirust-tooling" ~/minirust-tooling
+ln -s ~/work/minirust/minirust ~/minirust-tooling/minirust
+
+cd ~/minirust-tooling
+git checkout "$TOOLING_COMMIT"
+
+(cd specr-transpile; cargo run)
```
memoryleak47, line 12: This cannot be `cargo run --manifest-path specr-transpile`, as `specr-transpile` needs this working directory.
That's also something I intend to improve later on.

lang/step.md
```diff
@@ -423,7 +423,7 @@ impl<M: Memory> Machine<M> {
         let func = frame.func;
         let Some((ret_local, _)) = func.ret else {
             throw_ub!("Return from a function that does not have a return place");
-        }
+        };
```
memoryleak47, line 426: good thing we have a CI now ^^

.github/workflows/build.yml
RalfJung: Ah, a `rust-toolchain.toml`... I'm not a huge fan of those but I guess it'll do for now.

PR #66, opened 2022-12-05
impl Intrinsics

memoryleak47:
A few assumptions I have made, which should be checked by you first:

Every function or intrinsic either "might-return" or is "never-returning".
"never-returning" are `Program::start`, `Intrinsic::Exit`, `-> !` fns.
Everything else "might-return".

For a "never-returning" function, `caller_ret_place`, `Function::ret`, and `next_block + ret` from `Terminator::Call` are required to be `None`.
Analog for `Intrinsic`s.
For functions that "might-return", those are required to be `Some`.

### DIFF
--- a/lang/intrinsic.md
+++ b/lang/intrinsic.md
@@ -1,0 +1,49 @@
+# Intrinsics
+
+```rust
+pub enum Intrinsic {
+    Exit,
+    PrintStdout,
+    PrintStderr,
+}
+
+impl<M: Memory> Machine<M> {
+    fn eval_intrinsic(
+        &mut self,
+        intrinsic: Intrinsic,
+        arguments: List<Value<M>>,
+        ret: Option<Place<M>>,
+        next_block: Option<BbName>,
+    ) -> NdResult {
+        match intrinsic {
+            Intrinsic::Exit => throw_machine_stop!("0"),
+            Intrinsic::PrintStdout => self.eval_print(&mut std::io::stdout(), arguments, ret, next_block)?,
+            Intrinsic::PrintStderr => self.eval_print(&mut std::io::stderr(), arguments, ret, next_block)?,
+        }
+    }
+
+    fn eval_print(
+        &mut self,
+        stream: &mut impl std::io::Write,
+        arguments: List<Value<M>>,
+        _ret: Option<Place<M>>,
+        next_block: Option<BbName>
+    ) -> NdResult {
+        let Some(next_block) = next_block else {
+            throw_ub!("missing next_block for print intrinsic");
+        };
+
+        for arg in arguments {
+            match arg {
+                Value::Int(i) => write!(stream, "{}\n", i).unwrap(),
+                Value::Bool(b) => write!(stream, "{}\n", b).unwrap(),
+                _ => throw_ub!("unsupported value for printing"),
+            }
+        }
+
+        self.mutate_cur_frame(|frame| {
+            frame.jump_to_block(next_block);
+        });
+    }
+}
+```
--- a/lang/step.md
+++ b/lang/step.md
@@ -359,19 +359,30 @@ impl<M: Memory> Machine<M> {
         };
         let mut locals: Map<LocalName, Place<M>> = default();
 
+        // If the callee might return, then `ret` and `next_block` should be `Some`.
+        // And if it never returns, they are required to be `None`.
+        let fn_might_return = func.ret.is_some();
+        if fn_might_return != ret.is_some() || fn_might_return != next_block.is_some() {
+            throw_ub!("Contradicting information on whether function might return");
+        }
+
         // First evaluate the return place and remember it for `Return`. (Left-to-right!)
-        let (caller_ret_place, caller_ret_abi) = ret;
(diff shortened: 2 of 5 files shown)


lang/intrinsic.md
```diff
@@ -0,0 +1,49 @@
...
+            Intrinsic::PrintStdout => self.eval_print(&mut std::io::stdout(), arguments, ret, next_block)?,
+            Intrinsic::PrintStderr => self.eval_print(&mut std::io::stderr(), arguments, ret, next_block)?,
+        }
+    }
+
+    fn eval_print(
+        &mut self,
+        stream: &mut impl std::io::Write,
+        arguments: List<Value<M>>,
+        _ret: Option<Place<M>>,
```
memoryleak47, line 29: Do we need to assign `()` to `ret`?
`print` is definitely not "never-returning".

memoryleak47: As of this PR `print` would accept both a `Some` and a `None` `ret`.
I think we would decide on which is correct before merging this.
I'm tending towards `Some`, as `print` behaves like a `-> ()` function, which also has a return place.

lang/step.md
```diff
@@ -359,19 +359,30 @@ impl<M: Memory> Machine<M> {
         };
         let mut locals: Map<LocalName, Place<M>> = default();
 
+        // If the callee might return, then `ret` and `next_block` should be `Some`.
+        // And if it never returns, they are required to be `None`.
+        let fn_might_return = func.ret.is_some();
+        if fn_might_return != ret.is_some() || fn_might_return != next_block.is_some() {
+            throw_ub!("Contradicting information on whether function might return");
+        }
```
memoryleak47, line 367: Instead of using `match` or `if let` to access `ret` and `next_block` I prefer to check `fn_might_return` and then unwrap.
This makes it clear that there are only two options:
`func.ret, ret, next_block` are all `Some` or `None`.

Do you agree with this? This would mean that `-> ()` functions do return like normal functions.
And only `-> !`, `start` and `Exit` have `None`, I think.

lang/step.md
```diff
@@ -359,19 +359,30 @@ impl<M: Memory> Machine<M> {
...
             if caller_ret_abi != callee_ret_abi {
                 throw_ub!("call ABI violation: return ABI does not agree");
             }
+
+            Some(caller_ret_place)
         } else {
             // FIXME: Can we truly accept any caller ABI if the callee does not return anything?
-        }
+            None
+        };
```
memoryleak47, line 385: I know this feels a bit like a `ret.map`, but there is no `Option::try_map` combinator to account for those `?`.
You know a work-around?

lang/step.md
```diff
@@ -414,6 +428,25 @@ impl<M: Memory> Machine<M> {
...
+```rust
+impl<M: Memory> Machine<M> {
+    fn eval_terminator(
+        &mut self,
+        Terminator::CallIntrinsic { intrinsic, arguments, ret, next_block }: Terminator
+    ) -> NdResult {
+        let ret = match ret {
+            Some(ret) => Some(self.eval_place(ret)?),
+            None => None,
+        };
```
memoryleak47, line 442: Another case of severe `Option::try_map` demand. ^^

lang/step.md
RalfJung: This looks mighty complicated and I don't understand the motivation for doing this. I'd not couple things together like that. Instead just cause UB when information is missing that you absolutely need at that moment.

lang/step.md
```diff
@@ -395,9 +406,12 @@ impl<M: Memory> Machine<M> {
...
         // Advance the PC for this stack frame.
-        self.mutate_cur_frame(|frame| {
-            frame.jump_to_block(next_block);
-        });
+        if fn_might_return {
+            let next_block = next_block.unwrap();
+            self.mutate_cur_frame(|frame| {
+                frame.jump_to_block(next_block);
+            });
+        }
```
RalfJung, line 414: Okay here we have a problem. If `next_block` is `None` we cannot jump to the next block already. And we don't want UB *now* in that case, we want UB only if the callee ever returns.

In Miri this works because we only jump to the next frame when the callee returns. Maybe we should do that here, too?

lang/step.md
RalfJung: Don't you define your own `Option` in specr anyway? Might as well add a `try_map` there?^^

memoryleak47: Not currently, but I could do that ^^

lang/intrinsic.md
```diff
@@ -0,0 +1,49 @@
...
+
+```rust
+pub enum Intrinsic {
+    Exit,
+    PrintStdout,
+    PrintStderr,
+}
+
+impl<M: Memory> Machine<M> {
+    fn eval_intrinsic(
```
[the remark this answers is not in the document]
RalfJung: Hm, I think I'd be in favor of making this an argmatch function though. As the number of intrinsic grows, that should make things more manageable.

lang/intrinsic.md
RalfJung: I think accepting both is fine, and you don't need to write anything for a ZST return value. A copy of 0 bytes is a NOP anyway.

PR #68, opened 2022-12-05
make sure our code is warning-free

RalfJung:


RalfJung:
@memoryleak47 would be nice to get this to pass... :)

memoryleak47:
As of now we only have
`function ´bytes_valid_for_type´ is never used`
`function ´transmute´ is never used`

I think those functions are intentionally unused.
Do we want to make them `pub`, or add some `#[allow(dead_code)`?

RalfJung:
Yeah makes sense to mark them unused. They are just for demonstration.

Can you open a PR for that, also including the commit from this PR?

RalfJung:
Superseded by https://github.com/RalfJung/minirust/pull/73

PR #69, opened 2022-12-07
optional return

memoryleak47:


ci.sh
```diff
@@ -2,7 +2,7 @@
 set -ex
 
 # Fixed `minirust-tooling` commit, we need to bump this occasionally.
-TOOLING_COMMIT="d2a4cec"
+TOOLING_COMMIT="a0a4a5b"
```
memoryleak47, line 5: required for `Option::try_map`

lang/machine.md
```diff
@@ -41,27 +41,36 @@ struct StackFrame<M: Memory> {
...
     next_stmt: Int,
 }
+
+struct CallerReturnInfo<M: Memory> {
+    /// The basic block to jump to when the callee returns.
+    /// If `None`, UB will be raised when the callee returns.
+    next_block: Option<BbName>,
+    /// The place where the caller wants to see the return value.
+    /// If `None`, the return value will be discarded.
+    place: Option<Place<M>>
```
bjorn3, line 51: Codegen backends handle this using a dummy stack slot. Ignoring writes is a bit awkward as we allow for a return type `(u8, !)` to do `_0.1 = foo();`, which needs some room to store the return value and going the ignore route would require propagating the blackhole place to this call, which requires generating an extra copy of `foo` specifically for the case where the return value is ignored.

lang/step.md
```diff
@@ -360,18 +360,22 @@ impl<M: Memory> Machine<M> {
...
+                    throw_ub!("call ABI violation: return ABI does not agree");
+                }
+            } else {
+                // FIXME: Can we truly accept any caller ABI if the callee does not return anything?
             }
-        } else {
-            // FIXME: Can we truly accept any caller ABI if the callee does not return anything?
-        }
+
+            return caller_ret_place;
```
memoryleak47, line 377: This `return caller_ret_place;` instead of `caller_ret_place`,
because `specr` otherwise doesn't wrap it with `specr::hidden::ret(_)`.

(wrapping closure bodies in `ret` causes a few type inference problems, sadly)

lang/machine.md
RalfJung: @bjorn3 note that MiniRust differs from codegen/Miri in that on a call, we always allocate a return place for the callee. The caller-provided return place only becomes relevant when the function returns.

So using your terms, I think we always have such a "dummy stack slot". If the place here is `None`, that means the contents of that stack slot are discarded upon return; if it is `Some`, that means the contents are copied to the caller-provided place.

lang/step.md
```diff
@@ -422,14 +424,18 @@ impl<M: Memory> Machine<M> {
...
         // would never ensure that the value is valid at that type.
-        let ret_pty = func.locals[ret_local];
-        let ret_val = self.mem.typed_load(frame.locals[ret_local], ret_pty)?;
-        self.mem.typed_store(frame.caller_ret_place, ret_val, ret_pty)?;
+        if let Some(ret_place) = frame.caller_return_info.place {
+            let ret_pty = func.locals[ret_local];
+            let ret_val = self.mem.typed_load(frame.locals[ret_local], ret_pty)?;
+            self.mem.typed_store(ret_place, ret_val, ret_pty)?;
+        }
+
```
RalfJung, line 438: This function also needs to jump to `frame.caller_return_info.next_block`.

lang/syntax.md
```diff
@@ -86,9 +86,9 @@ pub enum Terminator {
         /// The arguments to pass, and which ABIs to use for that.
         arguments: List<(ValueExpr, ArgAbi)>,
         /// The place to put the return value into, and which ABI to use for that.
-        ret: (PlaceExpr, ArgAbi),
+        ret: Option<(PlaceExpr, ArgAbi)>,
```
RalfJung, line 89:
```suggestion
        /// If `None`, the function's return value will be discarded.
        ret: Option<(PlaceExpr, ArgAbi)>,
```

lang/syntax.md
```diff
@@ -86,9 +86,9 @@ pub enum Terminator {
...
         /// The block to jump to when this call returns.
-        next_block: BbName,
+        next_block: Option<BbName>,
```
RalfJung, line 91:
```suggestion
        /// If `None`, UB will be raised when the function returns.
        next_block: Option<BbName>,
```

lang/well-formed.md
```diff
@@ -324,9 +324,12 @@ impl Terminator {
...
                 }
-                let (ret_place, _ret_abi) = ret;
-                ret_place.check_wf::<M>(live_locals)?;
-                list![next_block]
+
+                if let Some((ret_place, _ret_abi)) = ret {
+                    ret_place.check_wf::<M>(live_locals)?;
+                }
+
+                next_block.into_iter().collect()
```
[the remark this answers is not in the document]
RalfJung: Yeah...

lang/machine.md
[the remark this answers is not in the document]
memoryleak47: Hm.. feels a bit redundant `CallerReturnInfo::ret_place`, but okay.

lang/step.md
```diff
@@ -360,17 +360,25 @@ impl<M: Memory> Machine<M> {
...
+        })?;
+
         // Create place for return local, if needed.
-        if let Some((ret_local, callee_ret_abi)) = func.ret {
+        if let Some((ret_local, _abi)) = func.ret {
             let callee_ret_layout = func.locals[ret_local].layout::<M>();
             locals.insert(ret_local, self.mem.allocate(callee_ret_layout.size, callee_ret_layout.align)?);
+        }
+
+        // check ABI compatibility
```
RalfJung, line 375:
```suggestion
        // Check ABI compatibility.
```

lang/step.md
```diff
@@ -422,20 +428,32 @@ impl<M: Memory> Machine<M> {
...
             let layout = func.locals[local].layout::<M>();
             self.mem.deallocate(place, layout.size, layout.align)?;
         }
+
+        if let Some(next_block) = frame.caller_return_info.next_block {
+            self.mutate_cur_frame(|frame| {
+                frame.jump_to_block(next_block);
+            });
+        } else {
+            throw_ub!("Function returned without specifying next block");
```
RalfJung, line 455:
```suggestion
            throw_ub!("Return from a function where caller did not specify next block");
```

lang/step.md
```diff
@@ -360,17 +360,26 @@ impl<M: Memory> Machine<M> {
         let mut locals: Map<LocalName, Place<M>> = Map::new();
 
         // First evaluate the return place and remember it for `Return`. (Left-to-right!)
-        let (caller_ret_place, caller_ret_abi) = ret;
-        let caller_ret_place = self.eval_place(caller_ret_place)?;
+        let ret_place = ret.try_map(|(caller_ret_place, _abi)| {
+            let caller_ret_place = self.eval_place(caller_ret_place)?;
+
+            // FIXME: spec Ok-wrapping of return expressions does not work in closures.
+            return caller_ret_place;
```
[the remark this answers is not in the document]
memoryleak47: @RalfJung alright, time for the next iteration ^^

PR #70, opened 2022-12-08
add Intrinsics

memoryleak47:


lang/step.md
```diff
@@ -459,3 +459,29 @@ impl<M: Memory> Machine<M> {
...
+```rust
+impl<M: Memory> Machine<M> {
+    fn eval_terminator(
+        &mut self,
+        Terminator::CallIntrinsic { intrinsic, arguments, ret, next_block }: Terminator
+    ) -> NdResult {
+        let ret = ret.try_map(|p| self.eval_place(p))?;
+
+        // Evaluate all arguments.
+        let arguments = arguments.iter().map(|arg| self.eval_value(arg)).try_collect()?;
```
RalfJung, line 474: Should we add `try_map` on lists? I think this is a common pattern, isn't it?

lang/intrinsic.md
```diff
@@ -0,0 +1,73 @@
+# Intrinsics
+
+```rust
+pub enum Intrinsic {
+    Exit,
+    PrintStdout,
+    PrintStderr,
+}
```
RalfJung, line 8: This enum is part of the syntax, should be in `syntax.md`.

lang/intrinsic.md
```diff
@@ -0,0 +1,73 @@
...
+
+```rust
+impl<M: Memory> Machine<M> {
+    fn eval_intrinsic(
+        &mut self,
+        Intrinsic::Exit: Intrinsic,
+        arguments: List<Value<M>>,
+        ret: Option<Place<M>>,
+    ) -> NdResult {
+        throw_machine_stop!("0")
```
RalfJung, line 31: Didn't you said you'd remove the string?

lang/intrinsic.md
```diff
@@ -0,0 +1,73 @@
...
+        Intrinsic::PrintStderr: Intrinsic,
+        arguments: List<Value<M>>,
+        ret: Option<Place<M>>,
+    ) -> NdResult {
+        self.eval_print(&mut std::io::stderr(), arguments, ret)?
+    }
+
+    fn eval_print(
+        &mut self,
+        stream: &mut impl std::io::Write,
```
RalfJung, line 60: Oh the Coq translator will not like this type... but I guess we can figure this out when we get there. (Probably we'll have to add print/eprint functions to specr so that different specr backends can have their own implementation.)

lang/step.md
```diff
@@ -459,3 +459,29 @@ impl<M: Memory> Machine<M> {
...
+
+### Intrinsic
+
+```rust
+impl<M: Memory> Machine<M> {
+    fn eval_terminator(
+        &mut self,
+        Terminator::CallIntrinsic { intrinsic, arguments, ret, next_block }: Terminator
+    ) -> NdResult {
+        let ret = ret.try_map(|p| self.eval_place(p))?;
```
RalfJung, line 471:
```suggestion
        // First evaluate return place (left-to-right evaluation).
        let ret = ret.try_map(|p| self.eval_place(p))?;
```

PR #71, opened 2022-12-09
intrinsics follow-up

memoryleak47:


ci.sh
```diff
@@ -2,7 +2,7 @@
 set -ex
 
 # Fixed `minirust-tooling` commit, we need to bump this occasionally.
-TOOLING_COMMIT="a0a4a5b"
+TOOLING_COMMIT="5c4c69a"
```
memoryleak47, line 5: for `List::try_map`

lang/values.md
```diff
@@ -103,6 +103,7 @@ impl Type {
             throw!();
         }
         // Fails if any byte is `Uninit`.
+        // FIXME: We want a List::try_map supporting options.
```
memoryleak47, line 106: so, currently `List::try_map` only works for `NdResult` (same for `Option::try_map`).

I originally implemented a general `Option::try_map` function, but iirc it wasn't able to infer types, and was thus a bit unpractical.

@RalfJung I presume you'd prefer a general `.try_map` impl.
If you want to, I can revisit my original impl

RalfJung: Yeah would be nice if it could be as general as `try_collect`.

lang/step.md
```diff
@@ -361,8 +359,7 @@ impl<M: Memory> Machine<M> {
 
         // First evaluate the return place and remember it for `Return`. (Left-to-right!)
         let ret_place = ret.try_map(|(caller_ret_place, _abi)| {
-            // FIXME: spec Ok-wrapping of return expressions does not work in closures.
-            return self.eval_place(caller_ret_place)?;
```
memoryleak47, line 365: Removing the `?` also does the trick, then the `specr::hidden::ret(_)` isn't even required.

lang/values.md
```diff
@@ -466,9 +466,9 @@ impl<M: Memory> TypedMemory for M {
             (Value::Ptr(ptr), Type::Ptr(ptr_type)) => Value::Ptr(self.retag_ptr(ptr, ptr_type, fn_entry)?),
             // recurse into tuples/arrays/enums
             (Value::Tuple(vals), Type::Tuple { fields, .. }) =>
-                Value::Tuple(vals.iter().zip(fields).map(|(val, (_offset, ty))| self.retag_val(val, ty, fn_entry)).try_collect()?),
```
memoryleak47, line 469: I added this `List::zip`.
Generally I think all those `.iter()` calls make little sense for a spec.

memoryleak47:
@RalfJung ready

lang/values.md
memoryleak47: I'm happy to open another PR to get rid of all unnecessary `.iter()` calls, by implementing some `List` methods.
You agree?

lang/values.md
memoryleak47: Worked out using the `Residual` trait

lang/step.md
RalfJung: That's a bug though, it should not work without the `?` -- this shows that Ok-wrapping is not happening.

I guess we are already doing that for a lot of other closures though... so I guess fine to do it here, but something we should revisit at some point. Right now we have Ok-wrapping for functions but not closures and that is somewhat odd.

lang/values.md
RalfJung: What is the type of all these functions?

Of course we could avoid iterators entirely, and have a bunch of functions that just create lists. But if that means creating more intermediate lists, that sounds bad for the interpreter -- we don't want `list1.zip(list2).try_map(...)` to construct an intermediate list, I think.

So what are the operations we need?
- `fn try_map(List<T>, impl Fn(T) -> Result<U, E>) -> Result<List<U>, E>`
- `fn try_zip_map(List<T1>, List<T2>, impl Fn(T1, T2) -> Result<U, E>) -> Result<List<U>, E>`

anything else?

lang/step.md
```diff
@@ -361,8 +359,7 @@ impl<M: Memory> Machine<M> {
...
+            self.eval_place(caller_ret_place)
         })?;
```
RalfJung, line 363: This could probably be one line now?
```
let ret_place = ret.try_map(|(caller_ret_place, _abi)| self.eval_place(caller_ret_place))?;
```

lang/step.md
memoryleak47: Yeah, it's imperfect - I agree.

Note that it working without `?` is not a bug.
`specr::hidden::ret(_)` will wrap a `T` to become `NdResult<T, _>`, but will just pass through a `NdResult<T, _>` without wrapping anything.

Previously I added `specr::hidden::ret(_)` too for closures, iirc it required a few more type annotations to make minirust compile then. I think I'll add the type annotations and make a PR.

RalfJung: > specr::hidden::ret(_) will wrap a T to become NdResult<T, _>, but will just pass through a NdResult<T, _> without wrapping anything.

It does? That sounds like a bug, too.^^ Also how is it doing that?

memoryleak47: > That sounds like a bug, too.^^

Oh, really?
It was what I intended, for what its worth ^^

Here's how it works:
https://github.com/memoryleak47/minirust-tooling/blob/main/libspecr/src/hidden/ret.rs

Alternatively we could do
```rust
trait MonadicReturn {
  type Inner;
  fn monadic_return(t: Self::Inner) -> Self;
}

impl<T, E> MonadicReturn for NdResult<T, E> {
  type Inner = T;
  fn monadic_return(t: T) -> Self { NdResult(Ok(t)) }
}
// similarly for the other monads
```
that will probably solve some of the inference problems, but then we don't get the passthrough for `NdResult<_, _>`.

lang/values.md
memoryleak47: > we don't want list1.zip(list2).try_map(...) to construct an intermediate list, I think.

Interesting, I expected you'd prefer the conciseness over the performance benefit from using iterators in that regard.

Then I guess doing `List::zip(impl IntoIterator) -> impl Iterator` and implementing `IteratorExt::try_map` for an extension trait for `Iterator`s would be the best of both worlds?

Then the code stays `list1.zip(list2).try_map` but it's quick.

lang/step.md
RalfJung: How on earth does Rust even accept those instances? Doesn't `MonadicReturn<T> for T` overlap with `MonadicReturn<T> for Option<T> `??

Anyway my intention was that in a `Result<T, E>`-returning function, I *have to* give a return expression of type `T`. That is how `try` blocks work in Rust, so this can mostly be implemented by wrapping all function bodies in a `try`, except something has to be done for early `return`s.

(And to be clear that is not a blocker for this PR since it is pre-existing, I just noticed this only now.)

memoryleak47: > How on earth does Rust even accept those instances? Doesn't MonadicReturn<T> for T overlap with MonadicReturn<T> for Option<T> ??

`Option<T>` just implements both `MonadicReturn<T>` and `MonadicReturn<Option<T>>`, those impls do not collide.

memoryleak47: > Anyway my intention was that in a Result<T, E>-returning function, I have to give a return expression of type T.

I see, I think the snippet above "Alternatively we could do" should be able to express that.
I'll open another PR for it.

lang/values.md
memoryleak47: For now I reverted my change here. We'll do those things in another PR.

memoryleak47:
@RalfJung ready

lang/values.md
RalfJung: > For now I reverted my change here. We'll do those things in another PR.

Okay.

If `try_zip_map` is the only thing we will need to completely remove iterators from MiniRust, it might be worth it, despite being rather ad-hoc.

lang/step.md
RalfJung: > I see, I think the snippet above "Alternatively we could do" should be able to express that.
I'll open another PR for it.

I think that will fail due to overlapping instances but I might be too pessimistic.

RalfJung: Don't you still need some fallback instance for all the functions that return neither Option nor Result?

memoryleak47: Yep, that would only work if we are able to insert `ret` to only those functions that actually return an `NdResult` or something similar.

PR #72, opened 2022-12-09
fix pick

memoryleak47:
This requires a follow-up PR to update the README to this new `pick` API

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -2,7 +2,7 @@
 set -ex
 
 # Fixed `minirust-tooling` commit, we need to bump this occasionally.
-TOOLING_COMMIT="a0a4a5b"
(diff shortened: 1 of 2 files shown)


ci.sh
```diff
@@ -2,7 +2,7 @@
 set -ex
 
 # Fixed `minirust-tooling` commit, we need to bump this occasionally.
-TOOLING_COMMIT="a0a4a5b"
+TOOLING_COMMIT="7531357"
```
memoryleak47, line 5: This will collide with the intrinsics-followup PR.
I'd prefer this one to go through quicker.

RalfJung:
> This requires a follow-up PR to update the README to this new pick API

Indeed. :)

PR #79, opened 2022-12-14
drop implicit ret-wrapping

memoryleak47:
This mirrors changes from https://github.com/memoryleak47/minirust-tooling/commit/08f6874162c3c8fb208733ac589e030a8ec390b7.


RalfJung:
The README also needs to be adjusted to the new reality.

PR #82, opened 2022-12-26
Array `count` needs to be non-negative

memoryleak47:


RalfJung:
Yeah that would be better. :)

PR #83, opened 2022-12-27
do return value copy at caller type, not callee type

RalfJung:


RalfJung:
@memoryleak47 not entirely sure what happens here (sadly the error is not really readable since it refers to the generated file), but something seems to go wrong with the `let ret_place` that this PR changes? Why would that have type `Option<()>`, that makes no sense?

memoryleak47:
> `ret((p, pty)) as NdResult`

You are missing generic arguments to NdResult.
It's defined as `pub type NdResult<T=()> = specr::NdResult<T, TerminationInfo>;` and hence it defaults to `()`.

RalfJung:
:facepalm:

PR #86, opened 2023-01-09
fix unnecessary braces

memoryleak47:


RalfJung:
I added these deliberately. I think `match`es are easier to read when all arms are formatted consistently -- either all with braces, or all without braces.

PR #87, opened 2023-01-09
fix array decoding

memoryleak47:


ci.sh
```diff
@@ -2,7 +2,7 @@
 set -ex
 
 # Fixed `minirust-tooling` commit, we need to bump this occasionally.
-TOOLING_COMMIT="08f6874"
+TOOLING_COMMIT="4a86d97"
```
memoryleak47, line 5: required for `Int::ZERO..count`

PR #88, opened 2023-02-01
implement Intrinsic::{Allocate, Deallocate}

memoryleak47:


lang/intrinsic.md
```diff
@@ -71,3 +71,68 @@ impl<M: Memory> Machine<M> {
...
+        let size = Size::from_bytes(size);
+
+        let Value::Int(align) = arguments[1] else {
+            throw_ub!("invalid second argument to `Intrinsic::Allocate`");
+        };
+        let align = Align::from_bytes(align);
+
+        let alloc = self.mem.allocate(size, align)?;
+        let bytes = encode_ptr::<M>(alloc);
+        self.mem.store(ret_place, bytes, M::PTR_ALIGN)?;
```
RalfJung, line 104:
```suggestion
        // FIXME: mark this as being distinct from stack allocations, to detect
        // (de)allocation mixups.
        self.mem.store(ret_place, bytes, M::PTR_ALIGN)?;
```

RalfJung:
Implementation LGTM, but now that was have a test suite... wouldn't it be nice to have a test in the same PR so that CI ensures that this indeed works? :)

memoryleak47:
Hm.. the disadvantage of having the test suite here instead of in minirust-tooling was that the workflow is quite a bit slower due to PRs. I thought we talked this through in the last meeting, and decided that keeping it external was more practical for now?

RalfJung:
Oh, I somehow thought we had tests in this repo already. I forgot that CI only checks whether minirust builds. Sorry for that.

PR #89, opened 2023-02-05
replace Constant::Tuple with ValueExpr::Tuple

memoryleak47:


lang/syntax.md
```diff
@@ -145,6 +145,10 @@ And finally, the syntax of expressions:
 pub enum ValueExpr {
     /// Just return a constant value.
     Constant(Constant, Type),
+
+    /// An n-tuple, used for arrays, structs, tuples (including unit).
+    Tuple(List<ValueExpr>, Type),
```
RalfJung, line 150: Why does this come with a `Type`? Looking at `check_wf`, that seems like redundant information?

memoryleak47: That's correct for the inner types, but the information on whether this represents Array or Tuple (and what the tuple offsets are) cannot be deduced from the ValueExpr.

RalfJung: Ah... well that is unfortunate. Generally I'm wondering if there isn't e better setup for this typing stuff that avoids some of the redundancy... but I've already overhauled it once and that only helped so much. :/

PR #92, opened 2023-02-07
Size::from_* returns Option

memoryleak47:
Size::from_* also previously had a panic if the input is negative.
Good that we caught that now, as this would have caused a panic with allocate / deallocate.

Annoying thing is that we end up having quite a few more unwraps.

### DIFF
--- a/mem/basic.md
+++ b/mem/basic.md
@@ -49,7 +49,7 @@ The model represents a 64-bit little-endian machine.
 
 ```rust
 impl Memory for BasicMemory {
-    const PTR_SIZE: Size = Size::from_bits_const(64);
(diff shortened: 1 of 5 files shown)


mem/basic.md
```diff
@@ -204,7 +204,7 @@ impl BasicMemory {
             throw_ub!("out-of-bounds memory access");
         }
         // All is good!
-        ret(Some((id, Size::from_bytes(offset_in_alloc))))
+        ret(Some((id, Size::from_bytes(offset_in_alloc).unwrap())))
```
memoryleak47, line 207: A few lines above, we asserted that offset_in_loc is not negative, so this unwrap should be safe too.

PR #96, opened 2023-02-13
add ValueExpr::Union

memoryleak47:
Maybe de-aggregating in minimize is the better solution?

### DIFF
--- a/lang/syntax.md
+++ b/lang/syntax.md
@@ -151,7 +151,7 @@ pub enum ValueExpr {
 
     /// A `Union` value.
     Union {
-        /// The unions field which will be initialized.
+        /// The union's field which will be initialized.
         field: Int,
         /// The value it will be initialized with.
         #[specr::indirection]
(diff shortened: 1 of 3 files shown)


lang/well-formed.md
```diff
@@ -159,10 +159,14 @@ impl ValueExpr {
         use ValueExpr::*;
         ret(match self {
             Constant(value, ty) => {
+                ty.check_wf::<M>()?;
+
                 value.check_wf(ty)?;
                 ty
             }
             Tuple(exprs, t) => {
+                t.check_wf::<M>()?;
```
memoryleak47, line 168: Are these checks redundant?

lang/syntax.md
```diff
@@ -149,6 +149,17 @@ pub enum ValueExpr {
     /// An n-tuple, used for arrays, structs, tuples (including unit).
     Tuple(List<ValueExpr>, Type),
 
+    /// A `Union` value.
+    Union {
+        /// The unions field which will be initialized.
```
RalfJung, line 154:
```suggestion
        /// The union's field which will be initialized.
```

lang/well-formed.md
RalfJung: Good question. Better safe than sorry. :)

RalfJung (approved):
Looks good!

PR #97, opened 2023-02-17
const ptrs

memoryleak47:


lang/syntax.md
```diff
@@ -278,6 +283,27 @@ pub enum PlaceExpr {
         index: ValueExpr,
     },
 }
+
+/// A template for a global allocation.
+pub struct Global {
+    /// The raw bytes of the allocation.
```
RalfJung, line 289:
```suggestion
    /// The raw bytes of the allocation. `None` represents uninitialized bytes.
```

lang/syntax.md
```diff
@@ -127,6 +130,8 @@ pub enum Constant {
     Int(Int),
     /// A Boolean value, used for `bool`.
     Bool(bool),
+    /// A pointer pointing to a global allocation with a given offset.
+    Pointer(GlobalName, Size),
```
RalfJung, line 134: This could be a `Relocation` as well, right?

lang/machine.md
```diff
@@ -57,30 +60,58 @@ Next, we define how to create a machine.
 
 ```rust
 impl<M: Memory> Machine<M> {
-    pub fn new(prog: Program) -> Option<Machine<M>> {
+    pub fn new(prog: Program) -> Option<NdResult<Machine<M>>> {
```
RalfJung, line 63: This is a strange return type. If we truly need an NdResult here we probably want to also represent check_wf failure with an Err, not a None.

lang/machine.md
```diff
@@ -57,30 +60,58 @@ Next, we define how to create a machine.
...
+        ret(try {
+            let start_fn = prog.functions[prog.start];
+
+            // Setup the initial stack frame.
+            // Well-formedness ensures that this has neither arguments nor a return local.
+            let init_frame = StackFrame {
+                func: start_fn,
+                locals: Map::new(),
+                caller_return_info: CallerReturnInfo {
+                    // the start function should never return.
```
RalfJung, line 75:
```suggestion
                    // The start function should never return.
```

lang/machine.md
```diff
@@ -57,30 +60,58 @@ Next, we define how to create a machine.
...
+                    next_block: None,
+                    ret_place: None,
+                },
+                next_block: start_fn.start,
+                next_stmt: Int::ZERO,
+            };
+            let mut mem = M::new();
+            let mut global_ptrs = Map::new();
+
+            // allocate every global.
```
RalfJung, line 85:
```suggestion
            // Allocate every global.
```

lang/machine.md
```diff
@@ -57,30 +60,58 @@ Next, we define how to create a machine.
...
+            for (global_name, global) in prog.globals {
+                let size = Size::from_bytes(global.bytes.len()).unwrap();
+                let alloc = mem.allocate(size, global.align)?;
+                global_ptrs.insert(global_name, alloc);
+            }
+
+            // fill the allocations.
```
RalfJung, line 92:
```suggestion
            // Fill the allocations.
```

lang/syntax.md
```diff
@@ -14,11 +14,14 @@ pub struct Program {
     pub functions: Map<FnName, Function>,
     /// The function where execution starts.
     pub start: FnName,
+    /// Associate each global name with the associated global.
+    pub globals: Map<GlobalName, Global>,
```
RalfJung, line 18: There needs to be some wf checking for this. At least all Globalname occurring everywhere must actually be present in this list. And probably in a Global, we want to make sure that the relocations are in-bounds? If `bytes` has length 4 then a relocation at offset 2 doesn't work out... that might even lead to panics in `Machine::new`? And finally, we might want to sanity-check the offset on the relocations and const pointers: maybe we want to require it to be in-bounds of the global it points to? That should then ensure that `Machine::new` is never UB due to pointer arithmetic overflow.

memoryleak47:
@RalfJung ready

lang/machine.md
```diff
@@ -68,19 +73,45 @@ impl<M: Memory> Machine<M> {
...
+                    None => AbstractByte::Uninit
+                }
+            );
+            for (i, relocation) in global.relocations {
+                let ptr = global_ptrs[relocation.name].offset::<M>(relocation.offset).unwrap();
+                let encoded_ptr = encode_ptr::<M>(ptr);
+                bytes.write_subslice_at_index(i.bytes(), encoded_ptr);
+            }
+            mem.store(global_ptrs[global_name], bytes, global.align)?;
+        }
```
RalfJung, line 107: Nit: it'd make sense to do this before setting up the first stack frame, since arguably globals are "more global" than the stack.

lang/well-formed.md
```diff
@@ -471,6 +471,29 @@ impl Program {
...
+        ret(())
+    }
+
+    // Checks whether the relocation is within bounds.
+    fn check_relocation(self, relocation: Relocation) -> Option<()> {
+        // The global where are pointing to needs to exist.
+        let Some(global) = self.globals.get(relocation.name) else { throw!() };
+        let size = Size::from_bytes(global.bytes.len()).unwrap();
+
+        // And the offset needs to be in-bounds of it's size.
```
RalfJung, line 494:
```suggestion
        // And the offset needs to be in-bounds of its size.
```

mem/interface.md
```diff
@@ -132,3 +132,20 @@ This is a very basic memory interface that is incomplete in at least the followi
 * Maybe we want operations that can compare pointers without casting them to integers. Or else we decide only the address can matter for comparison.
 
 [`Ordering`]: https://doc.rust-lang.org/nightly/core/sync/atomic/enum.Ordering.html
+
+```rust
+impl<Provenance> Pointer<Provenance> {
+    /// Offsets the `self` pointer.
+    /// Returns `Ǹone` if the resulting pointer does not fit into `PTR_SIZE`.
+    /// This does not check whether the offset stays in bounds of it's allocation.
```
RalfJung, line 140:
```suggestion
    /// This does not check whether the offset stays in bounds of its allocation.
```

mem/interface.md
```diff
@@ -132,3 +132,20 @@ This is a very basic memory interface that is incomplete in at least the followi
...
+    pub fn offset<M: Memory<Provenance=Provenance>>(self, offset: Size) -> Option<Self> {
```
RalfJung, line 141: It's a bit strange that this is only used for globals... I think you can make this `wrapping_offset` and change the return type to just `Self` and then use it for ptr_offset_wrapping ?

lang/syntax.md
```diff
@@ -127,6 +130,8 @@ pub enum Constant {
     Int(Int),
     /// A Boolean value, used for `bool`.
     Bool(bool),
+    /// A pointer pointing into a global allocation with a given offset.
+    Pointer(Relocation),
```
RalfJung, line 134: I don't see a WF check for this new `Constant` variant?

RalfJung (commented):
mostly just nits, but I think one WF check is still missing

memoryleak47:
@RalfJung ready.

Btw, as `Constant::check_wf` now requires `globals`, this is passed through quite a lot.
Maybe there is a cleaner way to do it.

RalfJung:
> Btw, as Constant::check_wf now requires globals, this is passed through quite a lot.
Maybe there is a cleaner way to do it.

Yeah @EnderGnar already made the same comment for the function pointers... we might want to find some solution for this, but that's a separate PR.

PR #99, opened 2023-02-23
new specr CI

memoryleak47:


ci.sh
```diff
@@ -1,18 +1,11 @@
...
 set -ex
 
-# Fixed `minirust-tooling` commit, we need to bump this occasionally.
-TOOLING_COMMIT="b1b41a1"
+# Fixed specr-transpile version
+VERSION="0.1.3"
 
-# where to check out the tooling
-TOOLING_DIR="$HOME/minirust-tooling"
+rustup default nightly-2023-02-12
```
memoryleak47, line 7: `specr-transpile` needs this. Is there a way to fix the rust nightly version inside of the specr-transpile crate?
rust-toolchain.toml is ignored by crates.io

ci.sh
```diff
@@ -1,18 +1,10 @@
...
-git clone "https://github.com/memoryleak47/minirust-tooling" "$TOOLING_DIR"
-ln -s "$PWD" "$TOOLING_DIR"/minirust
-
-cd "$TOOLING_DIR"
-git checkout "$TOOLING_COMMIT"
-
-# transpile, and build the transpiled result
-(cd specr-transpile; cargo run)
-(cd gen-minirust; RUSTFLAGS="-D warnings" cargo build)
+cd gen-minirust; RUSTFLAGS="-D warnings" cargo build
```
RalfJung, line 10:
```suggestion
(cd gen-minirust; RUSTFLAGS="-D warnings" cargo build)
```
or have it in two lines? not sure why you made this one line

specr.toml
```diff
@@ -0,0 +1,13 @@
+input = "."
```
RalfJung, line 1: How does this define which files to use? If it's just "all markdown files recursively" then we probably want to move this all into a subdir at some point, so that it doesn't pick up things e.g. from inside minimize.

ci.sh
RalfJung: Nice. :)

specr.toml
memoryleak47: Agreed, I think when minisyntax & minitest will be merged into this repo, it's a good time to add the `spec` and `tooling` subdirectories.

PR #100, opened 2023-02-27
Fn ptrs

EnderGnar:
Changed the call system to go over addresses. This allows for operations with function pointers.


RalfJung:
https://github.com/RalfJung/minirust/pull/97 has landed, can you do a rebase? Looks like there are quite a few conflicts unfortunately, though if you skip the const ptrs commit for the rebase hopefully most of them disappear.

mem/basic.md
```diff
@@ -74,7 +75,7 @@ impl Allocation {
     fn overlaps(self, other_addr: Int, other_size: Size) -> bool {
         let end_addr = self.addr + self.size().bytes();
         let other_end_addr = other_addr + other_size.bytes();
-        if end_addr <= other_addr || other_end_addr <= self.addr {
+        if self.addr != other_addr && (end_addr <= other_addr || other_end_addr <= self.addr) {
             // Our end is before their beginning, or vice versa -- we do not overlap.
```
RalfJung, line 79:
```suggestion
            // Our end is before their beginning, or vice versa -- we do not overlap.
            // However, also make sure that each allocation has a unique address.
            // FIXME: This is not necessarily realistic, e.g. for zero-sized stack variables.
            // OTOH the function pointer logic currently relies on this.
```

lang/machine.md
```diff
@@ -68,19 +82,59 @@ impl<M: Memory> Machine<M> {
...
+                let ptr = global_ptrs[relocation.name].offset::<M>(relocation.offset).unwrap();
+                let encoded_ptr = encode_ptr::<M>(ptr);
+                bytes.write_subslice_at_index(i.bytes(), encoded_ptr);
+            }
+            mem.store(global_ptrs[global_name], bytes, global.align)?;
+        }
+
+        for (fn_name, function) in prog.functions {
+            let alloc = mem.allocate(Size::ZERO, M::FN_ALIGN)?;
+            let addr = alloc.addr;
```
RalfJung, line 123: Here it might be good to `assert!` that `addr` is not yet assigned to a function.

lang/machine.md
```diff
@@ -20,6 +20,18 @@ pub struct Machine<M: Memory> {
...
+    global_ptrs: Map<GlobalName, Pointer<M::Provenance>>,
+
+    /// Stores an address for each function name.
+    fn_addrs: Map<FnName, mem::Address>,
+
+    /// Maps address of functions to the actual function
+    fn_map: Map<mem::Address, Function>,
+
+    /// The names of all functions defined. Used for Constant wellformedness checking.
+    fn_names: Set<FnName>,
```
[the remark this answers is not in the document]
EnderGnar: Rebase: `check_wf` changed all `globals` to `prog` and removed `fn_names` completely.

lang/machine.md
```diff
@@ -20,6 +20,18 @@ pub struct Machine<M: Memory> {
...
     stack: List<StackFrame<M>>,
+
+    /// Stores a pointer to each of the global allocations.
+    global_ptrs: Map<GlobalName, Pointer<M::Provenance>>,
+
+    /// Stores an address for each function name.
+    fn_addrs: Map<FnName, mem::Address>,
+
+    /// Maps address of functions to the actual function
+    fn_map: Map<mem::Address, Function>,
```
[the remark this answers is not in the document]
EnderGnar: Is there a method `Map.search` or should I just use a for loop to iterate over all of them?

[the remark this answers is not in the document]
EnderGnar: The maps aren't really inverse of each other since one works with `FnName` and the other with `Function`. I think if we look at redundancy then there is redundancy in the set `prog`, `fn_addrs`, and `fn_map`. 

Of course if the Program is needed for anything different than translating `FnName` to `Function` removing `fn_map` is the right choice. 
I prefer to remove `prog` from the Machine, but since I probably overlooked something I commited the removal of `fn_map`.

lang/machine.md
```diff
@@ -91,6 +95,15 @@ impl<M: Memory> Machine<M> {
             mem.store(global_ptrs[global_name], bytes, global.align)?;
         }
 
+        // Allocate functions.
+        for (fn_name, _function) in prog.functions {
+            let alloc = mem.allocate(Size::ZERO, M::FN_ALIGN)?;
+            let addr = alloc.addr;
+            //Ensure that no two functions lie on the same address.
```
RalfJung, line 102:
```suggestion
            // Ensure that no two functions lie on the same address.
```

lang/machine.md
```diff
@@ -91,6 +95,15 @@ impl<M: Memory> Machine<M> {
             mem.store(global_ptrs[global_name], bytes, global.align)?;
         }
 
+        // Allocate functions.
+        for (fn_name, _function) in prog.functions {
+            let alloc = mem.allocate(Size::ZERO, M::FN_ALIGN)?;
```
RalfJung, line 100: Why do you hard-code the size to 0 but the alignment is a parameter? Might as well hard-code both of them at that point. Also I don't think `8` is right, functions do not have any alignment requirements to my knowledge.

lang/step.md
```diff
@@ -394,16 +400,23 @@ impl<M: Memory> Machine<M> {
...
         Terminator::Call { callee, arguments, ret: ret_expr, next_block }: Terminator
     ) -> NdResult {
-        let Some(func) = self.prog.functions.get(callee) else {
+        let Value::Ptr(ptr) = self.eval_value(callee)? else {
+            panic!("call on a non-pointer")
+        };
+
+        let mut funcs = self.fn_addrs.iter().filter(|(_, fn_addr)| *fn_addr == ptr.addr);
+        let Some((func_name, _)) = funcs.next() else {
             throw_ub!("calling non-existing function");
```
RalfJung, line 409:
```suggestion
            throw_ub!("calling an address where there is no function");
```

lang/syntax.md
```diff
@@ -133,6 +133,9 @@ pub enum Constant {
     /// A pointer pointing into a global allocation with a given offset.
     Pointer(Relocation),
 
+    /// A function pointer value, used for named dynamic pointer
+    FnPtr(FnName),
```
RalfJung, line 137:
```suggestion
    /// A pointer pointing to a function.
    FnPointer(FnName),
```
Might also make sense to rename `Constant::Pointer` to `GlobalPointer`?

mem/basic.md
```diff
@@ -250,6 +254,7 @@ impl Memory for BasicMemory {
             lang::PtrType::Ref { pointee, .. } => pointee,
             lang::PtrType::Box { pointee } => pointee,
             lang::PtrType::Raw { pointee } => pointee,
+            lang::PtrType::FnPtr => throw_ub!("dereferencing function pointer"),
```
RalfJung, line 257:
```suggestion
            // Raw and fn ptrs do not have any requirements, skip them.
            lang::PtrType::Raw { .. } | lang::PtrType::FnPtr => return ret(ptr),
```

lang/machine.md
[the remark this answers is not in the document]
RalfJung: > I prefer to remove prog from the Machine

I guess the only part of `prog` that we still need is `prog.functions`, so that would be a possibility, yeah -- good point. Let's not do this in the same PR though.

RalfJung:
Looks great, thanks! Can you squash the commits into fewer (or just one if you prefer)?

EnderGnar:
Squashed it into just one commit.

PR #102, opened 2023-03-04
fix redundant decode_ptr call

memoryleak47:
There is a line `let ptr = decode_ptr::<M>(bytes)?;` just above this.


RalfJung:
Good catch!

PR #104, opened 2023-03-09
Threads

EnderGnar:
Simple implementation of threads.

### DIFF
--- a/spec/lang/intrinsic.md
+++ b/spec/lang/intrinsic.md
@@ -7,8 +7,7 @@ impl<M: Memory> Machine<M> {
         &mut self,
         intrinsic: Intrinsic,
         arguments: List<Value<M>>,
-        ret_place: Option<Place<M>>,
-    ) -> NdResult { .. }
+    ) -> NdResult<(Value<M>, Type)> { .. }
 }
 ```
 
@@ -20,8 +19,7 @@ impl<M: Memory> Machine<M> {
         &mut self,
         Intrinsic::Exit: Intrinsic,
         arguments: List<Value<M>>,
-        ret_place: Option<Place<M>>,
-    ) -> NdResult {
+    ) -> NdResult<(Value<M>, Type)> {
         throw_machine_stop!()
     }
 }
@@ -35,29 +33,26 @@ impl<M: Memory> Machine<M> {
         &mut self,
         Intrinsic::PrintStdout: Intrinsic,
         arguments: List<Value<M>>,
-        ret_place: Option<Place<M>>,
-    ) -> NdResult {
-        self.eval_print(&mut std::io::stdout(), arguments, ret_place)?;
+    ) -> NdResult<(Value<M>, Type)> {
+        self.eval_print(&mut std::io::stdout(), arguments)?;
 
-        ret(())
+        ret(unit())
     }
 
     fn eval_intrinsic(
         &mut self,
         Intrinsic::PrintStderr: Intrinsic,
         arguments: List<Value<M>>,
-        ret_place: Option<Place<M>>,
-    ) -> NdResult {
-        self.eval_print(&mut std::io::stderr(), arguments, ret_place)?;
+    ) -> NdResult<(Value<M>, Type)> {
+        self.eval_print(&mut std::io::stderr(), arguments)?;
 
-        ret(())
+        ret(unit())
     }
 
     fn eval_print(
         &mut self,
         stream: &mut impl std::io::Write,
         arguments: List<Value<M>>,
-        _ret_place: Option<Place<M>>,
     ) -> Result {
         for arg in arguments {
             match arg {
@@ -72,7 +67,7 @@ impl<M: Memory> Machine<M> {
 }
 ```
 
-And finally the intrinsics used for memory allocation and deallocation.
+Next the intrinsics used for memory allocation and deallocation.
 
 ```rust
 impl<M: Memory> Machine<M> {
@@ -80,12 +75,7 @@ impl<M: Memory> Machine<M> {
         &mut self,
         Intrinsic::Allocate: Intrinsic,
         arguments: List<Value<M>>,
-        ret_place: Option<Place<M>>,
-    ) -> NdResult {
-        let Some(ret_place) = ret_place else {
-            throw_ub!("call to `Intrinsic::Allocate` is missing a return place");
-        };
-
+    ) -> NdResult<(Value<M>, Type)> {
         if arguments.len() != 2 {
             throw_ub!("invalid number of arguments for `Intrinsic::Allocate`");
         }
@@ -104,18 +94,22 @@ impl<M: Memory> Machine<M> {
         };
 
         let alloc = self.mem.allocate(size, align)?;
-        let bytes = encode_ptr::<M>(alloc);
-        self.mem.store(ret_place, bytes, M::PTR_ALIGN)?;
 
-        ret(())
+        // An allocate returns *mut ()
+        let pointee = Layout {
+            size: Size::ZERO,
+            align: Align::ONE,
+            inhabited: true,
+        };
+
+        ret((Value::Ptr(alloc), Type::Ptr(PtrType::Raw{pointee})))
     }
 
     fn eval_intrinsic(
         &mut self,
         Intrinsic::Deallocate: Intrinsic,
         arguments: List<Value<M>>,
-        ret_place: Option<Place<M>>,
-    ) -> NdResult {
+    ) -> NdResult<(Value<M>, Type)> {
         if arguments.len() != 3 {
             throw_ub!("invalid number of arguments for `Intrinsic::Deallocate`");
         }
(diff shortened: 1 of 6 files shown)


lang/step.md
```diff
@@ -533,18 +545,28 @@ impl<M: Memory> Machine<M> {
...
+                    throw_ub!("Return from an intrinsic where caller did not specify next block");
+                }
+            },
+            ExecutionStatus::Blocked => {
+                // In the future once this thread is `Enabled` again, we retry this intrinsic.
+                
+                // If an execution was blocked, the thread can't be enabled.
+                // This is a sanity check that all intrinsics are implemented correctly.
+                assert!(self.cur_thread().state != ThreadState::Enabled);
+            },
```
RalfJung, line 565: If I understand the code here correctly, then once the thread gets unblocked, it will run *the same intrinsic again*? That seems wrong, unblocking should make it *return from the intrinsic*.

So, why do you even need `ExecutionStatus`? Can't we do `jump_to_block` in both cases? Just in one case the thread will be blocked so it will take a while until the next statement is executed.

lang/thread.md
```diff
@@ -0,0 +1,162 @@
...
+
+```rust
+// This is the thread manager it has state about the overall execution
+// as well as the execution of a single step.
+pub struct ThreadManager<M: Memory> {
+    // The set of threads.
+    threads: List<Thread<M>>,
+
+    // This is to keep state during a step.
+    active_thread: Option<ThreadId>,
```
RalfJung, line 89:
```suggestion
    /// To avoid passing around the active thread through all the eval_ functions,
    /// we store it globally here.
    active_thread: Option<ThreadId>,
```
What does `None` mean though?

lang/thread.md
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    pub fn terminate_active_thread(&mut self) -> NdResult {
+        let active = self.active_thread.unwrap();
+
+        if active == 0 {
+            // Consistency with previous system, where the bottom
+            // frame in the stack was never allowed to return.
+            throw_ub!("Master thread terminated.");
+        }
```
RalfJung, line 147: This is a strange special case... if we need it anyway we might as well throw a nice MachineStop here?

RalfJung (commented):
not done yet, just as far as I had time for...

lang/step.md
EnderGnar: True. I thought this will be useful for locks, because I can simply wake all threads and put them back to sleep. But it would be easier to just wake an arbitrary.

RalfJung: If we go down that route, we don't need a 'blocked' state at all, we can just make Join spin until the target thread is finished.

Properly blocking threads is more realistic though so let's go for that.

lang/intrinsic.md
```diff
@@ -140,7 +152,93 @@ impl<M: Memory> Machine<M> {
...
+
+        let Value::Ptr(ptr) = arguments[0] else {
+            throw_ub!("invalid first argument to `Intrinsic::Spawn`");
+        };
+
+        let mut funcs = self.fn_addrs.iter().filter(|(_, fn_addr)| *fn_addr == ptr.addr);
+        let Some((func_name, _)) = funcs.next() else {
+            throw_ub!("invalid first argument to `Intrinsic::Spawn`, calling an address where there is no function");
+        };
+        let func = self.prog.functions[func_name];
```
RalfJung, line 182: This looks like we want a helper function to obtain a `Function` from a pointer, to share this code with `Call`.

lang/machine.md
```diff
@@ -137,17 +123,44 @@ We also define some helper functions that will be useful later.
...
+
+    fn cur_thread(&self) -> Thread<M> {
+        let Some(active_thread) = self.thread_manager.active_thread else {
+            panic!("`cur_thread` called without active thread!");
+        };
+
+        self.thread_manager.threads[active_thread]
+    }
+
+    fn mut_cur_thread<O>(&mut self, f: impl FnOnce(&mut Thread<M>) -> O) -> O {
```
RalfJung, line 157: Should be called `mutate_cur_thread` for consistency with the other similar functions

lang/thread.md
```diff
@@ -0,0 +1,163 @@
...
+            panic!("`mutate_cur_frame` called on empty stack!");
+        }
+
+        let last_idx = self.stack.len() - 1;
+        self.stack.mutate_at(last_idx, f)
+    }
+
+    fn mutate_stack<O>(&mut self, f: impl FnOnce(&mut List<StackFrame<M>>) -> O) -> O {
+        f(&mut self.stack)
+    }
```
RalfJung, line 72: This doesn't look like we need a `mutate` function? We need one when mutating things inside a GC ptr but here that is not the case.

lang/step.md
```diff
@@ -481,8 +483,18 @@ The callee should probably start with a bunch of `Finalize` statements to ensure
...
+        let frame = self.mutate_cur_stack(
+            |stack| stack.pop().unwrap()
+        );
         let func = frame.func;
+
+        let Some(caller_return_info) = frame.caller_return_info else {
+            // Only the bottom frame in a stack has no caller.
+            // Therefore the thread must terminate now.
+
+            return self.thread_manager.terminate_active_thread();
```
RalfJung, line 495: Here it might be good to `assert!` that indeed the stack is empty now.

lang/intrinsic.md
EnderGnar: Where should I define such helper functions?

lang/intrinsic.md
```diff
@@ -140,7 +152,93 @@ impl<M: Memory> Machine<M> {
...
+        });
+
+        let id_align = Align::max_for_offset(id_size).unwrap();
+
+        if let Some(ret_place) = ret_place {
+            let bytes = Type::encode::<M>(id_type, Value::Int(thread_id));
+            self.mem.store(ret_place, bytes, id_align)?;
+        }
+
+        ret(ExecutionStatus::Finished)
```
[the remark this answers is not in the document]
EnderGnar: Sounds like a good idea. Will think about it and if I can't find any issues implement it.

lang/intrinsic.md
RalfJung: `machine.md` seems like a good place for now

lang/machine.md
```diff
@@ -136,6 +123,112 @@ We also define some helper functions that will be useful later.
...
+pub enum ThreadState {
+    // The thread is enabled and can get executed.
+    Enabled,
+    // The thread is trying to join another thread and is blocked until that thread finishes.
+    BlockedOnJoin(ThreadId),
+    // The thread has terminated.
+    Terminated,
+}
+
+type ThreadId = Int;
```
RalfJung, line 199: Please keep all the type definitions together at the top of the file. They will be the most common thing people are looking for.

EnderGnar:
Yes once everything is resolved, I will squash all commits into one again. :)

RalfJung:
Just remember to do that *before* doing a rebase. :)

lang/machine.md
```diff
@@ -162,6 +196,16 @@ impl<M: Memory> Machine<M> {
 
         self.thread_manager.threads.mutate_at(active_thread, f)
     } 
+
+    fn fn_from_addr(&self, addr: mem::Address) -> NdResult<Function> {
```
RalfJung, line 200: This can probably be just `Result`? (also see https://github.com/RalfJung/minirust/pull/98)

lang/intrinsic.md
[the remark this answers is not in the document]
EnderGnar: I implemented this now. Some slight behavior changes were made. First: It is now not UB to call `allocate` without a `ret_place` and the pointer align is calculated with `Align::max_for_offset`.

spec/lang/machine.md
```diff
@@ -59,6 +60,40 @@ struct CallerReturnInfo<M: Memory> {
 }
 ```
 
+This defines the internal representation of a thread of execution.
+
+```rust
+pub struct Thread<M: Memory> {
+    state: ThreadState,
+
+    // The stack.
```
RalfJung, line 69:
```suggestion
    /// The stack.
```
Same for all the other comments on struct fields etc, please make them doc comments.

spec/lang/intrinsic.md
```diff
@@ -7,8 +7,7 @@ impl<M: Memory> Machine<M> {
         &mut self,
         intrinsic: Intrinsic,
         arguments: List<Value<M>>,
-        ret_place: Option<Place<M>>,
-    ) -> NdResult;
+    ) -> NdResult<Option<(Value<M>, Type)>>;
```
RalfJung, line 10: Why is there an `Option` here?

If an intrinsic returns `()`, that can be represented as an empty tuple.

EnderGnar: This would be an option. I don't yet see how this would make the code more understandable, since it would always need `ret((Value::Tuple(list![]), Type::Tuple{fields:list![], size: Size::ZERO}))`. I tried just creating constants, but `Value` is dependent on M and both need an empty list.

EnderGnar: This works. But I have changed the behavior of intrinsics. So the test `ub::alloc::alloc_noret` fails.

RalfJung: That's just a changed error message it seems? Seems fine to me.

EnderGnar: No. Since we now can't distinguish between intrinsics that return something and those who don't (at least it doesn't do that currently). It is now allowed to call `alloc` without a place to put the pointer into. This was not the case so far.

EnderGnar: Okay I looked closer into the error message. The issue is, since the test assumes the alloc to never return, It does not specify the next block to be executed. The UB we currently get is because of this missing next block. Not because the allocation failed. This is because the interpreter newly allows for `alloc` without a return place. I would need to completely remove this test to pass the checks.

RalfJung: Yeah seems fine to remove that test.

RalfJung:
Let's land this. :) Thanks! I will do some minor tweaks but seems easier to do that post-landing.

PR #106, opened 2023-03-22
add miniutil and minitest

memoryleak47:


tooling/minitest/README.md
```diff
@@ -0,0 +1,5 @@
+# minitest
+
+`minitest` is the MiniRust test suite.
+
+Use `./cov.sh` for the test coverage.
```
RalfJung, line 5:
```suggestion
Use `./cov.sh` to generate test coverage.
```

tooling/minitest/src/ub/alloc.rs
```diff
@@ -0,0 +1,145 @@
...
+
+#[test]
+fn alloc_wrongarg2() {
+    let locals = [ <*const i32>::get_ptype() ];
+
+    let b0 = block!(
+        storage_live(0),
+        Terminator::CallIntrinsic {
+            intrinsic: Intrinsic::Allocate,
+            arguments: list![const_int::<usize>(4), const_bool(true)], // bool is unexpected here!
```
RalfJung, line 134:
```suggestion
            // Second argument should be an int, so bool is unexpected here!
            arguments: list![const_int::<usize>(4), const_bool(true)],
```

tooling/minitest/src/ub/alloc.rs
```diff
@@ -0,0 +1,145 @@
...
+
+#[test]
+fn alloc_wrongarg1() {
+    let locals = [ <*const i32>::get_ptype() ];
+
+    let b0 = block!(
+        storage_live(0),
+        Terminator::CallIntrinsic {
+            intrinsic: Intrinsic::Allocate,
+            arguments: list![const_bool(true), const_int::<usize>(4)], // bool is unexpected here!
```
RalfJung, line 113:
```suggestion
            // First argument should be an int, so bool is unexpected here!
            arguments: list![const_bool(true), const_int::<usize>(4)],
```

tooling/minitest/src/ub/print.rs
```diff
@@ -0,0 +1,31 @@
...
+
+    let b0 = block!(
+        print(const_int::<u32>(42), 1), // ints can be printed
+    );
+    let b1 = block!(exit());
+
+    let f = function(Ret::No, 0, &locals, &[b0, b1]);
+    let p = program(&[f]);
+    dump_program(p);
+    assert_stop(p);
```
RalfJung, line 15: Ideally, this would test that we got the right output. But I assume that's non-trivial? In that case we should open an issue to track this.

tooling/miniutil/src/build/function.rs
```diff
@@ -0,0 +1,116 @@
...
+    },
+    (@{$($stmts:expr),*} $terminator:expr) => {
+        block(&[$($stmts),*], $terminator)
+    },
+    (@{} $stmt:expr, $($rest:expr),*) => {
+        block!(@{$stmt} $($rest),*)
+    },
+    (@{$($stmts:expr),*} $stmt:expr, $($rest:expr),*) => {
+        block!(@{$($stmts),*, $stmt} $($rest),*)
+    },
```
RalfJung, line 115: Isn't the `@{} $stmt:expr` clause redundant since it is an instance of the last one?

tooling/miniutil/src/fmt/function.rs
```diff
@@ -0,0 +1,224 @@
...
+    // Format locals
+    let mut locals: Vec<(LocalName, PlaceType)> = f.locals.iter().collect();
+
+    // The locals are formatted in the order of their names.
+    locals.sort_by_key(|(LocalName(name), _place_ty)| *name);
+
+    for (l, pty) in locals {
+        let local = fmt_local_name(l);
+        let ptype = fmt_ptype(pty, comptypes);
+        out += &format!("  let {local}: {ptype};\n");
```
RalfJung, line 63: It's probably not worth to rearchitect all this, but FYI you could save a lot of string copies here by (a) passing `out: &mut String` to these functions (rather than have them return a fresh `String`), and (b) using
```suggestion
        write!(out, "  let {local}: {ptype};\n");
```

tooling/miniutil/src/run.rs
```diff
@@ -0,0 +1,33 @@
...
+    }
+}
+
+// This drops everything not reachable from the machine.
+fn mark_and_sweep<M: Memory>(machine: &Machine<M>) {
+    // `set` is the set of gc pointers directly pointed to by `machine`.
+    let mut set = HashSet::new();
+    machine.points_to(&mut set);
+
+    gen_minirust::libspecr::hidden::mark_and_sweep(set);
```
RalfJung, line 32: A nicer API might be to pass `machine` to `mark_and_sweep`? Basically that function could take any parameter that has a `points_to` function, then the caller doesn't have to do this.

tooling/minitest/src/ub/print.rs
RalfJung: Ideally whoever sets up the interpreter can provide a `Box<dyn Writer>` that `print` will print to, or something like that. Then one can use stdout or one can use a `Vec<u8>` and check the result.

But anyway, I'm okay with not making this part of this PR.

tooling/minitest/src/ub/use_after_free.rs
```diff
@@ -0,0 +1 @@
+use crate::*;
```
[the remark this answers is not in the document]
memoryleak47: Whoops, this moved to `check_ptr.rs`

tooling/miniutil/src/build/function.rs
memoryleak47: I wholeheartedly agree, but for some reason, the rust compiler does not.
Removing it causes `no rules expected the token ","`.

tooling/miniutil/src/run.rs
memoryleak47: I originally just received a single `<T: GcCompat>` there, but in the general case you might want to have multiple roots.
So to keep it flexible, I went for a `HashSet`.

I could change it to `&[dyn GcCompat]` or something similar though

tooling/miniutil/src/build/function.rs
RalfJung: Uh... that's strange...
Maybe just leave a comment then. :shrug:

tooling/miniutil/src/run.rs
RalfJung: This is the only user of that API, it doesn't seem worth to over-generalize. But if you really want, then maybe have a general `mark_and_sweep1 that takes a list and a little helper (still in the gc lib) that works for a single root?

memoryleak47: Okay, I'll change the API back to `fn mark_and_sweep(root: impl GcCompat)`.
new libspecr and specr-transpile version is on the way

tooling/miniutil/src/run.rs
```diff
@@ -0,0 +1,25 @@
...
+// We fix `BasicMemory` as a memory for now.
+pub fn run_program(prog: Program) -> TerminationInfo {
+    let res: NdResult<!> = try {
+        let mut machine = Machine::<BasicMemory>::new(prog)?;
+
+        loop {
+            machine.step()?;
+
+            // Drops everything not reachable from `machine`.
+            gen_minirust::libspecr::hidden::mark_and_sweep::<Machine<BasicMemory>>(machine);
```
memoryleak47, line 15: Do you have any clue on why Rust can't figure out this generic parameter itself?
Even  `let mut machine: Machine<BasicMemory>` doesn't help.

mark_and_sweep just get's the machine as `impl GcCompat`

memoryleak47:
@RalfJung ready

tooling/miniutil/src/run.rs
RalfJung: `impl GcCompat` can't even be specified via turbofish... can you link me to where the function is declared?

memoryleak47: Haha, I just confused myself with `fn mark_and_sweep<T: GcCompat>(root: impl GcCompat)` ^^
New version is on the way

memoryleak47:
@RalfJung ready

PR #107, opened 2023-03-27
formatting improvements

memoryleak47:
- `fn foo() ↛` instead of `fn foo() -> none`
- `fn foo(_1) → _0` instead of `fn foo(_1: i32<align=4>) -> i32<align=4>`
- `i32@align(4)` instead of `i32<align=4>`
- `_ = exit()` instead of `none = exit()`


memoryleak47:
I think I might prefer `-/>` over the fancy UTF-8 arrow.
It's pretty small and unreadable for me.

RalfJung:
That also works for me.

memoryleak47:
@RalfJung ready

tooling/miniutil/src/fmt/expr.rs
```diff
@@ -1,21 +1,54 @@
...
+            let root = fmt_place_expr(root.extract(), comptypes).to_atomic_string();
+            FmtExpr::NonAtomic(format!("{root}.{field}"))
         }
         PlaceExpr::Index { root, index } => {
-            let root = fmt_place_expr(root.extract(), comptypes);
-            let index = fmt_value_expr(index.extract(), comptypes);
-            format!("{root}[{index}]")
+            let root = fmt_place_expr(root.extract(), comptypes).to_atomic_string();
+            let index = fmt_value_expr(index.extract(), comptypes).to_string();
+            FmtExpr::NonAtomic(format!("{root}[{index}]"))
```
RalfJung, line 51: In which context is this ambiguous?

memoryleak47: It could be ambiguous with operators that are written on the left, like `&raw _` or `-_`.
`&raw foo.bar` could either be `(&raw foo).bar` or `&raw (foo.bar)` if we don't define an operator priority.

tooling/miniutil/src/fmt/expr.rs
```diff
@@ -1,21 +1,54 @@
...
+            let root = fmt_place_expr(root.extract(), comptypes).to_atomic_string();
+            FmtExpr::Atomic(format!("{root}.{field}"))
         }
         PlaceExpr::Index { root, index } => {
-            let root = fmt_place_expr(root.extract(), comptypes);
-            let index = fmt_value_expr(index.extract(), comptypes);
-            format!("{root}[{index}]")
+            let root = fmt_place_expr(root.extract(), comptypes).to_atomic_string();
+            let index = fmt_value_expr(index.extract(), comptypes).to_string();
+            FmtExpr::Atomic(format!("{root}[{index}]"))
```
RalfJung, line 51: Still worth having a comment IMO

memoryleak47: done!

tooling/miniutil/src/fmt/expr.rs
```diff
@@ -1,21 +1,58 @@
...
+            let expr = fmt_value_expr(operand.extract(), comptypes).to_string();
+            FmtExpr::Atomic(format!("deref<{ptype}>({expr})"))
         }
         PlaceExpr::Field { root, field } => {
-            let root = fmt_place_expr(root.extract(), comptypes);
-            format!("{root}.{field}")
+            let root = fmt_place_expr(root.extract(), comptypes).to_atomic_string();
+            // You might think that `&raw foo.bar` is ambiguous and hence `foo.bar` is non-atomic.
+            // But we define `&raw foo.bar` to always mean `&raw (foo.bar)` and not `(&raw foo).bar`.
+            // Hence this can be atomic.
```
RalfJung, line 48:
```suggestion
            let root = fmt_place_expr(root.extract(), comptypes).to_atomic_string();
            // `&raw foo.bar` in Rust unambiguously means `&raw (foo.bar)`, and there is
            // no other context we have to worry about. Hence this can be atomic.
```
We don't get to define that, that's a Rust thing.^^

PR #111, opened 2023-04-11
explicit imports

memoryleak47:
specr-transpile@0.1.14 removes the implicit importing system.


spec/mem/prelude.md
```diff
@@ -0,0 +1,8 @@
+# MiniRust Memory prelude
+
+For the files in this folder, we assume some definitions and parameters to always be in scope.
+
+```rust
+use crate::prelude::*;
+use lang::PtrType;
```
RalfJung, line 7: Wait, `mem` depends on `PtrType`? Ugh...

spec/prelude/main.md
RalfJung, line 1: This is the only file remaining in `src/prelude` now, right? Can we rename it to `spec/prelude.md` or does specr not support that?

PR #112, opened 2023-04-17
Evaluate return place expressions in calls before other expressions

mattheww:
I think this is the intended order, based on the comments as they stood before PR 100.

If this isn't the intended order, then I think at least the comment in the `CallIntrinsic` case needs moving.

### DIFF
--- a/spec/lang/step.md
+++ b/spec/lang/step.md
@@ -416,12 +416,6 @@ impl<M: Memory> Machine<M> {
         &mut self,
         Terminator::Call { callee, arguments, ret: ret_expr, next_block }: Terminator
     ) -> NdResult {
-        let Value::Ptr(ptr) = self.eval_value(callee)? else {
-            panic!("call on a non-pointer")
-        };
-
-        let func = self.fn_from_addr(ptr.addr)?;
-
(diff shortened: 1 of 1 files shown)


spec/lang/step.md
```diff
@@ -552,14 +552,14 @@ impl<M: Memory> Machine<M> {
         Terminator::CallIntrinsic { intrinsic, arguments, ret: ret_expr, next_block }: Terminator
     ) -> NdResult {
         // First evaluate return place (left-to-right evaluation).
+        let ret_place = ret_expr.try_map(|ret_expr| self.eval_place(ret_expr))?;
```
RalfJung, line 555: Oh, good point, I might have screwed this up.

We should really have a test covering this...

spec/lang/step.md
```diff
@@ -432,6 +426,12 @@ impl<M: Memory> Machine<M> {
             ret::<NdResult<_>>((p, pty))
         })?;
 
+        let Value::Ptr(ptr) = self.eval_value(callee)? else {
```
RalfJung, line 429:
```suggestion
        // Then evaluate the function that will be called.
        let Value::Ptr(ptr) = self.eval_value(callee)? else {
```

PR #115, opened 2023-04-21
Fix align of intrinsic return.

EnderGnar:
Fixed issue that the alignment was chosen wrongly when an intrinsic was called with a return place.


spec/lang/step.md
```diff
@@ -554,16 +554,21 @@ impl<M: Memory> Machine<M> {
...
         let arguments = arguments.try_map(|arg| self.eval_value(arg))?;
 
-        let (value, ret_type) = self.eval_intrinsic(intrinsic, arguments)?;
+        let (value, ret_ty) = self.eval_intrinsic(intrinsic, arguments)?;
 
         if let Some(ret_place) = ret_place {
-            let align = Align::max_for_offset(ret_type.size::<M>()).unwrap();
-            
-            let bytes = Type::encode::<M>(ret_type, value);
+            let place_ty = place_ty.unwrap();
```
RalfJung, line 567: I think you can avoid the unwrap if you do the ret_expr.check_wf inside the `if let` here.

spec/lang/step.md
```diff
@@ -554,16 +554,21 @@ impl<M: Memory> Machine<M> {
...
+
+            let align = place_ty.align;
+
+            let bytes = Type::encode::<M>(ret_ty, value);
             self.mem.store(ret_place, bytes, align)?;
```
RalfJung, line 572: This should use `typed_store`

spec/lang/step.md
EnderGnar: Hmm but ret_expr would still need to be unwrapped. the only thing I would see is to just do `if let Some= ret_expr`. but then `eval_place` happens after evaluating the arguments.

RalfJung: Sure, but this `map` outside the `if` should not be needed.

spec/lang/step.md
```diff
@@ -558,13 +558,13 @@ impl<M: Memory> Machine<M> {
...
 
-        let (value, ret_type) = self.eval_intrinsic(intrinsic, arguments)?;
+        let (value, _ret_ty) = self.eval_intrinsic(intrinsic, arguments)?;
 
         if let Some(ret_place) = ret_place {
-            let align = Align::max_for_offset(ret_type.size::<M>()).unwrap();
-            
-            let bytes = Type::encode::<M>(ret_type, value);
-            self.mem.store(ret_place, bytes, align)?;
+            let ret_expr = ret_expr.unwrap();
```
RalfJung, line 564: Oh now I see what you mean... we should really just make `eval_place` return the type. But that's for a future change.

spec/lang/step.md
```diff
@@ -558,13 +558,13 @@ impl<M: Memory> Machine<M> {
         // Evaluate all arguments.
         let arguments = arguments.try_map(|arg| self.eval_value(arg))?;
 
-        let (value, ret_type) = self.eval_intrinsic(intrinsic, arguments)?;
+        let (value, _ret_ty) = self.eval_intrinsic(intrinsic, arguments)?;
```
RalfJung, line 561: I didn't realize this means we entirely ignore the `ret_ty` here. That's not good. Argh why are function calls so annoying.^^ I'll have to think about what we want to do here.

We should definitely add a test for calling an intrinsic where the result is put into a place of the wrong type.

spec/lang/step.md
EnderGnar: Wouldn't this introduce a lot of redundancy?

spec/lang/step.md
EnderGnar: Isn't it the case that we also don't check anything about types when a function returns? I have just written a test where a u64 is returned into a u8 and it passes.

spec/lang/step.md
RalfJung: Yes. But that might be the lesser evil.

spec/lang/step.md
RalfJung: That sounds potentially concerning, could you open an issue and paste the example there?

PR #118, opened 2023-05-08
eval_value & eval_place return tuple

EnderGnar:
eval_value & eval_place now return both the Value and the Type. This introduces a redundancy between eval_... and check_wf but allows for only one traversal of the tree.

Additionally the behavior of intrinsics was changed again. Now they take a return type and check that it is valid.

This forces the return value of every intrinsic except for the exit(). We might consider removing the optionality of the return place all together.


spec/lang/intrinsic.md
```diff
@@ -7,16 +7,29 @@ impl<M: Memory> Machine<M> {
...
+    Type::Tuple{fields:list![], size: Size::ZERO}
+}
+
+fn is_unit(ty: Type) -> bool {
+    let Type::Tuple{size, ..} = ty else {
+        return false;
+    };
+
+    size == Size::ZERO
+}
```
RalfJung, line 32: For a unit tuple, the field list also should be empty

spec/lang/intrinsic.md
```diff
@@ -7,16 +7,29 @@ impl<M: Memory> Machine<M> {
...
 
 ```rust
-fn unit<M: Memory>() -> (Value<M>, Type) {
-    (Value::Tuple(list![]), Type::Tuple{fields:list![], size: Size::ZERO})
+fn unit_value<M: Memory>() -> Value<M> {
+    Value::Tuple(list![])
 }
+
+fn unit_type() -> Type {
+    Type::Tuple{fields:list![], size: Size::ZERO}
```
RalfJung, line 23:
```suggestion
    Type::Tuple { fields: list![], size: Size::ZERO }
```
Let's follow standard Rust formatting

spec/lang/intrinsic.md
```diff
@@ -175,22 +204,22 @@ impl<M: Memory> Machine<M> {
             throw_ub!("invalid first argument to `Intrinsic::Spawn`, function returns something");
         }
 
+        if !matches!(ret_ty, Type::Int(_)) {
+            throw_ub!("invalid return type for `Intrinsic::Spawn`")
+        }
+
         // FIXME: What if the thread_id doesn't fit into a u32?
```
RalfJung, line 211: `u32` here is outdated since this now accepts any integer type

spec/lang/step.md
```diff
@@ -162,14 +161,50 @@ The functions `eval_un_op` and `eval_bin_op` are defined in [a separate file](op
...
-    fn eval_value(&mut self, ValueExpr::UnOp { operator, operand }: ValueExpr) -> NdResult<Value<M>> {
-        let operand = self.eval_value(operand)?;
-        ret(self.eval_un_op(operator, operand)?)
+    fn eval_value(&mut self, ValueExpr::UnOp { operator, operand }: ValueExpr) -> NdResult<(Value<M>, Type)> {
+        use lang::UnOp::*;
+
+        let (operand, _op_ty) = self.eval_value(operand)?;
+
+        let val = self.eval_un_op(operator, operand)?;
+        let ty = match operator {
```
RalfJung, line 170: This function should not match on `operator`. Instead `eval_un_op` should work the type as well.

Same for binops.

spec/lang/step.md
```diff
@@ -230,36 +268,48 @@ impl<M: Memory> Machine<M> {
...
+            Type::Union { fields, .. } => fields[field],
             _ => panic!("field projection on non-projectable type"),
         };
-        assert!(offset <= ty.size::<M>());
-        ret(self.ptr_offset_inbounds(root, offset.bytes())?)
+        assert!(offset <= ptype.ty.size::<M>());
+
+        let place = self.ptr_offset_inbounds(root, offset.bytes())?;
+        let ptype = PlaceType {
+            align: ptype.align.restrict_for_offset(offset),
```
RalfJung, line 282:
```suggestion
            // `offset` is statically known here (it is part of the field type)
            // so we are fine using it for `ptype`.
            align: ptype.align.restrict_for_offset(offset),
```

spec/lang/step.md
```diff
@@ -230,36 +268,48 @@ impl<M: Memory> Machine<M> {
...
             }
             _ => panic!("index projection on non-indexable type"),
         };
-        assert!(offset <= ty.size::<M>());
-        ret(self.ptr_offset_inbounds(root, offset.bytes())?)
+        assert!(offset <= ptype.ty.size::<M>());
+
+        let place = self.ptr_offset_inbounds(root, offset.bytes())?;
+        let ptype = PlaceType {
+            align: ptype.align.restrict_for_offset(field_ty.size::<M>()),
```
RalfJung, line 308:
```suggestion
            // We do *not* use `offset` here since that is only dynamically known.
            align: ptype.align.restrict_for_offset(field_ty.size::<M>()),
```

spec/lang/step.md
```diff
@@ -416,18 +465,16 @@ impl<M: Memory> Machine<M> {
...
-        let caller_locals = self.cur_frame().func.locals;
         let mut locals: Map<LocalName, Place<M>> = Map::new();
 
         // First evaluate the return place and remember it for `Return`. (Left-to-right!)
         let ret_place = ret_expr.try_map(|(caller_ret_place, _abi)| {
-            let p = self.eval_place(caller_ret_place)?;
-            let pty = caller_ret_place.check_wf::<M>(caller_locals, self.prog).unwrap(); // FIXME avoid a second traversal
-            ret::<NdResult<_>>((p, pty))
+            let place = self.eval_place(caller_ret_place)?;
+            ret::<NdResult<_>>(place)
```
RalfJung, line 473:
```suggestion
            self.eval_place(caller_ret_place)
```

tooling/minitest/Cargo.lock
```diff
@@ -75,9 +75,9 @@ checksum = "6a987beff54b60ffa6d51982e1aa1146bc42f19bd26be28b0586f252fccf5317"
 
 [[package]]
 name = "libspecr"
-version = "0.1.15"
+version = "0.1.14"
```
RalfJung, line 78: This downgrades specr to an earlier version. Why that?

tooling/minitest/src/ub/spawn.rs
```diff
@@ -123,3 +123,22 @@ fn spawn_func_returns() {
...
+    let b1 = block!(
+        join(load(local(0)), 2),
+    );
+    let b2 = block!(exit());
+
+    let f = function(Ret::No, 0, &locals, &[b0, b1, b2]);
+    let p = program(&[f, dummy_function()]);
+
+    assert_ub(p, "invalid return type for `Intrinsic::Spawn`");
+}
```
RalfJung, line 144: missing newline at the end of the file

tooling/minitest/Cargo.lock
EnderGnar: Because I'm stupid, did not update my `specr-transpile`. It always reset the required version for gen-minirust.

spec/lang/intrinsic.md
EnderGnar: Is this a single case or should I go over all my changes reviewing this?

RalfJung: Ideally this should be consistent everywhere.

EnderGnar:
Ready for next round of reviews.

PR #119, opened 2023-05-10
Fix LockState when Acquirer changes.

EnderGnar:
You warned me that the names are confusing when handing over the lock... And I might have made a mistake.


RalfJung:
Ah, good catch.

Can you craft a test case that was broken before, and is fixed now?

EnderGnar:
Well I realized it because the concurrency test failed on it. So I will post one soon I guess. It throws `Ub("releasing non-owned lock")` because the lock is still locked on the other thread.

RalfJung:
Hm maybe there could be a more targeted test for this specific problem?

EnderGnar:
This would be possible but it is just a statistics issue. It must happen, that a thread tries to acquire a lock while another is holding it. But I will craft one specifically for this case if you think it is important.

RalfJung:
I guess I should have asked, is there a way to *force* the test to fail? We might need other concurrency primitives such as atomics for that.

So maybe make a note that once atomics exist, you should add a deterministic test for this.

RalfJung:
So, merging for now and leaving the test for later.

RalfJung:
In Miri we have a mode where a thread never gets preempted, it has to explicitly 'yield' execution. That would also make this work. That might be useful for tests like this.

PR #120, opened 2023-05-10
Concurrency testing

EnderGnar:
First tests for concurrency: actually check we get a non-deterministic outcome.

### DIFF
--- a/tooling/minitest/src/pass/concurrency.rs
+++ b/tooling/minitest/src/pass/concurrency.rs
@@ -50,7 +50,7 @@ fn arbitrary_order() {
 
     let globals = [global_int::<u32>(), global_int::<u32>()];
 
-    let p = program(&[f, write(1), write(2)], &globals);
+    let p = program_with_globals(&[f, write(1), write(2)], &globals);
     
     // We now test, that the program can both finish with global(1) = 1 and = 2.
     let mut write_1 = false;
--- a/tooling/minitest/src/pass/concurrency.rs
+++ b/tooling/minitest/src/pass/concurrency.rs
@@ -33,7 +33,7 @@ fn arbitrary_order() {
     );
     let b2 = block!(
         storage_live(1),
-        spawn(fn_ptr(1), Some(local(1)), 3)
+        spawn(fn_ptr(2), Some(local(1)), 3)
     );
     let b3 = block!(
         join(load(local(0)), 4)
@@ -41,9 +41,12 @@ fn arbitrary_order() {
     let b4 = block!(
         join(load(local(1)), 5)
     );
-    let b5 = block!(exit());
+    let b5 = block!(
+        print(load(global::<u32>(1)), 6)
+    );
+    let b6 = block!( exit() );
 
-    let f = function(Ret::No, 0, &locals, &[b0, b1, b2, b3, b4, b5]);
+    let f = function(Ret::No, 0, &locals, &[b0, b1, b2, b3, b4, b5, b6]);
 
     let globals = [global_int::<u32>(), global_int::<u32>()];
 
@@ -53,11 +56,16 @@ fn arbitrary_order() {
     let mut write_1 = false;
     let mut write_2 = false;
 
-    for _ in 0..50 {
-        let Ok(m) = get_final_machine(p) else {
-            panic!("Machine not valid!");
+    for _ in 0..20 {
+        let out = match get_out(p) {
+            Ok(out) => out,
+            Err(err) => panic!("{:?}", err),
         };
-        
-        m.st
+
+        if out[0] == "1" { write_1 = true; }
+        if out[0] == "2" { write_2 = true; }
     }
+
+    assert!( write_1 );
+    assert!( write_2 );
 }
--- a/tooling/minitest/src/pass/concurrency.rs
+++ b/tooling/minitest/src/pass/concurrency.rs
@@ -1,0 +1,63 @@
+use crate::*;
+
+#[test]
+fn arbitrary_order() {
+
+    // A function that writes x to the global(1).
+    fn write(x: u32) -> Function {
+        let locals = [<u32>::get_ptype()];
+
+        let b0 = block!(
+            storage_live(0),
+            assign(local(0), load(global::<u32>(0))),
+            acquire(load(local(0)), 1)
+        );
+        let b1 = block!(
+            assign(global::<u32>(1), const_int::<u32>(x)),
+            release(load(local(0)), 2)
+        );
+        let b2 = block!(return_());
+
+        function(Ret::No, 0, &locals, &[b0, b1, b2])
+    }
+
+    // Main function, creates a lock and two threads.
+    let locals = [<u32>::get_ptype(), <u32>::get_ptype()];
+
+    let b0 = block!(
+        create_lock(global::<u32>(0), 1)
+    );
+    let b1 = block!(
+        storage_live(0),
+        spawn(fn_ptr(1), Some(local(0)), 2)
+    );
+    let b2 = block!(
+        storage_live(1),
+        spawn(fn_ptr(1), Some(local(1)), 3)
+    );
+    let b3 = block!(
+        join(load(local(0)), 4)
+    );
+    let b4 = block!(
+        join(load(local(1)), 5)
+    );
+    let b5 = block!(exit());
+
+    let f = function(Ret::No, 0, &locals, &[b0, b1, b2, b3, b4, b5]);
+
+    let globals = [global_int::<u32>(), global_int::<u32>()];
+
+    let p = program(&[f, write(1), write(2)], &globals);
+    
+    // We now test, that the program can both finish with global(1) = 1 and = 2.
+    let mut write_1 = false;
+    let mut write_2 = false;
+
+    for _ in 0..50 {
+        let Ok(m) = get_final_machine(p) else {
+            panic!("Machine not valid!");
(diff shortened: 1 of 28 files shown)


tooling/minitest/src/pass/concurrency.rs
```diff
@@ -0,0 +1,71 @@
...
+    let b0 = block!(
+        create_lock(global::<u32>(0), 1)
+    );
+    let b1 = block!(
+        storage_live(0),
+        spawn(fn_ptr(1), Some(local(0)), 2)
+    );
+    let b2 = block!(
+        storage_live(1),
+        spawn(fn_ptr(2), Some(local(1)), 3)
```
RalfJung, line 36: There's a lot of bare numbers here. Would be good to explain what they mean in comments.
(Would be even better if they could be handled automatically by the builder API but that is a bigger change.^^)

What are the fn_ptr numbers here?

EnderGnar: Yes I already thought about a better way to write programs. But I came to the conclusion minimize would be the optimal solution here.

The numbers stand for the function name. It is given based on the position in the function slice. Will comment this.

RalfJung: There might be something nice even without going all the way to MIR... if you have time left after finishing the concurrency stuff we could talk about that. For now some comments will do.

tooling/minitest/src/pass/concurrency.rs
```diff
@@ -0,0 +1,71 @@
+use crate::*;
+
+#[test]
+fn arbitrary_order() {
+
+    // A function that writes x to the global(1).
```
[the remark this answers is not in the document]
EnderGnar: I guess just for this function and not all comments explaining what the code should do.

tooling/miniutil/src/mock_write.rs
```diff
@@ -0,0 +1,76 @@
...
+        let string = from_utf8(&slice).unwrap().to_string();
+
+        string.split("\n").map( 
+            |str| str.to_string() 
+        ).collect()
+    }
+}
+
+#[derive(Default)]
+pub(crate) struct MockWrite {
```
RalfJung, line 31: I am confused by why this is so complicated. A single buffer type should be enough. Just `derive(Clone)` for `MockBuffer` and then `impl Write` for it?

Also every type should come with a doc comment explaining its purpose.  The fact that this entire file has 0 comments is always a bad sign. ;)

tooling/miniutil/src/run.rs
```diff
@@ -23,3 +27,30 @@ pub fn run_program(prog: Program) -> TerminationInfo {
...
+    };
+
+    let res = res.get_internal();
+
+    match res {
+        Ok(never) => never,
+        Err(TerminationInfo::MachineStop) => Ok(out.into_strings()),
+        Err(info) => Err(info)
+    }
+}
```
RalfJung, line 56: more missing newlines

tooling/miniutil/src/build/global.rs
```diff
@@ -0,0 +1,12 @@
...
+//bullshit
+pub fn global_int<T: TypeConv>() -> Global {
+    let bytes = List::from_elem(Some(0), T::get_size().bytes());
+    
+    Global { 
+        bytes, 
+        relocations: list!(), 
+        align: T::get_align(), 
+    }
+}
```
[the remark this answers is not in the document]
EnderGnar: Enabled it.

tooling/minitest/src/pass/concurrency.rs
```diff
@@ -0,0 +1,85 @@
...
+    );
+
+    // Join thread again.
+    let b4 = block!(
+        join(load(local(0)), 5)
+    );
+
+    // Print out the value last written to global(1).
+    let b5 = block!(
+        print(load(global::<u32>(1)), 6)
```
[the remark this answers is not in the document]
EnderGnar: We currently have `eprint` and `print`. 
I don't see a way to keep it simple and clearly mark it as stdout.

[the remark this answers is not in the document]
RalfJung: I guess that's consistent with surface Rust so seems fine.

RalfJung:
Can you rebase so that we get latest specr and see if CI works?

tooling/miniutil/src/run.rs
```diff
@@ -1,25 +1,48 @@
-use crate::*;
+use crate::{*, mock_write::MockWrite};
 
 use gen_minirust::prelude::NdResult;
 
 // Run the program and return its TerminationInfo.
 // We fix `BasicMemory` as a memory for now.
```
RalfJung, line 6:
```suggestion
/// Run the program and return its TerminationInfo.
/// Stdout/stderr are just forwarded to the host.
/// We fix `BasicMemory` as a memory for now.
```

tooling/miniutil/src/run.rs
```diff
@@ -1,25 +1,48 @@
...
+    let res: Result<!, TerminationInfo> = run(prog, out, err);
+    match res {
+        Ok(never) => never,
+        Err(t) => t,
+    }
+}
+
+// Run the program and return the output as a `Vec<String>` or a termination info if it
+// did not terminate correctly.
+// We fix `BasicMemory` as a memory for now.
```
RalfJung, line 20:
```suggestion
/// Run the program and return stdout as a `Vec<String>`  or a termination info 
/// if it did not terminate correctly. Stderr is just forwarded to the host.
///
/// We fix `BasicMemory` as a memory for now.
```

tooling/miniutil/src/run.rs
```diff
@@ -1,25 +1,48 @@
...
+
+    let res = run(prog, out.clone(), err);
+    match res {
+        Ok(never) => never,
+        Err(TerminationInfo::MachineStop) => Ok(out.into_strings()),
+        Err(info) => Err(info)
+    }
+}
+
+fn run(prog: Program, stdout: impl GcWrite, stderr: impl GcWrite) -> Result<!, TerminationInfo> {
```
RalfJung, line 33:
```suggestion
/// Run the program to completion using the given writers for stdout/stderr.
fn run(prog: Program, stdout: impl GcWrite, stderr: impl GcWrite) -> Result<!, TerminationInfo> {
```

tooling/miniutil/src/mock_write.rs
```diff
@@ -0,0 +1,67 @@
...
+    }
+}
+
+// Nothing within has anything to do with specr-lang. This points to nothing.
+impl GcCompat for MockWrite {
+    fn points_to(&self, _buffer: &mut std::collections::HashSet<usize>) { }
+
+    fn as_any(&self) -> &dyn std::any::Any {
+        self
+    }
```
RalfJung, line 47:
```suggestion
```
I updated libspecr so that `as_any` does not have to be implemented everywhere any more.

ci.sh
```diff
@@ -2,7 +2,7 @@
 set -ex
 
 # Fixed specr-transpile version
-VERSION="0.1.16"
+VERSION="0.1.18"
```
RalfJung, line 5: This required some more changes, please just rebase over master. :)

EnderGnar: Done. But we get a new warning, will look into it.
```
warning: ambiguous glob re-exports
  --> /minirust/tooling/miniutil/src/lib.rs:9:9
   |
9  | pub use gen_minirust::prelude::*;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^ the name `NdResult` in the type namespace is first re-exported here
...
13 | pub use gen_minirust::libspecr::*;
   |         ------------------------- but the name `NdResult` in the type namespace is also re-exported here
   |
```

RalfJung: Yeah we have an overlap between the libspecr prelude and the minirust prelude. Not sure what the best solution is. We do want the minirust one to take precedence though so we should probably at least reorder the imports.

AFAIK explicit imports take precedence, so maybe we can just add `pub use gen_minirust::prelude::NdResult` before the `prelude::*` import?

EnderGnar:
Ready for next round of reviews.

tooling/miniutil/src/build/expr.rs
```diff
@@ -135,6 +135,22 @@ pub fn local(x: u32) -> PlaceExpr {
     PlaceExpr::Local(LocalName(Name::from_internal(x)))
 }
 
+pub fn global<T: TypeConv>(x: u32) -> PlaceExpr {
+    let relocation = Relocation { 
+        name: GlobalName(Name::from_internal(x as _)),
```
RalfJung, line 140: Why `as _`? Just above we have a `from_internal` from a `u32` without `as`.

`as` is a red flag as data might get lost on a cast, so if it's not needed please remove it.

digama0: There's a clippy lint for this, perhaps you should consider enabling it

EnderGnar: I copied this from `miniutil/src/build/function.rs` where it is used multiple times.
The only possible reason I could find for this is in libspecr where Name is defined and the doc comment says, that the `u32` is hidden on purpose.

RalfJung: > I copied this from miniutil/src/build/function.rs where it is used multiple times.

Yes, over there `x` has type `usize` so a cast is needed. Here `x` already has the right type so it is not needed.

PR #131, opened 2023-06-19
Lock bug test

EnderGnar:
This is a test written for a bug I created and then removed again.

It requires a "yield only mode" since the test would be probabilistic otherwise.
Because the way I implemented this setting of the machine is rough and it uses a commit from #130 it is still a draft.

### DIFF
--- a/tooling/minitest/src/pass/lock_handover.rs
+++ b/tooling/minitest/src/pass/lock_handover.rs
@@ -1,0 +1,55 @@
+use crate::*;
+
+#[test]
+/// This test is written in response to a bug that was found.
+/// The bug was: When handing over a lock instead of making the lock owned by the acquiring thread
+/// it was once again owned by the releasing thread.
+/// This lead to UB once the acquirer tried to release the lock.
+/// 
+/// What it wants to check is: Does the lock handover work correctly?
+/// By making the critical section extremly large we get a high probability that the hand over happend.
+fn lock_handover() {
+    let locals = [<()>::get_ptype(), <u32>::get_ptype()];
+
+    let b0 = block!(
+        storage_live(1),
+        assign(local(1), const_int::<u32>(0)),
+        acquire(load(global::<u32>(0)), 1) 
+    );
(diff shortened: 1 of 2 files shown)


RalfJung:
Hm, I am not entirely happy with how this "yield only mode" affects the global machine. That's a bit too invasive for a test...

One possible idea for how to avoid that invasion: we could make the machine generic over a scheduler. Then the test could plug in a different, yield-only scheduler while usually we use a completely non-deterministic scheduler.

EnderGnar:
Agreed but this would mean that the scheduler is once again hidden behind a function call. 
We once "inlined" `schedule` because it was considered a main part of a step.

RalfJung (commented):
(just noticed Github hasn't posted these comments yet...)

RalfJung:
> Agreed but this would mean that the scheduler is once again hidden behind a function call. We once "inlined" `schedule` because it was considered a main part of a step.

Let's discuss this on Friday.

tooling/minitest/src/pass/lock_handover.rs
```diff
@@ -0,0 +1,55 @@
+use crate::*;
+
+#[test]
+/// This test is written in response to a bug that was found.
+/// The bug was: When handing over a lock instead of making the lock owned by the acquiring thread
+/// it was once again owned by the releasing thread.
+/// This lead to UB once the acquirer tried to release the lock.
+/// 
+/// What it wants to check is: Does the lock handover work correctly?
+/// By making the critical section extremly large we get a high probability that the hand over happend.
```
RalfJung, line 10: "extremely"? :) If I see correctly it is counting to 256, right?

RalfJung, line 10:
```suggestion
/// By making the critical section large (256 times around a loop)
/// we get a high probability that the hand over happend.
```

EnderGnar: Maybe 256 isn't that big... but I'd claim $2^{-256}$ is extremely small. :)

RalfJung:
Looking great, thanks!

PR #133, opened 2023-07-08
Minimize concurrency intrinsics

EnderGnar:
Added some intrinsics to minimize.
Changed such that thread roots now can return unit.


spec/lang/intrinsic.md
```diff
@@ -199,16 +199,19 @@ impl<M: Memory> Machine<M> {
             throw_ub!("invalid first argument to `Intrinsic::Spawn`, function takes arguments");
         }
 
-        // This is taken from Miri. It also does not allow for a return value in the root function of a thread.
-        if func.ret.is_some() {
-            throw_ub!("invalid first argument to `Intrinsic::Spawn`, function returns something");
+        // This is taken from Miri. It discards the return value of the function.
+        // We enforce this by only allowing functions that return () and !.
+        if let Some((name, _)) = func.ret {
+            if !is_unit(func.locals[name].ty) {
```
RalfJung, line 205: I think we should either force it to be always unit, or just ignore the return type. Or maybe accept all types of size 0. Accepting exactly `!` and `()` is somewhat odd.

spec/lang/machine.md
```diff
@@ -274,9 +278,18 @@ impl<M: Memory> ThreadManager<M> {
         }
     }
 
-    pub fn spawn(&mut self, func: Function) -> NdResult<ThreadId> {
+    pub fn spawn(&mut self, mem: &mut AtomicMemory<M>, func: Function) -> NdResult<ThreadId> {
```
RalfJung, line 281: Please just move this function to `impl Machine`, if it needs memory.

spec/lang/machine.md
```diff
@@ -274,9 +278,18 @@ impl<M: Memory> ThreadManager<M> {
...
         let thread_id = ThreadId::from(self.threads.len());
-        self.threads.push(Thread::new(func));
+
+        let mut locals = Map::new();
+        
+        // Create place for return local, if needed.
+        if let Some((ret_local, _abi)) = func.ret {
+            let callee_ret_layout = func.locals[ret_local].layout::<M>();
+            locals.insert(ret_local, mem.allocate(callee_ret_layout.size, callee_ret_layout.align)?);
```
RalfJung, line 289: Someone somewhere should deallocate this again... who is doing that?

tooling/minimize/intrinsics/src/lib.rs
```diff
@@ -29,3 +34,112 @@ pub unsafe fn deallocate(ptr: *mut u8, size: usize, align: usize) {
     let layout = Layout::from_size_align(size, align).unwrap();
     unsafe { System.deallocate(ptr, layout); }
 }
+
+static THREAD_HANDLES: Mutex<Vec<Option<JoinHandle<()>>>> = Mutex::new(Vec::new());
+
+// Spawn returns the index+1 of the thread because we don't have a JoinHandle for thread 0.
```
RalfJung, line 40: I don't understand how that is a reason for such a strange off-by-1

tooling/minimize/intrinsics/src/lib.rs
```diff
@@ -29,3 +34,112 @@ pub unsafe fn deallocate(ptr: *mut u8, size: usize, align: usize) {
...
+}
+
+pub fn join(index: usize) {
+    if index == 0 { panic!("Can not join thread 0.") }
+
+    let mut vec = THREAD_HANDLES.lock().unwrap();
+    let join_handle = vec.get_mut(index-1);
+
+    if let Some(join_handle) = join_handle {
+        let join_handle = mem::replace(join_handle, None);
```
RalfJung, line 57: This can be written a lot shorter. I'd start with

If you use `vec.get_mut(...).unwrap_or_else(|| panic!("joining non-existing thread")).take()`, then you have one level of nesting less and don't need the `replace` later.

Then you can also release the lock immediately again.

tooling/minimize/intrinsics/src/lib.rs
```diff
@@ -29,3 +34,112 @@ pub unsafe fn deallocate(ptr: *mut u8, size: usize, align: usize) {
...
+    if *lock == LockState::Unlocked {        
+        panic!("Releasing non locked lock.");
+    }
+    *lock = LockState::Unlocked;
+    drop(locks);
+
+    let thread_handles = THREAD_HANDLES.lock().unwrap();
+    for handle in thread_handles.as_slice() {
+        if let Some(handle) = handle {
+            handle.thread().unpark();
```
RalfJung, line 124: Uh what is going on here?^^

I suggest you reuse the locks from the standard library instead of implementing your own. (Also, unparking *all threads* is a bit too much.^^ A custom implementation should store a queue of waiting threads in the `Locked` state.)

tooling/minimize/intrinsics/src/lib.rs
```diff
@@ -29,3 +34,112 @@ pub unsafe fn deallocate(ptr: *mut u8, size: usize, align: usize) {
     let layout = Layout::from_size_align(size, align).unwrap();
     unsafe { System.deallocate(ptr, layout); }
 }
+
+static THREAD_HANDLES: Mutex<Vec<Option<JoinHandle<()>>>> = Mutex::new(Vec::new());
+
+// Spawn returns the index+1 of the thread because we don't have a JoinHandle for thread 0.
+pub fn spawn<F>(f: F) -> usize
+where F: FnOnce() -> () + Send + 'static,
```
RalfJung, line 42: Your current implementation will only actually work with `f: fn()`. That's also all you test. So please adjust the type signature for now.

But really this should have a signature more like `pthread_create`: `fn spawn(f: fn(*const ()), data: *const ())`. Then one can implement a wrapper around it that actually works for all `f: impl FnOnce()`.

spec/lang/machine.md
EnderGnar: To be honest I was thinking of just removing the ThreadManager entirely. It does not really have any purpose anymore.

RalfJung: Yeah that also works for me.

tooling/minimize/intrinsics/src/lib.rs
EnderGnar: Will add a comment to the code explaining why there is my own lock implementation.

EnderGnar:
Replaced by a completely new pull request.

PR #139, opened 2023-08-08
Spawn rework

EnderGnar:
Reworked spawn to allow functions to have a zero sized return value as well as forcing functions we spawn with to take a data pointer as an argument.

Added `Constant::Null` for the case that there is no data pointer. Open to find a better solution.

### DIFF
--- a/spec/lang/intrinsic.md
+++ b/spec/lang/intrinsic.md
@@ -185,7 +185,7 @@ impl<M: Memory> Machine<M> {
         arguments: List<(Value<M>, Type)>,
         ret_ty: Type,
     ) -> NdResult<Value<M>> {
-        if arguments.len() != 1 {
+        if arguments.len() != 2 {
             throw_ub!("invalid number of arguments for `Intrinsic::Spawn`");
         }
 
@@ -194,21 +194,28 @@ impl<M: Memory> Machine<M> {
         };
 
         let func = self.fn_from_addr(ptr.addr)?;
-
-        if func.args.len() != 0 {
-            throw_ub!("invalid first argument to `Intrinsic::Spawn`, function takes arguments");
+        if func.args.len() != 1 {
+            throw_ub!("invalid first argument to `Intrinsic::Spawn`, function should take one argument.");
         }
 
-        // This is taken from Miri. It also does not allow for a return value in the root function of a thread.
-        if func.ret.is_some() {
-            throw_ub!("invalid first argument to `Intrinsic::Spawn`, function returns something");
+        let (data_ptr, data_ptr_ty) = arguments[1];
+        if !matches!(data_ptr_ty, Type::Ptr(_)) {
+            throw_ub!("invalid second argument to `Intrinsic::Spawn`, data pointer should be a raw pointer.");
+        }
+
+        // This is taken from Miri. It discards the return value of the function.
+        // We enforce this by only allowing functions that return zero-sized types such as () and !.
+        if let Some(name) = func.ret {
+            if func.locals[name].ty.size::<M::T>() != Size::ZERO {
+                throw_ub!("invalid first argument to `Intrinsic::Spawn`, function returns something non zero sized");
+            }
         }
 
         if !matches!(ret_ty, Type::Int(_)) {
             throw_ub!("invalid return type for `Intrinsic::Spawn`")
         }
 
-        let thread_id = self.spawn(func)?;
+        let thread_id = self.spawn(func, data_ptr, data_ptr_ty)?;
 
         ret(Value::Int(thread_id))
     }
--- a/spec/lang/intrinsic.md
+++ b/spec/lang/intrinsic.md
@@ -192,7 +192,6 @@ impl<M: Memory> Machine<M> {
         let Value::Ptr(ptr) = arguments[0].0 else {
             throw_ub!("invalid first argument to `Intrinsic::Spawn`");
         };
-
         let func = self.fn_from_addr(ptr.addr)?;
         if func.args.len() != 1 {
             throw_ub!("invalid first argument to `Intrinsic::Spawn`, function should take one argument.");
@@ -200,7 +199,7 @@ impl<M: Memory> Machine<M> {
 
         let (data_ptr, data_ptr_ty) = arguments[1];
         if !matches!(data_ptr_ty, Type::Ptr(_)) {
-            throw_ub!("invalid second argument to `Intrinsic::Spawn`, data pointer should be a raw pointer.");
+            throw_ub!("invalid second argument to `Intrinsic::Spawn`, data pointer should be a pointer.");
         }
 
         // This is taken from Miri. It discards the return value of the function.
--- a/spec/lang/machine.md
+++ b/spec/lang/machine.md
@@ -148,7 +148,7 @@ impl<M: Memory> Machine<M> {
             intptrcast: IntPtrCast::new(),
             global_ptrs,
             fn_addrs,
-            threads: list![Thread::new(start_fn)],
+            threads: list![Thread::main(start_fn)],
             locks: List::new(),
             active_thread: ThreadId::ZERO,
             stdout,
@@ -198,7 +198,7 @@ Next, we define how to create a thread.
 
 ```rust
 impl<M: Memory> Thread<M> {
-    fn new(func: Function) -> Self {
+    fn new(func: Function, locals: Map<LocalName, Place<M>>) -> Self {
         // Setup the initial stack frame.
         // For the main thread, well-formedness ensures that the func has
         // no return value and no arguments.
@@ -206,7 +206,7 @@ impl<M: Memory> Thread<M> {
         // that the func has no arguments.
         let init_frame = StackFrame {
             func,
-            locals: Map::new(),
+            locals,
             caller_return_info: None,
             next_block: func.start,
             next_stmt: Int::ZERO,
@@ -216,6 +216,10 @@ impl<M: Memory> Thread<M> {
             state: ThreadState::Enabled,
             stack: list![init_frame],
         }
+    }
+
+    fn main(func: Function) -> Self {
+        Self::new(func, Map::new())
     }
 }
 ```
@@ -243,13 +247,41 @@ impl<M: Memory> Machine<M> {
         self.threads[self.active_thread]
     }
 
-    pub fn spawn(&mut self, func: Function) -> NdResult<ThreadId> {
+    fn spawn(&mut self, func: Function, data_pointer: Value<M>, data_ptr_ty: Type) -> NdResult<ThreadId> {
         let thread_id = ThreadId::from(self.threads.len());
-        self.threads.push(Thread::new(func));
+
+        let mut locals = Map::new();
+
+        // Create place for data pointer and place it there.
+        let callee_local = func.args[0];
+        let callee_pty = func.locals[callee_local];
+        // Allocate place with callee layout (a lot like `StorageLive`).
+        let callee_layout = callee_pty.layout::<M::T>();
+        let p = self.mem.allocate(callee_layout.size, callee_layout.align)?;
+        locals.insert(callee_local, p);
+
+        // This logic is taken from `eval(Terminator::Call)` and `eval_argument`.
+        // FIXME: find a way to avoid the code duplication.
+        let data_ptr_pty = PlaceType::new(data_ptr_ty, M::T::PTR_ALIGN);
+        if !Self::check_abi_compatibility(data_ptr_pty, func.locals[func.args[0]]) {
+            throw_ub!("invalid first argument to `Intrinsic::Spawn`, function should take a pointer as an argument.");
+        }
+        
+        self.mem.typed_store(p, data_pointer, callee_pty, Atomicity::None).unwrap();
+
+        // Create place for return local, if needed.
+        // The return value will be ignored in `terminate_active_thread`.
+        if let Some(ret_local) = func.ret {
+            let callee_ret_layout = func.locals[ret_local].layout::<M::T>();
+            locals.insert(ret_local, self.mem.allocate(callee_ret_layout.size, callee_ret_layout.align)?);
+        }
+
+        self.threads.push(Thread::new(func, locals));
+
         ret(thread_id)
     }
 
-    pub fn join(&mut self, thread_id: ThreadId) -> NdResult {
+    fn join(&mut self, thread_id: ThreadId) -> NdResult {
         let Some(thread) = self.threads.get(thread_id) else {
             throw_ub!("`Intrinsic::Join`: join non existing thread");
         };
@@ -266,7 +298,7 @@ impl<M: Memory> Machine<M> {
         ret(())
     }
 
-    pub fn terminate_active_thread(&mut self) -> NdResult {
+    fn terminate_active_thread(&mut self, frame: StackFrame<M>) -> NdResult {
         let active = self.active_thread;
         assert!(active != 0, "the main thread cannot terminate");
 
@@ -281,6 +313,14 @@ impl<M: Memory> Machine<M> {
             }
             thread
         }).collect();
+
+        // Deallocate everything. Same as in return.
+        // FIXME: avoid duplicating this code with `Return`.
+        for (local, place) in frame.locals {
+            // A lot like `StorageDead`.
+            let layout = frame.func.locals[local].layout::<M::T>();
+            self.mem.deallocate(place, layout.size, layout.align)?;
+        }
 
         ret(())
     }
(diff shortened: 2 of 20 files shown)


spec/lang/intrinsic.md
```diff
@@ -194,21 +194,33 @@ impl<M: Memory> Machine<M> {
...
+        if func.args.len() != 1 {
+            throw_ub!("invalid first argument to `Intrinsic::Spawn`, function should take one argument.");
+        }
 
-        if func.args.len() != 0 {
-            throw_ub!("invalid first argument to `Intrinsic::Spawn`, function takes arguments");
+        let (data_ptr, data_ptr_ty) = arguments[1];
+        if !matches!(data_ptr_ty, Type::Ptr(PtrType::Raw)) {
+            throw_ub!("invalid second argument to `Intrinsic::Spawn`, data pointer should be a raw pointer.");
+        }
```
RalfJung, line 204: Please treat this the same as `arguments[0]` above: it should be a pointer value. We don't care much for the type otherwise.

spec/lang/intrinsic.md
```diff
@@ -194,21 +194,33 @@ impl<M: Memory> Machine<M> {
...
+        // This logic is taken from `eval(Terminator::Call)` and `eval_argument`.
```
RalfJung, line 205:
```suggestion
        // This logic is taken from `eval(Terminator::Call)` and `eval_argument`.
        // FIXME: find a way to avoid the code duplication.
```

spec/lang/intrinsic.md
```diff
@@ -194,21 +194,33 @@ impl<M: Memory> Machine<M> {
...
+        let data_ptr_pty = PlaceType::new(data_ptr_ty, M::T::PTR_ALIGN);
+        if !Self::check_abi_compatibility(data_ptr_pty, func.locals[func.args[0]]) {
+            throw_ub!("invalid first argument to `Intrinsic::Spawn`, function should take a pointer as an argument.");
         }
 
-        // This is taken from Miri. It also does not allow for a return value in the root function of a thread.
-        if func.ret.is_some() {
-            throw_ub!("invalid first argument to `Intrinsic::Spawn`, function returns something");
+        // This is taken from Miri. It discards the return value of the function.
+        // We enforce this by only allowing functions that return () and !.
```
RalfJung, line 212:
```suggestion
        // We enforce this by only allowing functions that return zero-sized types such as () and !.
```

spec/lang/intrinsic.md
```diff
@@ -194,21 +194,33 @@ impl<M: Memory> Machine<M> {
...
+        if let Some(name) = func.ret {
+            if func.locals[name].ty.size::<M::T>() != Size::ZERO {
+                throw_ub!("invalid first argument to `Intrinsic::Spawn`, function returns something non zero sized");
+            }
         }
```
RalfJung, line 217: This check isn't actually needed any more I think. Now that we allocate a proper return place in `spawn`, we can accept arbitrary return types.

spec/mem/atomic.md
```diff
@@ -18,6 +18,10 @@ pub enum Atomicity {
 
     /// A non-atomic memory access.
     None,
+
+    /// An access that can not be seen by any other thread except
+    /// for the one executing it. Requires comment explaining reason if used.
```
RalfJung, line 23: This comment is wrong,. when spawning a thread then it is a different thread seeing the access!
```suggestion
    /// The initial write to an allocation that is only accessible by a single thread.
    /// Will be ignored for data race purposes. Requires comment explaining reason if used.
```

spec/lang/machine.md
```diff
@@ -282,6 +310,13 @@ impl<M: Memory> Machine<M> {
             thread
         }).collect();
 
+        // Deallocate everything. Same as in return.
```
RalfJung, line 313:
```suggestion
        // Deallocate everything. Same as in return.
        // FIXME: avoid duplicating this code with `Return`.
```

spec/lang/intrinsic.md
```diff
@@ -194,21 +194,33 @@ impl<M: Memory> Machine<M> {
...
 
-        if func.args.len() != 0 {
-            throw_ub!("invalid first argument to `Intrinsic::Spawn`, function takes arguments");
+        let (data_ptr, data_ptr_ty) = arguments[1];
+        if !matches!(data_ptr_ty, Type::Ptr(PtrType::Raw)) {
+            throw_ub!("invalid second argument to `Intrinsic::Spawn`, data pointer should be a raw pointer.");
+        }
+        // This logic is taken from `eval(Terminator::Call)` and `eval_argument`.
+        let data_ptr_pty = PlaceType::new(data_ptr_ty, M::T::PTR_ALIGN);
+        if !Self::check_abi_compatibility(data_ptr_pty, func.locals[func.args[0]]) {
```
RalfJung, line 207: This check should be in `spawn`. It's not great to have the check in one place and the logic that relies on the check in a completely different file.

spec/lang/machine.md
```diff
@@ -243,13 +249,35 @@ impl<M: Memory> Machine<M> {
...
+        // Create place for data pointer and place it there.
+        let callee_local = func.args[0];
+        let callee_pty = func.locals[callee_local];
+        // Allocate place with callee layout (a lot like `StorageLive`).
+        let callee_layout = callee_pty.layout::<M::T>();
+        let p = self.mem.allocate(callee_layout.size, callee_layout.align)?;
+        locals.insert(callee_local, p);
+        // Copy the value at callee type. We know the types are both pointer so this will fit.
+        // `p` is a fresh pointer so there should be no reason the store can fail.
+        // `Atomicity::Init` is okay since no other thread can have this allocation.
```
RalfJung, line 266:
```suggestion
        // `Atomicity::Init` is okay since only the newly created thread can access this allocation.
```

spec/lang/machine.md
```diff
@@ -243,13 +249,35 @@ impl<M: Memory> Machine<M> {
...
+        self.mem.typed_store(p, data_pointer, callee_pty, Atomicity::Init).unwrap();
+
+        // Create place for return local, if needed.
```
RalfJung, line 269:
```suggestion
        // Create place for return local, if needed.
        // The return value will be ignored in `terminate_active_thread`.
```

spec/lang/syntax.md
```diff
@@ -73,6 +73,8 @@ pub enum Constant {
     GlobalPointer(Relocation),
     /// A pointer pointing to a function.
     FnPointer(FnName),
+    /// The null pointer, pointing to address 0.
+    Null,
```
RalfJung, line 77:
```suggestion
    /// The null pointer, pointing to address 0.
    /// FIXME: generalize this to pointers from arbitrary constant integers.
    Null,
```

tooling/minitest/src/pass/concurrency.rs
```diff
@@ -16,7 +17,7 @@ fn arbitrary_order() {
         );
         let b2 = block!(return_());
 
-        function(Ret::No, 0, &[], &[b0, b1, b2])
+        function(Ret::Yes, 1, &locals, &[b0, b1, b2])
```
RalfJung, line 20: They would still be allowed to not have a return place, right? Why is this changed?

EnderGnar:
I have found an issue with our current method of preventing data races because of spawn happening on one thread and any data accesses happening on the spawned thread.

```
fn main() {
    let dp = allocate(sizeof *const ()); //A1
    *dp = dp;
    let id = spawn(second, *dp);
    join(id);
}

fn second(dp) {
    *dp = null;
}
```
What happens is, that the spawning thread will load from A1 because it dereferences `dp`. This is just an `Atomicity::None` because it is not directly an access because of the spawn function. In the next step the spawned thread accesses A1 with a write access and also `Atomicity::None`. This detects a data race where there is clearly not a data race.

I have added the test case on [weird datapointer](https://github.com/EnderGnar/minirust/tree/weird_datapointer) and it does detect a data race.

In terms of solutions I would either consider having "time frames" where each thread can at most do one step before we move on to the next time frame. And limiting data race detection to be within a time frame (We could then add both the current thread and the spawned thread as "having done a step" in the current time frame.) or something different like vector clock based detection.

A quick fix would also be to always have a start up step for every thread, giving it a `ThreadState::Spawned` where it just changes its state to `ThreadState::Enabled` and nothing else.

RalfJung:
Sorry for the slow reply. Good catch!

Why don't we have the same problem with locks? If a thread is blocked when acquiring the lock, then another thread releases, now the acquiring step is woken up and immediately takes a step writing to somewhere under the lock. I think we can construct a bad example there, too? We just need to make sure the releasing thread does something *in the step that releases*, like read the actual lock to release from a pointer. Then we have the same problem, no?

So we need a general way to mark "this thread just acquired synchronization from that other thread. And then maybe skip the data race detection if that's the exact pair of threads that are being checked next?

EnderGnar:
Yes that is correct. I will add a list of threads that received synchronization and add Threads on all actions that "change" a ThreadState to enabled. Also to Join and Release.

spec/lang/intrinsic.md
EnderGnar: Doesn't this suggest that we do something with the return value?

spec/mem/atomic.md
EnderGnar: This is not needed any more since we will get rid of Init.

spec/lang/machine.md
```diff
@@ -266,12 +304,19 @@ impl<M: Memory> Machine<M> {
...
-    pub fn terminate_active_thread(&mut self) -> NdResult {
+    fn terminate_active_thread(&mut self, frame: StackFrame<M>) -> NdResult {
         let active = self.active_thread;
         assert!(active != 0, "the main thread cannot terminate");
 
         self.threads.mutate_at(active, |thread| thread.state = ThreadState::Terminated);
 
+        // All threads that waited to join this thread get synchronized by this termination.
+        for i in ThreadId::ZERO..self.threads.len() {
+            if self.threads[i].state == ThreadState::BlockedOnJoin(active) {
```
RalfJung, line 315: This function now does this loop twice -- here, and then just below to enable these threads. Please only iterate once.

spec/lang/step.md
```diff
@@ -38,10 +38,15 @@ impl<M: Memory> Machine<M> {
             thread.state == ThreadState::Enabled
         })?;
 
+        // Reset synchronized threads; remember previous threads for data race detection.
+        let mut prev_sync = self.synchronized_threads;
+        self.synchronized_threads = Set::new();
```
RalfJung, line 43:
```suggestion
        // Remember threads synchronized by the previous step for data race detection
        // after this step.
        let mut prev_sync = self.synchronized_threads;
        // Every thread is always synchronized with itself.
        prev_sync.insert(self.active_thread);
        // Reset synchronized_threads for the next step.
        self.synchronized_threads = Set::new();
```

spec/lang/machine.md
```diff
@@ -25,6 +25,9 @@ pub struct Machine<M: Memory> {
     /// The currently / most recently active thread.
     active_thread: ThreadId,
 
+    /// A set of threads that have been synchronized.
```
RalfJung, line 28:
```suggestion
    /// A set of threads that have been synchronized.
    /// A thread being added here in a given step means that if the very next step is
    /// by that thread, we do *not* do data race detection: there was synchronization
    /// between these two steps, so any potential accesses are not racing.
```

spec/lang/intrinsic.md
RalfJung: Why would it suggest that? We just discard the return value. I don't think we have to reject that.

spec/lang/intrinsic.md
```diff
@@ -194,21 +194,28 @@ impl<M: Memory> Machine<M> {
         };
 
         let func = self.fn_from_addr(ptr.addr)?;
```
RalfJung, line 196:
```suggestion
        };
        let func = self.fn_from_addr(ptr.addr)?;
```
It's part of the same block, handling argument 0.

spec/lang/intrinsic.md
```diff
@@ -194,21 +194,28 @@ impl<M: Memory> Machine<M> {
...
+        if func.args.len() != 1 {
+            throw_ub!("invalid first argument to `Intrinsic::Spawn`, function should take one argument.");
+        }
 
-        if func.args.len() != 0 {
-            throw_ub!("invalid first argument to `Intrinsic::Spawn`, function takes arguments");
+        let (data_ptr, data_ptr_ty) = arguments[1];
+        if !matches!(data_ptr_ty, Type::Ptr(_)) {
+            throw_ub!("invalid second argument to `Intrinsic::Spawn`, data pointer should be a raw pointer.");
```
RalfJung, line 203:
```suggestion
            throw_ub!("invalid second argument to `Intrinsic::Spawn`, data pointer should be a pointer.");
```
The code accepts all pointers after all, not just raw pointers.

spec/lang/machine.md
```diff
@@ -198,15 +201,15 @@ Next, we define how to create a thread.
 
 ```rust
 impl<M: Memory> Thread<M> {
-    fn new(func: Function) -> Self {
+    fn new(func: Function, locals: Map<LocalName, Place<M>>) -> Self {
```
RalfJung, line 204:
```suggestion
    /// Create a new thread. The locals must have been already allocated.
    fn new(func: Function, locals: Map<LocalName, Place<M>>) -> Self {
```

spec/lang/step.md
```diff
@@ -38,10 +38,15 @@ impl<M: Memory> Machine<M> {
...
+
         // Update current thread; remember previous thread for data race detection.
-        let prev_thread = self.active_thread;
+        prev_sync.insert(self.active_thread);
```
RalfJung, line 46:
```suggestion
```

spec/lang/step.md
```diff
@@ -38,10 +38,15 @@ impl<M: Memory> Machine<M> {
             thread.state == ThreadState::Enabled
         })?;
 
+        // Reset synchronized threads; remember previous threads for data race detection.
+        let mut prev_sync = self.synchronized_threads;
+        self.synchronized_threads = Set::new();
+
         // Update current thread; remember previous thread for data race detection.
```
RalfJung, line 45:
```suggestion
        // Update current thread.
```

spec/mem/atomic.md
```diff
@@ -110,8 +110,8 @@ pub type ThreadId = Int;
 
 impl<M: Memory> AtomicMemory<M> {
     /// Given a list of previous accesses, checks if any of the current accesses is in a data race with any of those.
-    pub fn check_data_races(&self, current_thread: ThreadId, prev_thread: ThreadId, prev_accesses: List<Access>) -> Result {
-        if current_thread == prev_thread { return Ok(()) }
+    pub fn check_data_races(&self, current_thread: ThreadId, prev_threads: Set<ThreadId>, prev_accesses: List<Access>) -> Result {
```
RalfJung, line 113: `prev_threads` should probably be `prev_sync_threads ` or so

tooling/minitest/src/pass/concurrency.rs
RalfJung: Ah right.

tooling/minitest/src/pass/lock_issue.rs
```diff
@@ -0,0 +1,70 @@
...
+///     release(*global_1)
+/// }
+/// 
+/// fn second() {
+///     critical();
+/// }
+/// 
+/// fn main() {
+///     global_0 = create_lock();
+///     let id = spawn(second, null);
```
RalfJung, line 22: Why not `spawn(critical)`?

tooling/minitest/src/pass/weird_datapointer.rs
```diff
@@ -0,0 +1,51 @@
...
+///     join(id);
+/// }
+/// 
+/// fn second(dp) {
+///     *dp = null;
+/// }
+/// 
+/// This program should obviously not have a data race, but since
+/// we do a trace based search it has one in the base. Atomicity::Init does not prevent this.
+/// Extending Atomicity::Init to prevent this is ugly.
```
RalfJung, line 17: Comment seems outdated

tooling/minitest/src/pass/weird_datapointer.rs
```diff
@@ -0,0 +1,51 @@
...
+#[test]
+fn weird_datapointer() {
```
RalfJung, line 19:
```suggestion
fn thread_spawn_spurious_race() {
```

tooling/minitest/src/pass/lock_issue.rs
EnderGnar: It was like that because critical should return which was not always legal.
But now it is useful to keep the paths the same length.

tooling/minitest/src/pass/weird_datapointer.rs
RalfJung, line 1: The filename still says "weird_datapointer" which is not very descriptive, please sync with the test function name in this file

RalfJung:
Thanks, looks good! Please squash, then we can land. :)

PR #140, opened 2023-08-23
re-do the way we treat alignment around places

RalfJung:
This gets rid of `PlaceType` entirely, and instead enriches `Place` values with an `aligned: bool` field. Alignment is checked at deref time, but the check is only enforced if the place gets loaded/stored.

### DIFF
--- a/spec/lang/syntax.md
+++ b/spec/lang/syntax.md
@@ -100,6 +100,8 @@ pub enum UnOp {
     PtrAddr,
     /// Integer-to-pointer cast (uses previously exposed provenance)
     PtrFromExposed(PtrType),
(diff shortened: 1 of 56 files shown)


spec/lang/syntax.md
```diff
@@ -154,8 +154,8 @@ pub enum PlaceExpr {
     Deref {
         #[specr::indirection]
         operand: ValueExpr,
-        // The type of the newly created place.
-        ptype: PlaceType,
+        // The type of the newley created place
```
digama0, line 157:
```suggestion
        // The type of the newly created place
```

RalfJung: Good catch, thanks!

PR #141, opened 2023-08-29
Minimize intriniscs for Concurrent operations

EnderGnar:
This replaces #133 it was just easier to rewrite it all instead of reorganizing the mess.

Since some of the minimize code already implied that spawn would only work with C calling conventions, this PR incorporates this as well.

We added an implementation for all intrinsics and added tests for all of them.


tooling/minimize/intrinsics/src/lib.rs
```diff
@@ -29,3 +33,96 @@ pub unsafe fn deallocate(ptr: *mut u8, size: usize, align: usize) {
     let layout = Layout::from_size_align(size, align).unwrap();
     unsafe { System.deallocate(ptr, layout); }
 }
+
+
+static JOIN_HANDLES: Mutex<Vec<Option<JoinHandle<()>>>> = Mutex::new( Vec::new() );
```
RalfJung, line 38: Needs a comment explaining the purpose of this global

tooling/minimize/intrinsics/src/lib.rs
```diff
@@ -29,3 +33,96 @@ pub unsafe fn deallocate(ptr: *mut u8, size: usize, align: usize) {
...
+    handle.join().unwrap();
+}
+
+
+#[derive(PartialEq)]
+enum LockState {
+    Open,
+    Locked,
+}
+static LOCKS: Mutex<Vec<LockState>> = Mutex::new( Vec::new() );
```
RalfJung, line 74: Needs a comment explaining the purpose of this global

tooling/minimize/intrinsics/src/lib.rs
```diff
@@ -29,3 +33,96 @@ pub unsafe fn deallocate(ptr: *mut u8, size: usize, align: usize) {
...
+static JOIN_HANDLES: Mutex<Vec<Option<JoinHandle<()>>>> = Mutex::new( Vec::new() );
+
+struct SendPtr<T>(*const T);
+unsafe impl<T> Send for SendPtr<T> {}
+
+pub fn spawn(fn_ptr: extern "C" fn(*const ()), data_ptr: *const ()) -> usize {
+    let mut join_handles = JOIN_HANDLES.lock().unwrap();
+
+    if join_handles.len() == 0 {
+        join_handles.push(None);
```
RalfJung, line 47: What is the point of this?

tooling/minimize/intrinsics/src/lib.rs
```diff
@@ -29,3 +33,96 @@ pub unsafe fn deallocate(ptr: *mut u8, size: usize, align: usize) {
...
+    let ptr = SendPtr(data_ptr);
+    let handle = thread::spawn(
+        move || {
+            let ptr = ptr;
+            fn_ptr(ptr.0);
+        }
+    );
+    join_handles.push(Some(handle));
+
+    join_handles.len()-1
```
RalfJung, line 59:
```suggestion
    // Return the index of the element we just pushed.
    join_handles.len()-1
```

tooling/minimize/intrinsics/src/lib.rs
```diff
@@ -29,3 +33,96 @@ pub unsafe fn deallocate(ptr: *mut u8, size: usize, align: usize) {
...
+    let mut join_handles = JOIN_HANDLES.lock().unwrap();
+
+    if join_handles.len() == 0 {
+        join_handles.push(None);
+    }
+
+    let ptr = SendPtr(data_ptr);
+    let handle = thread::spawn(
+        move || {
+            let ptr = ptr;
```
RalfJung, line 53:
```suggestion
            let ptr = ptr; // avoid per-field closure capturing
```

tooling/minimize/intrinsics/src/lib.rs
```diff
@@ -29,3 +33,96 @@ pub unsafe fn deallocate(ptr: *mut u8, size: usize, align: usize) {
...
+
+pub fn create_lock() -> usize {
+    let mut locks = LOCKS.lock().unwrap();
+
+    let id = locks.len();
+    locks.push(LockState::Open);
+    id
+}
+
+// Spin until the lock is open.
```
RalfJung, line 84:
```suggestion
// Spin (with parking) until the lock is open.
```

tooling/minimize/intrinsics/src/lib.rs
```diff
@@ -29,3 +33,96 @@ pub unsafe fn deallocate(ptr: *mut u8, size: usize, align: usize) {
...
+pub fn acquire(lock_id: usize) {
+    loop {
+        let mut locks = LOCKS.lock().unwrap();
+
+        if locks[lock_id] == LockState::Open {
+            locks[lock_id] = LockState::Locked;
```
RalfJung, line 90:
```suggestion
            // We can grab the lock! Return successfully.
            locks[lock_id] = LockState::Locked;
```

tooling/minimize/intrinsics/src/lib.rs
```diff
@@ -29,3 +33,96 @@ pub unsafe fn deallocate(ptr: *mut u8, size: usize, align: usize) {
...
+        thread::park()
+    }
+}
+
+// Unparks all threads for simplicity.
+pub fn release(lock_id: usize) {
+    LOCKS.lock().unwrap()[lock_id] = LockState::Open;
+
+    let join_handles = JOIN_HANDLES.lock().unwrap();
+    for handle in join_handles.iter().flatten() {
```
RalfJung, line 105:
```suggestion
    // We don't precisely track who is waiting for which lock, so
    // we just wake up *all* the threads.
    for handle in join_handles.iter().flatten() {
```

tooling/minimize/src/rvalue.rs
```diff
@@ -190,6 +190,20 @@ pub fn translate_rvalue<'cx, 'tcx>(
...
+            let rs::Operand::Constant(box f1) = func else { panic!() };
+            let rs::ConstantKind::Val(_, f2) = f1.literal else { panic!() };
+            let rs::TyKind::FnDef(f, substs_ref) = f2.kind() else { panic!() };
+            let key = (*f, *substs_ref);
+
+            if !fcx.cx.fn_name_map.contains_key(&key) {
+                let fn_name = fcx.cx.fn_name_map.len();
+                let fn_name = FnName(Name::from_internal(fn_name as _));
+                fcx.cx.fn_name_map.insert(key, fn_name);
+            }
```
RalfJung, line 203: Can you add a method to `cx` that encapsulates this, and use it everywhere that we currently directly work with `fn_name_map`?

tooling/minimize/src/ty.rs
```diff
@@ -78,6 +78,10 @@ pub fn translate_ty<'tcx>(ty: rs::Ty<'tcx>, tcx: rs::TyCtxt<'tcx>) -> Type {
             let elem = GcCow::new(translate_ty(*ty, tcx));
             Type::Array { elem, count }
         }
+        rs::TyKind::FnPtr(_) => {
+            // For now we use the C ABI for everything since that's what `spawn` needs...
+            Type::Ptr(PtrType::FnPtr(CallingConvention::C))
```
RalfJung, line 83: We already have logic to translate Rust calling conventions to MiniRust ones, please use that here.

tooling/minimize/tests/pass/locks.stdout
```diff
@@ -0,0 +1,2 @@
+0
+1
```
RalfJung, line 2: Isn't the order non-deterministic here?

tooling/minimize/intrinsics/src/lib.rs
EnderGnar: In the last version there was a weird off by one. This could also be implemented using Lazy but I was unsure if we should add dependencies on such crates in the intrinsics.

tooling/minimize/tests/pass/locks.stdout
EnderGnar: No. The lock is held by the main thread before the other thread is spawned. It will only release it once it has printed.

tooling/minimize/src/ty.rs
EnderGnar: I am unsure how I even get the Conv. The FnPtr type only provides `PolyFnSig` and there we only find an ABI. The only *very* ugly way I see to convert between them would be through `&str`.
Other than Rust and C almost none would match up.

EnderGnar: This would look something like this...
`translate_calling_convention(FromStr::from_str(sig.abi().name()).unwrap())`

tooling/minimize/intrinsics/src/lib.rs
RalfJung: Off by one in which sense?

The IDs returned by the intrinsics should be treated as opaque, so it doesn't matter whether this and MiniRust return the same IDs.

tooling/minimize/tests/pass/locks.stdout
RalfJung: Whenever there's something non-obvious like this going on, a comment in the code is required. :)

tooling/minimize/intrinsics/src/lib.rs
```diff
@@ -29,3 +33,109 @@ pub unsafe fn deallocate(ptr: *mut u8, size: usize, align: usize) {
...
+    }
+}
+
+// Unparks all threads for simplicity.
+pub fn release(lock_id: usize) {
+    LOCKS.lock().unwrap()[lock_id] = LockState::Open;
+
+    let mut waiting = WAITING.lock().unwrap();
+    // We don't precisely track who is waiting for which lock, so
+    // we just wake up *all* the threads.
```
RalfJung, line 117:
```suggestion
    // we just wake up all threads that are waiting for *some* lock.
    // They will re-queue themselves into `WAITING` if they cannot acquire their lock.
```

tooling/minimize/intrinsics/src/lib.rs
```diff
@@ -29,3 +33,109 @@ pub unsafe fn deallocate(ptr: *mut u8, size: usize, align: usize) {
...
+        if locks[lock_id] == LockState::Open {
+            // We can grab the lock! Return successfully.
+            locks[lock_id] = LockState::Locked;
+
+            return;
+        }
+
+        drop(locks);
+        WAITING.lock().unwrap().push(thread::current());
+        thread::park()
```
RalfJung, line 107:
```suggestion
        WAITING.lock().unwrap().push(thread::current());
        // In principle another thread could call `unpark` here, before we
        // `park` ourselves. However, in that case the next `park` of this thread
        // is guaranteed to return immediately.
        thread::park()
```

tooling/minimize/intrinsics/src/lib.rs
```diff
@@ -29,3 +33,109 @@ pub unsafe fn deallocate(ptr: *mut u8, size: usize, align: usize) {
     let layout = Layout::from_size_align(size, align).unwrap();
     unsafe { System.deallocate(ptr, layout); }
 }
+
+// This global keeps track of any join handles produced. It is needed
+// because the minirust intrinsic for spawn only returns an integer and
+// the join only takes an integer.
```
RalfJung, line 39:
```suggestion
// the join only takes an integer, so we have to map these integers to `JoinHandles`
// for the Rust-based implementation.
```

tooling/minimize/intrinsics/src/lib.rs
```diff
@@ -29,3 +33,109 @@ pub unsafe fn deallocate(ptr: *mut u8, size: usize, align: usize) {
...
+
+
+#[derive(PartialEq)]
+enum LockState {
+    Open,
+    Locked,
+}
+
+// We implement our own locks for the intrinsics. They do not have an implicit
+// logic to unlock and operate on integers. This is why they are implemented in this way.
```
RalfJung, line 79: Operate on integers? Implicit logic to unlock? I am very confused.
```suggestion
// We cannot use the locks from the standard library since MiniRust locks are acquired
// and released with an integer ID, but the standard library requires an `&Mutex` for
// `acquire` and a `MutexGuard` for `release`.
```

tooling/minimize/src/program.rs
```diff
@@ -64,6 +64,17 @@ impl<'tcx> Ctxt<'tcx> {
             globals: self.globals,
         }
     }
+
+    // Returns FnName associated with some key. If it does not exist it creates a new one.
+    pub fn get_fn_name_or_set(&mut self, key: (rs::DefId, &'tcx rs::List<rs::GenericArg<'tcx>>)) -> FnName {
```
RalfJung, line 69: I think just `get_fn_name` is a good name. The fact that this might lazily extend the map is an implementation detail.

tooling/minimize/src/program.rs
```diff
@@ -64,6 +64,17 @@ impl<'tcx> Ctxt<'tcx> {
...
+        if !self.fn_name_map.contains_key(&key) {
```
RalfJung, line 70: This is a great usecase for `fn_name_map.entry(&key).or_insert_with` :)

tooling/minimize/src/rvalue.rs
```diff
@@ -196,13 +196,10 @@ pub fn translate_rvalue<'cx, 'tcx>(
...
-            if !fcx.cx.fn_name_map.contains_key(&key) {
-                let fn_name = fcx.cx.fn_name_map.len();
-                let fn_name = FnName(Name::from_internal(fn_name as _));
-                fcx.cx.fn_name_map.insert(key, fn_name);
-            }
-
-            build::fn_ptr(fcx.cx.fn_name_map[&key].0.get_internal())
+            build::fn_ptr(
+                fcx.cx.get_fn_name_or_set(key)
+                .0.get_internal()
```
RalfJung, line 201: Please keep the formatting at least somewhat close to standard Rust formatting. This should all go in one line.

tooling/minimize/src/ty.rs
RalfJung: No no don't use a string please. :) You need to use `fn_abi_of_fn_ptr`. This should be similar to the existing code that uses `fn_abi_of_instance`, so maybe looking at that is helpful.

tooling/minimize/src/program.rs
EnderGnar: I use it now but we precompute the `length` since the borrow checker complains and the length is some what ambiguous while the `Entry` exists.

EnderGnar:
I am unsure if we discussed this but currently the behavior of join does not completely match the MiniRust implementation.

The Rust implementation only works for one thread joining, unlike what MiniRust does, where as many threads as want can join the same thread.

tooling/minimize/src/program.rs
RalfJung: Ah, that's unfortunate, so it's not as nice as I had hoped. Oh well.

RalfJung:
> The Rust implementation only works for one thread joining, unlike what MiniRust does, where as many threads as want can join the same thread.

Please add a comment in the code explaining that.

tooling/minimize/intrinsics/src/lib.rs
```diff
@@ -29,3 +33,115 @@ pub unsafe fn deallocate(ptr: *mut u8, size: usize, align: usize) {
...
+    join_handles.push(Some(handle));
+
+    // Return the index of the element we just pushed.
+    join_handles.len()-1
+}
+
+
+// This implementation differes slightly from how MiniRust does this.
+// Here only one thread can join another thread, while in MiniRust
+// a single thread can be joined by many.
```
RalfJung, line 65:
```suggestion
// This implementation differs slightly from how MiniRust does this.
// Here a thread can only be joined once, while in MiniRust
// a single thread can be joined many times.
```

PR #142, opened 2023-09-23
AtomicFetchAndOp intrinsic

EnderGnar:
Implementation of atomic add and atomic sub.
I am not sure yet if I should move the helper FetchOp into the spec since it would remove an UB check.


spec/lang/step/intrinsics.md
```diff
@@ -383,3 +383,69 @@ impl<M: Memory> Machine<M> {
     }
 }
 ```
+
+We also implement the atomic fetch operations. First we define a helper function to decide which operations can be in a fetch operation.
```
RalfJung, line 387: I think this can be just part of the `Atomic accesses` block.

spec/lang/step/intrinsics.md
```diff
@@ -383,3 +383,69 @@ impl<M: Memory> Machine<M> {
...
+    match op {
+        B::Add | B::Sub => true,
+        _ => false
+    }
+}
+
+impl<M: Memory> Machine<M> {
+    fn eval_intrinsic(
+        &mut self,
+        Intrinsic::AtomicFetch(op): Intrinsic,
```
RalfJung, line 402: In terms of naming, I think this should be called `AtomicFetchAndOp`, since it doesn't just fetch.

spec/lang/step/intrinsics.md
```diff
@@ -383,3 +383,69 @@ impl<M: Memory> Machine<M> {
     }
 }
 ```
+
+We also implement the atomic fetch operations. First we define a helper function to decide which operations can be in a fetch operation.
+
+```rust
+/// Predicate to indicate if integer bin-op can be used for atomic fetch operations.
+fn is_atomic_binop(op: BinOpInt) -> bool {
```
RalfJung, line 391: Why do we only support some of them?

spec/lang/step/intrinsics.md
```diff
@@ -383,3 +383,69 @@ impl<M: Memory> Machine<M> {
...
+
+        if !matches!(ret_ty, Type::Int(_)) {
+            throw_ub!("invalid return type for `Intrinis::AtomicFetch`, only works with integers");
+        }
+
+        let size = ret_ty.size::<M::T>();
+        // All integer sizes are powers of two.
+        let align = Align::from_bytes(size.bytes()).unwrap();
+        if size > M::T::MAX_ATOMIC_SIZE {
+            throw_ub!("invalid return type for `Intrinsic::AtomicFetch`, size to big");
```
RalfJung, line 427:
```suggestion
            throw_ub!("invalid return type for `Intrinsic::AtomicFetch`, size too big");
```

tooling/minimize/intrinsics/src/lib.rs
```diff
@@ -145,3 +145,13 @@ pub unsafe fn compare_exchange(ptr: *mut u32, current: u32, new: u32) -> u32 {
         Err(ret) => ret,
     }
 }
+
+pub unsafe fn atomic_add(ptr: *mut u32, delta: u32) -> u32 {
```
RalfJung, line 149: `atomic_fetch_add` would be more consistent with the Rust APIs.

spec/lang/step/intrinsics.md
EnderGnar: I only added support for the atomic fetch operations that are exposed by the Rust Atomics and by LLVM. My best guess is that since multiply and division are both cycle intensive operations they are not used atomically by the ISA.

But if you prefer having all integer operations as atomic, I don't have a problem with that.

spec/lang/step/intrinsics.md
EnderGnar: Should I put the helper function (if we will still use it) at the beginning of this block or keep it in the middle?

spec/lang/step/intrinsics.md
RalfJung: Hm okay that's fair, but needs a comment.

spec/lang/step/intrinsics.md
RalfJung: You can put the helper function inside the intrinsic function. :)

Though I think I'd prefer to have this as part of well-formed. Making it UB to pass the wrong operand doesn't feel right.

tooling/miniutil/src/fmt/function.rs
```diff
@@ -222,6 +223,15 @@ fn fmt_terminator(t: Terminator, comptypes: &mut Vec<CompType>) -> String {
     }
 }
 
+fn fmt_fetch(binop: BinOpInt) -> &'static str {
+    use BinOpInt as B;
+    match binop {
+        B::Add => "atomic-fetch-add",
+        B::Sub => "atomic-fetch-sub",
```
RalfJung, line 230:
```suggestion
        B::Add => "atomic_fetch_add",
        B::Sub => "atomic_fetch_sub",
```

tooling/miniutil/src/fmt/function.rs
```diff
@@ -222,6 +223,15 @@ fn fmt_terminator(t: Terminator, comptypes: &mut Vec<CompType>) -> String {
...
+        _ => "atomic-fetch-BAD",
```
RalfJung, line 231:
```suggestion
        _ => "atomic_fetch_ILL_FORMED",
```

RalfJung:
Thanks, looking good. :)

PR #143, opened 2023-09-28
Enum variant as ValueExpr

essickmango:
Moves enum variant expressions from Constant to ValueExpr in order to allow other sources than constants for the fields (including now tuples). Also adds miniutil formatting for enums because it broke with the move and I could fix it (but not test it).

### DIFF
--- a/tooling/miniutil/src/fmt/ty.rs
+++ b/tooling/miniutil/src/fmt/ty.rs
@@ -87,13 +87,6 @@ fn get_comptype_index(ty: CompType, comptypes: &mut Vec<CompType>) -> CompTypeIn
         None => {
             let n = comptypes.len();
             comptypes.push(ty);
-
-            if let Type::Enum { variants, .. } = ty.0 {
-                variants.iter().for_each(|v| {
-                    get_comptype_index(CompType(v), comptypes);
-                });
-            }
-
             n
         }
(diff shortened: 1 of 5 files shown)


tooling/miniutil/src/fmt/ty.rs
```diff
@@ -121,32 +120,58 @@ pub(super) fn fmt_comptypes(mut comptypes: Vec<CompType>) -> String {
 }
 
 fn fmt_comptype(i: CompTypeIndex, t: CompType, comptypes: &mut Vec<CompType>) -> String {
-    let (keyword, fields, opt_chunks, size, align) = match t.0 {
-        Type::Tuple { fields, size, align } => ("tuple", fields, None, size, align),
+    let (keyword, size, align) = match t.0 {
+        Type::Tuple { fields, size, align } => ("tuple", size, align),
```
RalfJung, line 124: CI shows an "unused variable" warning here

RalfJung (commented):
Thanks! Looks good, but CI found a warning. :)

tooling/miniutil/src/fmt/ty.rs
essickmango: saw it and understood: warnings = errors with regards to PRs

PR #146, opened 2023-10-25
Enum memory representation and well-formedness

essickmango:
Implements the enum memory representation and discriminant using a decision tree for decoding and tagger for encoding.

Also adds some well-formedness. Both an invalid discriminant and values of uninhabited enums are not allowed (see https://doc.rust-lang.org/reference/behavior-considered-undefined.html).

### DIFF
--- a/spec/lang/representation.md
+++ b/spec/lang/representation.md
@@ -243,7 +243,7 @@ fn compute_discriminant<M: Memory>(bytes: List<AbstractByte<M::Provenance>>, dis
         Discriminator::Known(val) => Some(val),
         Discriminator::Invalid => None,
         Discriminator::Unknown { offset, children, fallback } => {
-            let AbstractByte::Init(val, _) = bytes[offset]
+            let AbstractByte::Init(val, _) = bytes[offset.bytes()]
                 else { return None };
             let next_discriminator = children.get(val).unwrap_or(fallback);
             compute_discriminant::<M>(bytes, next_discriminator)
@@ -274,7 +274,7 @@ impl Type {
         // write tag afterwards into the encoded bytes. This is fine as we don't allow
         // encoded data and the tag to overlap at the moment.
         for (offset, value) in tagger.iter() {
-            bytes.set(offset, AbstractByte::Init(value, None));
+            bytes.set(offset.bytes(), AbstractByte::Init(value, None));
         }
         bytes
     }
--- a/spec/lang/representation.md
+++ b/spec/lang/representation.md
@@ -230,55 +230,51 @@ impl Type {
 ### Enums
 
 Enum encoding and decoding.
-Should we let the type decoding throw UB?
+Note that the discriminant setting and reading does not modify any bytes.
+This is does not allow types like `Result<bool, bool>` to be stored in one byte.
+However this is fine as Rust currently does not do any optimizations like that
+and doing it here will introduce more questions.
 
 ```rust
 fn compute_discriminant<M: Memory>(bytes: List<AbstractByte<M::Provenance>>, discriminator: Discriminator) -> Option<Int> {
-    let mut disc = discriminator;
-    loop {
-        match disc {
-            Discriminator::Known(val) => break Some(val),
-            Discriminator::Unknown { offset, children } => {
-                if offset < 0 || offset >= bytes.len() { throw!(); }
-                let AbstractByte::Init(val, _) = bytes.get(offset).unwrap()
-                    else { break None };
-                    // else { throw_ub!("Encountered uninitialized byte while computing enum discriminant.") };
-                let Some(new_disc) = children.get(val)
-                    else { break None };
-                    // else { throw_ub!("Encountered invalid discriminant value.") };
-                disc = new_disc;
-            }
+    // FIXME: we have multiple quite different fail sources,
+    // it would be nice to return more error information.
+    match discriminator {
+        Discriminator::Known(val) => Some(val),
+        Discriminator::Invalid => None,
+        Discriminator::Unknown { offset, children, fallback } => {
+            let AbstractByte::Init(val, _) = bytes[offset]
+                else { return None };
+            let next_discriminator = children.get(val).unwrap_or(fallback);
+            compute_discriminant::<M>(bytes, next_discriminator)
         }
     }
 }
 
 impl Type {
-    fn decode<M: Memory>(Type::Enum { variants, tag_encoding, size, .. }: Self, bytes: List<AbstractByte<M::Provenance>>) -> Option<Value<M>> {
+    fn decode<M: Memory>(Type::Enum { variants, discriminator, size, .. }: Self, bytes: List<AbstractByte<M::Provenance>>) -> Option<Value<M>> {
         if bytes.len() != size.bytes() { throw!(); }
-        let disc = compute_discriminant::<M>(bytes, tag_encoding.discriminator)?;
+        let disc = compute_discriminant::<M>(bytes, discriminator)?;
 
-        // The discriminator does the discriminant check and as such should not be able to return an invalid value.
-        if disc < Int::ZERO || disc >= variants.len() { throw!(); }
-        variants.get(disc).unwrap().decode(bytes)
+        // Decode into the variant.
+        // Because the variant is the same size as the enum we don't need to pass a subslice.
+        let Some(value) = variants[disc].ty.decode(bytes)
+            else { return None };
+
+        Some(Value::Variant { idx: disc, data: value })
     }
 
-    fn encode<M: Memory>(Type::Enum { variants, tag_encoding, size, .. }: Self, val: Value<M>) -> List<AbstractByte<M::Provenance>> {
+    fn encode<M: Memory>(Type::Enum { variants, .. }: Self, val: Value<M>) -> List<AbstractByte<M::Provenance>> {
         let Value::Variant { idx, data } = val else { panic!() };
-        assert_eq!(variants.len(), tag_encoding.tagger.len());
 
-        let mut bytes = list![AbstractByte::Uninit; size.bytes()];
+        // idx is to be guaranteed in bounds by the well-formed check in the type.
+        let Variant { ty: variant, tagger } = variants[idx];
+        let mut bytes = variant.encode(data.extract());
 
-        // idx is to be guaranteed in bounds by the well-formed check in the typed store.
-        let variant = variants.get(idx).unwrap();
-        let encoded_data = variant.encode(data.extract());
-        assert!(encoded_data.len() <= size.bytes());
-        bytes.write_subslice_at_index(Int::ZERO, encoded_data);
-
-        let tagger = tag_encoding.tagger.get(idx).unwrap();
-
-        // TODO: is there a nicer way instead of all these temporary lists?
+        // write tag afterwards into the encoded bytes. This is fine as we don't allow
+        // encoded data and the tag to overlap at the moment.
         for (offset, value) in tagger.iter() {
-            bytes.set(offset, value);
+            bytes.set(offset, AbstractByte::Init(value, None));
         }
         bytes
     }
@@ -466,7 +462,7 @@ impl<M: Memory> AtomicMemory<M> {
             (Value::Tuple(vals), Type::Array { elem: ty, .. }) =>
                 Value::Tuple(vals.try_map(|val| self.retag_val(val, ty, fn_entry))?),
             (Value::Variant { idx, data }, Type::Enum { variants, .. }) =>
-                Value::Variant { idx, data: self.retag_val(data, variants[idx], fn_entry)? },
+                Value::Variant { idx, data: self.retag_val(data, variants[idx].ty, fn_entry)? },
             _ =>
                 panic!("this value does not have that type"),
         })
--- a/spec/lang/representation.md
+++ b/spec/lang/representation.md
@@ -230,15 +230,13 @@ impl Type {
 ### Enums
 
 Enum encoding and decoding.
-Note that the discriminant setting and reading does not modify any bytes.
-This is does not allow types like `Result<bool, bool>` to be stored in one byte.
-However this is fine as Rust currently does not do any optimizations like that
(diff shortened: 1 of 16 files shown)


spec/lang/well-formed.md
```diff
@@ -108,11 +108,14 @@ impl Type {
...
                 ensure(size >= last_end)?;
             }
-            Enum { variants, size, tag_encoding: _, align: _ } => {
+            Enum { variants, size, discriminator: _, align: _ } => {
+                // All the variants need to be well-formed and fit in the enum.
                 for variant in variants {
-                    variant.check_wf::<T>()?;
-                    ensure(size >= variant.size::<T>())?;
+                    variant.inner.check_wf::<T>()?;
+                    ensure(size >= variant.inner.size::<T>())?;
```
RalfJung, line 115: Didn't we agree last time to make this `==`?

spec/lang/representation.md
```diff
@@ -231,16 +231,58 @@ impl Type {
 
 ### Enums
 
-TODO: implement Enum decoding & encoding.
+Enum encoding and decoding.
+Note that the discriminant setting and reading does not modify any bytes.
+This is does not allow types like `Result<bool, bool>` to be stored in one byte.
+However this is fine as Rust currently does not do any optimizations like that
+and doing it here will introduce more questions.
```
RalfJung, line 238: I don't understand this paragraph. How does it nit modify any bytes...?

`Result<bool, bool>` can't be stored in one byte since the types have overlapping valid byte ranges.

spec/lang/representation.md
```diff
@@ -231,16 +231,58 @@ impl Type {
...
 
 ```rust
+fn compute_discriminant<M: Memory>(bytes: List<AbstractByte<M::Provenance>>, discriminator: Discriminator) -> Option<Int> {
```
RalfJung, line 241: What is this function doing? Please add a doc comment.

spec/lang/representation.md
```diff
@@ -231,16 +231,58 @@ impl Type {
...
+
 impl Type {
-    fn decode<M: Memory>(Type::Enum { .. }: Self, bytes: List<AbstractByte<M::Provenance>>) -> Option<Value<M>> {
-        todo!()
+    fn decode<M: Memory>(Type::Enum { variants, discriminator, size, .. }: Self, bytes: List<AbstractByte<M::Provenance>>) -> Option<Value<M>> {
+        if bytes.len() != size.bytes() { throw!(); }
+        let disc = compute_discriminant::<M>(bytes, discriminator)?;
+
+        // The discriminator does the discriminant check and as such should not be able to return an invalid value.
+        if disc < Int::ZERO || disc >= variants.len() { throw!(); }
```
RalfJung, line 266: This should not be needed, well-formedness should check this.

spec/lang/representation.md
```diff
@@ -231,16 +231,58 @@ impl Type {
...
+and doing it here will introduce more questions.
 
 ```rust
+fn compute_discriminant<M: Memory>(bytes: List<AbstractByte<M::Provenance>>, discriminator: Discriminator) -> Option<Int> {
+    let mut disc = discriminator;
+    loop {
+        match disc {
+            Discriminator::Known(val) => break Some(val),
+            Discriminator::Unknown { offset, children } => {
+                if offset < 0 || offset >= bytes.len() { throw!(); }
```
RalfJung, line 247: This should be checked already by well-formedness.

spec/lang/representation.md
```diff
@@ -231,16 +231,58 @@ impl Type {
...
+                let AbstractByte::Init(val, _) = bytes.get(offset).unwrap()
```
RalfJung, line 248:
```suggestion
                let AbstractByte::Init(val, _) = bytes[offset]
```
does this not work?

spec/lang/representation.md
```diff
@@ -231,16 +231,58 @@ impl Type {
...
-TODO: implement Enum decoding & encoding.
+Enum encoding and decoding.
+Note that the discriminant setting and reading does not modify any bytes.
+This is does not allow types like `Result<bool, bool>` to be stored in one byte.
+However this is fine as Rust currently does not do any optimizations like that
+and doing it here will introduce more questions.
 
 ```rust
+fn compute_discriminant<M: Memory>(bytes: List<AbstractByte<M::Provenance>>, discriminator: Discriminator) -> Option<Int> {
+    let mut disc = discriminator;
```
RalfJung, line 242: Please use recursion instead of mutable state with a loop.

spec/lang/well-formed.md
```diff
@@ -564,8 +567,10 @@ impl<M: Memory> Value<M> {
                 }
             }
             (Value::Variant { idx, data }, Type::Enum { variants, .. }) => {
-                ensure(idx < variants.len())?;
-                data.check_wf(variants[idx])?;
+                // It is undefined behaviour to generate a value of an uninhabited enum
+                if !ty.inhabited() { throw!() };
```
RalfJung, line 571: Why this special check? We don't have it for any of the other types...

RalfJung (commented):
Largely looks good to me, with a bunch of nits. :)

Also, could you add some sort of test?
- A program that has a local of type `!`, and then reads from it (should be UB since it's an invalid value)
- A program that has a local of type `u8`, writes a 0, then does a load and store at type `enum Bool { True, False }`, and checks that we get back the 0 we started with.

spec/lang/representation.md
```diff
@@ -231,16 +231,58 @@ impl Type {
...
     }
 
-    fn encode<M: Memory>(Type::Enum { .. }: Self, val: Value<M>) -> List<AbstractByte<M::Provenance>> {
-        todo!()
+    fn encode<M: Memory>(Type::Enum { variants, size, .. }: Self, val: Value<M>) -> List<AbstractByte<M::Provenance>> {
+        let Value::Variant { idx, data } = val else { panic!() };
+
+        let mut bytes = list![AbstractByte::Uninit; size.bytes()];
+
+        // idx is to be guaranteed in bounds by the well-formed check in the typed store.
```
RalfJung, line 275: Please add a comment explaining why we write the data first

spec/lang/types.md
```diff
@@ -81,10 +80,21 @@ pub struct IntType {
...
+    pub tagger: Map<Int, u8>,
+}
+
+/// The discrimination decision tree for an enum type.
+pub enum Discriminator {
+    Known(Int),
+    Unknown { // Branch
+        offset: Int,
+        children: Map<u8, Discriminator>,
+    },
```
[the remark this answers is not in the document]
digama0: It could be `Unknown { offset: 0, children: map!{} }`.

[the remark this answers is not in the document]
RalfJung: `offset` is out-of-bounds then, though.

RalfJung:
> [Merge branch 'master' into enum_representation](https://github.com/minirust/minirust/pull/146/commits/44ff95fd23e2cd34d0da53f71432526ce350f16d)

Please don't have merge commits in a PR; instead, when syncing with master is necessary (but only then), rebase over master.

spec/lang/types.md
```diff
@@ -81,10 +80,33 @@ pub struct IntType {
...
+/// The decision tree that computes the discriminant out of the tag for a specific
+/// enum type.
+pub enum Discriminator {
+    /// We know the discriminant.
+    Known(Int),
+    /// Tag decoding failed, there is no valid discriminant.
+    Invalid,
+    /// We don't know the discriminant, so we branch on the value of a specific byte.
+    /// The fallback is for readability, as we often are only interested in a couple
+    /// of values.
```
RalfJung, line 102:
```suggestion
    /// The fallback keeps the representation more compact, as we often
    ///  are only interested in a couple of values and we don't want to always
    /// have 256 branches.
```

spec/lang/well-formed.md
```diff
@@ -108,17 +108,40 @@ impl Type {
...
+                // we don't have to handle different sizes in the memory representation.
                 for variant in variants {
-                    variant.check_wf::<T>()?;
-                    ensure(size >= variant.size::<T>())?;
+                    variant.ty.check_wf::<T>()?;
+                    ensure(size == variant.ty.size::<T>())?;
                 }
+
+                // check that all variants reached by the discriminator are valid and
+                // that it never accesses out-of-bounds area.
```
RalfJung, line 120:
```suggestion
                // that it never performs out-of-bounds accesses.
```

RalfJung:
Thanks! What's here looks good now. But there should be some testcases that hit the new code.

essickmango:
PR ready for review

spec/lang/representation.md
```diff
@@ -229,16 +229,54 @@ impl Type {
...
+
+        Some(Value::Variant { idx: disc, data: value })
     }
 
-    fn encode<M: Memory>(Type::Enum { .. }: Self, val: Value<M>) -> List<AbstractByte<M::Provenance>> {
-        todo!()
+    fn encode<M: Memory>(Type::Enum { variants, .. }: Self, val: Value<M>) -> List<AbstractByte<M::Provenance>> {
+        let Value::Variant { idx, data } = val else { panic!() };
+
+        // idx is to be guaranteed in bounds by the well-formed check in the type.
```
RalfJung, line 270:
```suggestion
        // `idx` is guaranteed to be in bounds by the well-formed check in the type.
```

spec/lang/representation.md
```diff
@@ -229,16 +229,54 @@ impl Type {
...
+        let Variant { ty: variant, tagger } = variants[idx];
+        let mut bytes = variant.encode(data.extract());
+
+        // write tag afterwards into the encoded bytes. This is fine as we don't allow
```
RalfJung, line 274:
```suggestion
        // Write tag into the encoded bytes. This is fine as we don't allow
```

spec/lang/representation.md
```diff
@@ -267,13 +266,14 @@ impl Type {
...
+        // `idx` is guaranteed to be in bounds by the well-formed check in the type.
         let Variant { ty: variant, tagger } = variants[idx];
         let mut bytes = variant.encode(data.extract());
 
-        // write tag afterwards into the encoded bytes. This is fine as we don't allow
-        // encoded data and the tag to overlap at the moment.
+        // Write tag into the bytes around the data.
+        // This is fine as we don't allow encoded data and the tag to overlap.
         for (offset, value) in tagger.iter() {
+            assert_eq!(Some(AbstractByte::Uninit), bytes.get(offset.bytes()));
```
essickmango, line 276: For some reason it doesn't compile when I use `bytes[offset.bytes()]` even though I am indexing in a `List` using an `Int`

essickmango: Otherwise this PR is ready for review again.

RalfJung: > doesn't compile

What's the error message? Don't just say "it doesn't work", always explain what happens instead. :)

spec/lang/representation.md
```diff
@@ -229,16 +229,54 @@ impl Type {
...
+        let Value::Variant { idx, data } = val else { panic!() };
+
+        // `idx` is guaranteed to be in bounds by the well-formed check in the type.
+        let Variant { ty: variant, tagger } = variants[idx];
+        let mut bytes = variant.encode(data.extract());
+
+        // Write tag into the bytes around the data.
+        // This is fine as we don't allow encoded data and the tag to overlap.
+        for (offset, value) in tagger.iter() {
+            assert_eq!(Some(AbstractByte::Uninit), bytes.get(offset.bytes()));
```
RalfJung, line 276: This assertion is somewhat dubious. We should have the invariant that "if your code passes well-formedness then you will never get a panic", but here, this panic is not excluded by the WF-check.

So I think this should be just a comment.
```suggestion
            // If this enum type is "reasonable", we'll only overwrite `Uninit` here`.
            // However check_wf cannot ensure this, so we do not have an assertion.
```

spec/lang/representation.md
```diff
@@ -229,16 +229,54 @@ impl Type {
...
 ### Enums
 
-TODO: implement Enum decoding & encoding.
+Enum encoding and decoding.
+Note that the discriminant may not be written into bytes that contain encoded data.
+This is to ensure that pointers to the data always contain valid values.
 
 ```rust
+/// Computes the discriminant from the value's bytes by the decision tree in the `Discriminator` by
+/// recursing over the branches defined by `Discriminator::Unknown`.
```
RalfJung, line 238:
```suggestion
/// Reads the discriminant from the representation bytes, using the given `Discriminator`.
```

spec/lang/representation.md
```diff
@@ -229,16 +229,54 @@ impl Type {
...
+/// FIXME: we have multiple quite different fail sources, it would be nice to return more error information.
+fn compute_discriminant<M: Memory>(bytes: List<AbstractByte<M::Provenance>>, discriminator: Discriminator) -> Option<Int> {
```
RalfJung, line 240: "compute" is not entirely accurate, we're not doing much computation here after all. I think this should be called "decode_discriminant"; that also makes it clear which operation this is a helper for.

spec/lang/representation.md
essickmango: "cannot index into a value of type `libspecr::List<mem::AbstractByte<<M as mem::Memory>::Provenance>>`".
Reading further into this, apparently specr doesn't transpile correctly in the case of `assert_eq!(AbstractByte::Uninit, bytes[offset.bytes()]);`. I noted this for myself to maybe fix it in the future, should I make an issue in the minirust-tooling repo?

RalfJung: Ah, is the macro getting in the way? That's possible. 

Sure, please file an issue! This will likely be non-trivial to fix though, and it's not very important.

essickmango: Changed everything else as requested.

RalfJung:
Looks good. :) Please squash the history a bit, this change doesn't need 15 commits. Then I'll land this :D

essickmango:
squashed them

RalfJung:
Thanks :)

PR #147, opened 2023-11-01
Offset type alias

essickmango:
Introduces an offset type alias so we don't have a list of fields that is of type `List<(Size, Size)>` but `List<(Offset, Size)>`.

Changed all instances of `Size` to `Offset` in the MiniRust spec, miniutil and minimizer where applicable.

### DIFF
--- a/spec/prelude/main.md
+++ b/spec/prelude/main.md
@@ -9,6 +9,9 @@ pub use libspecr::prelude::*;
 /// Make the two main modules available.
 pub use crate::{lang, mem};
 
+/// When a non-negative integer is used as an offset into an allocation or type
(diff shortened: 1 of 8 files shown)


spec/prelude/main.md
```diff
@@ -9,6 +9,8 @@ pub use libspecr::prelude::*;
 /// Make the two main modules available.
 pub use crate::{lang, mem};
 
+pub type Offset = Size;
```
RalfJung, line 12:
```suggestion
/// When a non-negative integer is used as an offset into an allocation or type
/// rather than to describe the size of an object or type, use this type instead
/// of `Size` for extra clarity.
pub type Offset = Size;
```

PR #150, opened 2023-12-16
Switch terminator

essickmango:
Replaces `Terminator::If` with `Terminator::Switch` for a more general branching terminator that now also supports enum discriminants.

As a side effect we now support int matching in the minimizer. I didn't look into the Rust switch on char so I didn't add support, could be trivial though.


spec/lang/syntax.md
```diff
@@ -239,12 +239,13 @@ pub enum Statement {
 pub enum Terminator {
     /// Just jump to the next block.
     Goto(BbName),
-    /// `condition` must evaluate to a `Value::Bool`.
-    /// If it is `true`, jump to `then_block`; else jump to `else_block`.
-    If {
-        condition: ValueExpr,
-        then_block: BbName,
-        else_block: BbName,
+    /// `value` needs to either evaluate to a `Value::Bool` or `Value::Int`.
```
RalfJung, line 242: I think we want this to be only for `Int`. We can add a bool-to-int cast for the `bool` case.

tooling/minimize/intrinsics/Cargo.toml
```diff
@@ -7,3 +7,6 @@ edition = "2021"
 
 [lib]
 crate-type = ["staticlib", "lib"]
+
+[workspace]
+members = ["."]
```
RalfJung, line 12: What's up with this diff?

If the goal is to add this to the workspace, then that's not what you did. You made this a new workspace instead I think. Add it in `tooling/Cargo.toml` instead, please.

tooling/miniutil/src/fmt/function.rs
```diff
@@ -159,20 +159,21 @@ fn fmt_terminator(t: Terminator, comptypes: &mut Vec<CompType>) -> String {
...
+            fallback,
         } => {
-            let branch_expr = fmt_value_expr(condition, comptypes).to_string();
-            let then_bb = fmt_bb_name(then_block).to_string();
-            let else_bb = fmt_bb_name(else_block).to_string();
+            let branch_expr = fmt_value_expr(value, comptypes).to_string();
+            let mut case_strs = cases.iter()
+                                     .map(|(constant, successor)| format!("{}: {}", fmt_constant(constant).to_string(), fmt_bb_name(successor)))
+                                     .collect::<Vec<String>>();
+            case_strs.push(format!("otherwise: {}", fmt_bb_name(fallback)));
```
RalfJung, line 171: Maybe we should spell this `default:` like in C?

essickmango: I based the formatting on how Rust represents MIR. It uses `otherwise`.

tooling/minimize/intrinsics/Cargo.toml
essickmango: Required for the rust.sh script in minimizer/tests to work, but will be implemented in a seperate PR.

spec/lang/syntax.md
essickmango: Implemented in the last commit.

spec/lang/step/terminators.md
```diff
@@ -31,15 +31,15 @@ impl<M: Memory> Machine<M> {
...
+## Switch
 
 ```rust
 impl<M: Memory> Machine<M> {
-    fn eval_terminator(&mut self, Terminator::If { condition, then_block, else_block }: Terminator) -> NdResult {
-        let (Value::Bool(b), _) = self.eval_value(condition)? else {
-            panic!("if on a non-boolean")
+    fn eval_terminator(&mut self, Terminator::Switch { value, cases, fallback }: Terminator) -> NdResult {
+        let Value::Int(value) = self.eval_value(value)?.0 else {
+            panic!("switch on a non-boolean");
```
RalfJung, line 40:
```suggestion
            panic!("switch on a non-integer");
```

spec/lang/well-formed.md
```diff
@@ -434,10 +438,26 @@ impl Terminator {
...
+                };
+
+                // gets the successor blocks of all cases that are valid.
+                let mut next_blocks = cases.iter().filter_map(|(case, block)| {
+                    ensure(switch_ty.can_represent(case))?;
+                    Some(block)
+                }).collect::<List<BbName>>();
+
+                // ensures that all cases were valid.
+                ensure(next_blocks.len() == cases.len())?;
```
RalfJung, line 457: A `for` loop would be easier to understand than doing a `filter_map` with a length check.

tooling/minimize/intrinsics/Cargo.lock
RalfJung, line 1: This diff is still a leftover from the rust.sh script changes I think.

tooling/minimize/src/bb.rs
```diff
@@ -70,19 +70,32 @@ fn translate_terminator<'cx, 'tcx>(
...
+            // FIXME: add support for switching on `char`
+            assert!(ty.is_bool() || ty.is_integral());
 
-            let else_block = targets.target_for_value(0);
-            let else_block = fcx.bb_name_map[&else_block];
+            // If the value is a boolean we need to cast it to an integer first as MiniRust switch only operates on ints.
+            let discr_op = translate_operand(discr, fcx);
+            let value = if ty.is_bool() {
+                ValueExpr::UnOp {
+                    operator: UnOp::BoolToIntCast(IntType { signed: Signedness::Unsigned, size: Size::from_bytes(1).unwrap() }),
```
RalfJung, line 83: `<u8>::get_type()` would be easier than constructing the `IntType` by hand

tooling/miniutil/src/fmt/function.rs
RalfJung: Okay let's be consistent with MIR then.

tooling/minimize/src/bb.rs
essickmango: I agree, though I'd have to extract the `IntType` out of the returned `Type`. However this gives us a bit more future-proofness and readability so it is worth it in my opinion.

RalfJung: Ah you are right, this needs an `IntType`, not a `Type`...

Can't this use the `if_` function from miniutil?

RalfJung: > Can't this use the if_ function from miniutil?

Hm, that doesn't fit the MIR structure very well. Let's stick to what you got.

PR #152, opened 2024-01-10
Adds enum type minimization without support for niche tag encoding

essickmango:
Adds enum type minimization without support for niche tag encoding.
Adds minimization of all other enum-related operations, i.e. `Assign(Aggregate(AdtKind::Enum ...)))`, `SetDiscriminant`, `GetDiscriminant`, Variant `Downcast`.
Adds a test that enums generally can be minimized and matching on enums works.


spec/lang/types.md
```diff
@@ -105,14 +105,15 @@ pub enum Discriminator {
     Known(Int),
     /// Tag decoding failed, there is no valid discriminant.
     Invalid,
-    /// We don't know the discriminant, so we branch on the value of a specific byte.
+    /// We don't know the discriminant, so we branch on the value of a specific value.
     /// The fallback keeps the representation more compact, as we often are only
-    /// interested in a couple of values and we don't want to always have 256 branches.
+    /// interested in a couple of values and don't want to always have > 256 branches.
```
RalfJung, line 110: I think with the int branching, we can just remove this last sentence. The non-fallback representation gets entirely ludicrous for types bigger than a byte.

spec/lang/well-formed.md
```diff
@@ -114,11 +114,18 @@ impl Type {
             Enum { variants, size, discriminator, discriminant_ty, .. } => {
                 // All the variants need to be well-formed and be the size of the enum so
                 // we don't have to handle different sizes in the memory representation.
-                // Also their alignment may not be larger than the total enum alignment.
+                // Also their alignment may not be larger than the total enum alignment and
+                // all the values written by the tagger must fit into the variant.
+                // FIXME (Timon): should we require alignment checks for the values written in the taggers?
```
RalfJung, line 119: I don't understand the fixme. Do you mean having a separate notion of alignment for the tagger? That would require changes in the type itself as well? I don't think we want or need that. But we'll have to make sure in minimize that we represent whatever alignment restriction might exist on the discriminant field.

spec/lang/well-formed.md
```diff
@@ -137,10 +144,14 @@ impl Discriminator {
         match self {
             Discriminator::Known(variant) => ensure(variant >= Int::ZERO && variant < n_variants && discriminant_ty.can_represent(variant)),
             Discriminator::Invalid => ret(()),
-            Discriminator::Branch { offset, fallback, children } => {
-                ensure(offset < size)?;
+            Discriminator::Branch { offset, value_type, fallback, children } => {
+                // FIXME (Timon): Do we require a alignment check on the branching value?
+                // Ensure that the value we branch on is in bounds and that all children all valid.
```
RalfJung, line 149:
```suggestion
                // Ensure that the value we branch on is stored in bounds and that all children all valid.
```

spec/lang/step/statements.md
```diff
@@ -77,10 +77,9 @@ impl<M: Memory> Machine<M> {
 
         // Write the tag directly into memory.
         // This should be fine as we don't allow encoded data and the tag to overlap for valid enum variants.
-        let accessor = |offset: Offset, value| {
+        let accessor = |offset: Offset, value_type: IntType, bytes| {
             let ptr = self.ptr_offset_inbounds(place.ptr, offset.bytes())?;
-            // We have ensured that the place is aligned, so no alignment requirement here
-            self.mem.store(ptr, [value].into_iter().collect(), Align::ONE, Atomicity::None)
```
RalfJung, line 83: I think you can still use ONE here, since as before we have ensured that the place is aligned.

spec/lang/representation.md
```diff
@@ -238,13 +238,14 @@ This is to ensure that pointers to the data always contain valid values.
...
-fn decode_discriminant<M: Memory>(mut accessor: impl FnMut(Offset) -> Result<AbstractByte<M::Provenance>>, discriminator: Discriminator) -> Result<Option<Int>> {
+fn decode_discriminant<M: Memory>(mut accessor: impl FnMut(Offset, IntType) -> Result<List<AbstractByte<M::Provenance>>>, discriminator: Discriminator) -> Result<Option<Int>> {
     match discriminator {
         Discriminator::Known(val) => ret(Some(val)),
         Discriminator::Invalid => ret(None),
-        Discriminator::Branch { offset, children, fallback } => {
-            let AbstractByte::Init(val, _) = accessor(offset)?
-                else { return ret(None) };
+        Discriminator::Branch { offset, value_type, children, fallback } => {
+            let bytes = accessor(offset, value_type)?;
```
RalfJung, line 246: It should be sufficient to pass the size to the accessor. (Same for `encode`.)

spec/lang/representation.md
```diff
@@ -238,13 +238,14 @@ This is to ensure that pointers to the data always contain valid values.
 /// The accessor is given an offset relative to the beginning of the encoded enum value,
```
RalfJung, line 238:
```suggestion
/// Returns `Ok(None)` when reaching `Discriminator::Invalid` and when any of the reads
/// for `Discriminator::Branch` encounters uninitialized memory.
/// Returns `Err` only if `accessor` returns `Err`.
///
/// The accessor is given an offset relative to the beginning of the encoded enum value,
```

spec/lang/representation.md
```diff
@@ -253,10 +254,14 @@ fn decode_discriminant<M: Memory>(mut accessor: impl FnMut(Offset) -> Result<Abs
 
 /// Writes the tag described by the tagger into the bytes accessed using the accessor.
```
RalfJung, line 255:
```suggestion
/// Writes the tag described by the tagger into the bytes accessed using the accessor.
/// Returns `Err` only if `accessor` returns `Err`.
///
```

spec/lang/step/expressions.md
```diff
@@ -106,15 +106,15 @@ impl<M: Memory> Machine<M> {
...
         }
 
         // We don't require the variant to be valid,
         // we are only interested in the bytes that the discriminator actually touches.
-        let accessor = |idx: Offset| {
+        let accessor = |idx: Offset, value_type: IntType| {
             let ptr = self.ptr_offset_inbounds(place.ptr, idx.bytes())?;
-            let byte = self.mem.load(ptr, Size::from_bytes(1).unwrap(), Align::ONE, Atomicity::None)?;
-            ret(byte[Int::ZERO])
+            // FIXME (Timon): Do we need an alignment check on the tag values?
```
RalfJung, line 116: No, we have already checked alignment above.

tooling/minimize/src/enums.rs
```diff
@@ -0,0 +1,91 @@
...
+    sref: rs::GenericArgsRef<'tcx>,
+    tcx: rs::TyCtxt<'tcx>,
+) -> Type {
+    let a = rs::ParamEnv::empty().and(ty);
+    let layout = tcx.layout_of(a).unwrap().layout;
+    let size = translate_size(layout.size());
+    let align = translate_align(layout.align().abi);
+
+    // The discriminant type in Rust is always isize which always gets translated as i64.
+    let discriminant_ty = IntType { signed: Signedness::Signed, size: Size::from_bytes_const(8) };
```
RalfJung, line 15: This doesn't seem right, see the `discriminant_ty` function in rustc.

tooling/minimize/src/enums.rs
```diff
@@ -0,0 +1,91 @@
...
+        rs::Variants::Multiple {
+            tag,
+            tag_encoding,
+            tag_field,
+            variants,
+        } => {
+            
+            // compute the offset of the tag for the tagger and discriminator construction
+            let tag_offset: Offset = translate_size(layout.fields().offset(*tag_field));
+            // compute the tag type (which will always be unsigned as we get the discriminant as u128).
```
RalfJung, line 31: Tags can be signed. Why do you make it always unsigned?

tooling/minimize/src/enums.rs
```diff
@@ -0,0 +1,91 @@
...
+            let tag_ty = match tag.primitive() {
+                rs::Primitive::Int(ity, _signed) => IntType { signed: Signedness::Unsigned, size: translate_size(ity.size()) },
+                _ => panic!("enum tag has invalid primitive type"),
+            };
+
+            // translate the variants
+            let translated_variants = adt_def.variants().iter_enumerated().map(|(variant_idx, variant_def)| {
+                let fields = translate_fields(&variants[variant_idx].fields, &variant_def, sref, tcx);
+                let discr = adt_def.discriminant_for_variant(tcx, variant_idx);
+                let tag_bits = discr.val & (u128::MAX >> (128 - tag_ty.size.bits().try_to_u8().unwrap()));
```
RalfJung, line 41: This bit magic needs explanation. But I think i should not be needed.

tooling/minimize/src/enums.rs
```diff
@@ -0,0 +1,91 @@
...
+    let discriminant_ty = IntType { signed: Signedness::Signed, size: Size::from_bytes_const(8) };
+
+    let (variants, discriminator) = match layout.variants() {
+        rs::Variants::Single { index } => {
+            let fields = translate_fields(layout.fields(), adt_def.variant(*index), sref, tcx);
+            ([Variant { ty: Type::Tuple { fields, size, align }, tagger: Map::new() }].into_iter().collect(), Discriminator::Known(0.into()))
+        },
+        rs::Variants::Multiple {
+            tag,
+            tag_encoding,
```
RalfJung, line 24: I think you'll want to do `tag_encoding: rs::TagEncoding::Direct` here; the niche one will need quite a bit of different logic probably.

tooling/minimize/src/enums.rs
```diff
@@ -0,0 +1,91 @@
...
+                    rs::TagEncoding::Niche { .. } => todo!("Implement Niche-encoded tags for enums (Timon)"),
+                };
+                Variant { ty: Type::Tuple { fields, size, align }, tagger }
+            }).collect();
+
+            // build the discriminator.
+            let discriminator = match tag_encoding {
+                rs::TagEncoding::Direct => {
+                    let children = adt_def.discriminants(tcx)
+                        .map(|(idx, discr)| (Int::from(discr.val & (u128::MAX >> (128 - tag_ty.size.bits().try_to_u8().unwrap()))), Discriminator::Known(idx.index().into())))
```
RalfJung, line 53: Instead of redoing the math here, wouldn't it be easier to build up the tagger and the discriminator together in the variant loop above? For direct tags, they are very similar.

tooling/minimize/tests/pass/enum_direct_tag.rs
```diff
@@ -0,0 +1,31 @@
...
+}
+
+fn main() {
+    let x = A::A1(12);
+    print_a(&x);
+    print_a_match(&x);
+    let x = A::A2;
+    print_a(&x);
+    print_a_match(&x);
+}
```
RalfJung, line 31: Missing newline at end of file

RalfJung (commented):
The first commit largely looks good. The minimize support still needs a bit of work I assume. :)

Looking at this file may be helpful:
https://github.com/rust-lang/rust/blob/fcb42b42d6fc078b15a34ed27b969fe2ba290989/compiler/rustc_const_eval/src/interpret/discriminant.rs

Please make sure you also add a test with an enum like this:
```rust
#[repr(i32)]
enum E { A = -1, B(u16) = 1 }
```

spec/lang/well-formed.md
essickmango: Rust does alignment checks upon loading the tag, and forces types like in the example to have 2 bytes alignment.
```rust
enum A {
    A1 = 0,
    A2 = 6000,
}
```

However I think this is mainly because of how some targets need their loads aligned and may be too restrictive to make this a thing in MiniRust.

RalfJung: We have already checked that the surrounding place is sufficiently aligned. That also implicitly ensures that the tag is aligned, so we don't need a separate check here.

RalfJung: > However I think this is mainly because of how some targets need their loads aligned and may be too restrictive to make this a thing in MiniRust.

MiniRust absolutely needs to check everything required for any target. Otherwise the compiler to the target has to be able to deal with unaligned loads. After all, we can't have code be okay for MiniRust but then fail on the target, that would be a critical bug.

RalfJung: No. The offset at which the values are read/written *defines* the alignment requirement.

There are Rust types where a 2-byte discriminant resides at an odd offset, i.e., it is 1-aligned. This is completely fine and the codegen backend knows how to handle it.

essickmango: Oh, okay... The alignment of the tag may only have an influence with direct tagging, I see.

RalfJung: No. It's more like, the offset of the tag in the surrounding enum *defines* which alignment guarantees there are, and the compiler computes the LLVM IR alignment annotations from that.

essickmango:
I largely redid the PR, it is back to you @RalfJung. (I resolved all the conversations to the old version since they no longer apply / are included)

spec/lang/step/terminators.md
```diff
@@ -108,11 +108,14 @@ fn check_abi_compatibility(
         (Type::Enum { variants: caller_variants, discriminator: caller_discriminator, discriminant_ty: caller_discriminant_ty, size: caller_size, align: caller_align },
          Type::Enum { variants: callee_variants, discriminator: callee_discriminator, discriminant_ty: callee_discriminant_ty, size: callee_size, align: callee_align }) =>
             caller_variants.len() == callee_variants.len() &&
-            caller_variants.iter().zip(callee_variants.iter()).all(|((caller_discriminant, caller_variant), (callee_discriminant, callee_variant))|
-                caller_discriminant == callee_discriminant &&
+            // Checking only one side is enough as we checked that they have the same amount of key-value pairs.
```
RalfJung, line 111:
```suggestion
            // Both sides must have the same discriminants. Checking only one direction
            // of the mutual inclusion enough as we checked that they have the same number of key-value pairs.
```

tooling/minimize/src/enums.rs
```diff
@@ -0,0 +1,121 @@
...
+    let align = translate_align(layout.align().abi);
+
+    let Type::Int(discriminant_ty) = translate_ty(ty.discriminant_ty(tcx), tcx) else {
+        panic!("Discriminant type is not integer!")
+    };
+
+    let (variants, discriminator) = match layout.variants() {
+        rs::Variants::Single { index } => {
+            let fields = translate_fields(layout.fields(), adt_def.variant(*index), sref, tcx);
+            ([(Int::ZERO, Variant { ty: Type::Tuple { fields, size, align }, tagger: Map::new() })].into_iter().collect(), Discriminator::Known(0.into()))
```
RalfJung, line 25:
```suggestion
            let variants = [(Int::ZERO, Variant { ty: Type::Tuple { fields, size, align }, tagger: Map::new() })];
            let discriminator = Discriminator::Known(0.into());
            (variants.into_iter().collect::<Map<_, _>>(), discriminator)
```
That makes this a lot more readable

tooling/minimize/src/enums.rs
```diff
@@ -0,0 +1,121 @@
...
+            tag,
+            tag_encoding,
+            tag_field,
+            variants,
+        } => {
+
+            // compute the offset of the tag for the tagger and discriminator construction
+            let tag_offset: Offset = translate_size(layout.fields().offset(*tag_field));
+            let tag_ty = match tag.primitive() {
+                rs::Primitive::Int(ity, signed) => IntType { signed: if signed { Signedness::Signed } else { Signedness::Unsigned }, size: translate_size(ity.size()) },
```
RalfJung, line 37: Shouldn't there already be some logic somewhere that can translate Rust types to MiniRust types, and that you can just call here?

tooling/minimize/src/enums.rs
```diff
@@ -0,0 +1,121 @@
...
+            let mut translated_variants = Map::new();
+            let mut discriminator_children = Map::new();
+            for (variant_idx, variant_def) in adt_def.variants().iter_enumerated() {
+                let fields = translate_fields(&variants[variant_idx].fields, &variant_def, sref, tcx);
+                let discr = adt_def.discriminant_for_variant(tcx, variant_idx);
+                let discr_int = int_from_bits(discr.val, discriminant_ty);
+                let (tagger, tag) = match tag_encoding {
+                    rs::TagEncoding::Direct => (
+                        // direct tagging places the discriminant in the tag for all variants
+                        [(tag_offset, (tag_ty, discr_int))].into_iter().collect(),
```
RalfJung, line 51: Wow there's some magic type inference going on here.
```suggestion
                        [(tag_offset, (tag_ty, discr_int))].into_iter().collect::<Map<_, _>>(),
```

tooling/minimize/src/enums.rs
```diff
@@ -0,0 +1,121 @@
...
+
+                (offset, ty)
+    }).collect()
+}
+
+pub fn int_from_bits(bits: u128, ity: IntType) -> Int {
+    let n_bits = ity.size.bits().try_to_u8().unwrap();
+    let rs_size = rs::Size::from_bits(n_bits);
+    let sign_extended = rs_size.sign_extend(bits);
+    if ity.signed == Signedness::Signed && sign_extended >> (n_bits - 1) != 0 {
```
RalfJung, line 104: No such shifts should be needed, what is going on here?

I think this should roughly be
```
if ity.signed == Signedness::Unsigned {
  Int::from(bits)
} else {
  let signed_val = rs_size.sign_extend(bits) as i128;
  Int::from(signed_val)
}
```

tooling/minimize/src/rvalue.rs
```diff
@@ -118,7 +118,15 @@ pub fn translate_rvalue<'cx, 'tcx>(
                     let ops: List<_> = operands.iter().map(|x| translate_operand(x, fcx)).collect();
                     ValueExpr::Tuple(ops, ty)
                 }
-                Type::Enum { .. } => todo!(),
+                Type::Enum { variants, .. } => {
+                    let rs::AggregateKind::Adt(_, variant_idx, _, _, _) = agg else { panic!() };
+                    let discriminant = discriminant_for_variant(rv.ty(&fcx.body, fcx.cx.tcx), fcx.cx.tcx, *variant_idx);
+                    let ops: List<_> = operands.iter().map(|x| translate_operand(x, fcx)).collect();
+
+                    // We have a guaranteed tuple as data when the type was minified from Rust.
```
RalfJung, line 126:
```suggestion
                    // We represent the multiple fields of an enum variant as a MiniRust tuple.
```

tooling/minimize/tests/pass/enum_direct_tag.rs
```diff
@@ -0,0 +1,64 @@
...
+    let x = A::A2;
+    print_a(&x);
+    print_a_match(&x);
+
+    print_i16_repr(I16Repr::Min);
+    print_i16_repr(I16Repr::Minus1);
+    print_i16_repr(I16Repr::Zero);
+    print_i16_repr(I16Repr::Max);
+
+    // While this is not going to run it is forcing the minimizer to minimize `!`.
```
RalfJung, line 56: This is not a "direct tag enum" though. Please move this part to a separate file, `never.rs`.

RalfJung (approved):
LGTM, thanks!

PR #153, opened 2024-01-26
Enum discriminant changes

essickmango:
Various changes around how MiniRust handles the discriminant.

- The `Discrimininator` now operates on arbitrary integer types, not only bytes.
- Replaces a list holding the variants in a map from discriminant to variant, completely removing the concept of a variant index in MiniRust. The minimizer will have to translate where rustc uses the variant index but I have checked all places known to me that this is possible.
- Adds a miniutil build interface for the Discriminator, removing a lot of explicit `Int` conversions in the test cases while also ensuring that we do not use invalid integers accidentally.
- Adds test cases for some situations that weren't tested previously.

### DIFF
--- a/spec/lang/step/expressions.md
+++ b/spec/lang/step/expressions.md
@@ -106,15 +106,15 @@ impl<M: Memory> Machine<M> {
             panic!("ValueExpr::GetDiscriminant requires enum type");
         };
         if !place.aligned {
-            throw_ub!("Getting the discriminant of a place based on a misaligned pointer");
+            throw_ub!("Getting the discriminant of a place based on a misaligned pointer.");
         }
 
         // We don't require the variant to be valid,
         // we are only interested in the bytes that the discriminator actually touches.
-        let accessor = |idx: Offset| {
+        let accessor = |idx: Offset, value_type: IntType| {
             let ptr = self.ptr_offset_inbounds(place.ptr, idx.bytes())?;
-            let byte = self.mem.load(ptr, Size::from_bytes(1).unwrap(), Align::ONE, Atomicity::None)?;
-            ret(byte[Int::ZERO])
+            // FIXME (Timon): Do we need an alignment check on the tag values?
+            self.mem.load(ptr, value_type.size, Align::ONE, Atomicity::None)
         };
         let Some(discriminant) = decode_discriminant::<M>(accessor, discriminator)? else {
             throw_ub!("ValueExpr::GetDiscriminant encountered invalid discriminant.");
--- a/spec/lang/step/expressions.md
+++ b/spec/lang/step/expressions.md
@@ -88,8 +88,8 @@ impl<M: Memory> Machine<M> {
 
 ```rust
 impl<M: Memory> Machine<M> {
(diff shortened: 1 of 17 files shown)


spec/lang/types.md
```diff
@@ -55,13 +55,13 @@ pub enum Type {
         align: Align,
     },
     Enum {
-        /// Each variant is given by a type and its tag description. All variants are thought
-        /// to "start at offset 0"; if the discriminant is encoded as an explicit tag,
-        /// then that will be put into the padding of the active variant. (This means it
-        /// is *not* safe to hand out mutable references to a variant at that type, as
-        /// then the tag might be overwritten!)
+        /// The map of discriminants to variants. Each variant is given by a type and its
```
RalfJung, line 58:
```suggestion
        /// The map variants, each identified by a discriminant. Each variant is given by a type and its
```

spec/lang/well-formed.md
```diff
@@ -114,17 +114,25 @@ impl Type {
...
+                    ensure(discriminant_ty.can_represent(discriminant))?;
+
                     variant.ty.check_wf::<T>()?;
                     ensure(size == variant.ty.size::<T>())?;
                     ensure(variant.ty.align::<T>().bytes() <= align.bytes())?;
+                    ensure(variant.tagger.iter().all(|(offset, (value_type, value))|
+                        value_type.check_wf().is_some() &&
+                        value_type.can_represent(value) &&
+                        offset + value_type.size <= size
+                    ))?;
```
RalfJung, line 129:
```suggestion
                    ))?;
                    // FIXME: check that the values written by the tagger do not overlap.
```
No need to add this check in this PR, but eventually we might want to do this.

spec/lang/step/statements.md
```diff
@@ -77,10 +77,9 @@ impl<M: Memory> Machine<M> {
 
         // Write the tag directly into memory.
         // This should be fine as we don't allow encoded data and the tag to overlap for valid enum variants.
-        let accessor = |offset: Offset, value| {
+        let accessor = |offset: Offset, bytes| {
             let ptr = self.ptr_offset_inbounds(place.ptr, offset.bytes())?;
-            // We have ensured that the place is aligned, so no alignment requirement here
```
RalfJung, line 82: This comment should be preserved.

spec/lang/step/expressions.md
```diff
@@ -106,15 +106,14 @@ impl<M: Memory> Machine<M> {
...
         }
 
         // We don't require the variant to be valid,
         // we are only interested in the bytes that the discriminator actually touches.
-        let accessor = |idx: Offset| {
+        let accessor = |idx: Offset, size: Size| {
             let ptr = self.ptr_offset_inbounds(place.ptr, idx.bytes())?;
-            let byte = self.mem.load(ptr, Size::from_bytes(1).unwrap(), Align::ONE, Atomicity::None)?;
-            ret(byte[Int::ZERO])
+            self.mem.load(ptr, size, Align::ONE, Atomicity::None)
```
RalfJung, line 116:
```suggestion
            // We have ensured that the place is aligned, so no alignment requirement here.
            self.mem.load(ptr, size, Align::ONE, Atomicity::None)
```

essickmango:
Ready for review again

RalfJung:
Looks good, thanks!

PR #154, opened 2024-02-12
Implement niche tagged enums in the minimizer

essickmango:
Implements niche tagged enums in the minimizer and adds some tests for it.

Also adds one special case of constant enum expressions that is sometimes used for `None::<NonZeroU8>` values. It doesn't nicely generalize though as we get the bits of the value while we don't have machine information for decoding.


RalfJung (commented):
Thanks, looks good overall. :)

tooling/minimize/tests/pass/enum_niche_tag.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    print_bool_result(Ok(true));
+    print_bool_result(Ok(false));
+    print_bool_result(Err(()));
+
+    print_option_ref(Some(&42));
+    print_option_ref(None);
+
+    let iter = RepeatN { val: 3, repetitions: 2 };
+    for i in iter {
```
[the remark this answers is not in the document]
essickmango: It kinda does. However we have some strange allocations for ranges that fail in the minimizer, so the following still doesn't work:
```rust
for i in 0..3 {
    print(i);
}
```

tooling/minimize/tests/pass/iter.rs
```diff
@@ -0,0 +1,31 @@
...
+    }
+}
+
+fn main() {
+    let iter = RepeatN {
+        val: false,
+        repetitions: 3,
+    };
+    for _ in iter {
+        print(1);
```
RalfJung, line 29: Can you assert that you get the right value?

tooling/minimize/src/constant.rs
```diff
@@ -31,6 +31,42 @@ fn translate_const_val<'cx, 'tcx>(
...
+        Type::Enum { size, discriminator, variants, .. } => {
+            let Some(bits) = val.try_to_bits(rs::Size::from_bits(size.bits().try_to_u8().unwrap())) else {
+                panic!("Can only create constant enum from bits. Got {:?}", val);
+            };
+
+            // special case for bits = 0 which can be indicative of a None constant.
+            // FIXME: Allow for other constant values by using the offset and value type.
+            //        However this is probably going to require machine information for
+            //        the value encoding which we don't have here.
+            if bits == 0 {
```
RalfJung, line 43: What is this doing?
This should be using the rustc interpreter APIs to get the discriminant and then recurse.
But actually what it should probably do is translate the raw bytes to a MiniRust allocation and then just load the value from there.

You don't have to worry about this for now, but I wonder why you had to write this code.

tooling/minimize/src/constant.rs
```diff
@@ -31,6 +31,42 @@ fn translate_const_val<'cx, 'tcx>(
...
+                let discriminant = match discriminator {
+                    Discriminator::Known(discriminant) => discriminant,
+                    Discriminator::Branch { fallback, children, .. } => {
+                        let child = children.into_iter()
+                                    .find_map(|((start, end), d)| if start <= Int::ZERO && Int::ZERO <= end { Some(d) } else { None })
+                                    .unwrap_or(fallback.extract());
```
RalfJung, line 49: Please don't do visual indentation, just intend by 4 spaces.

tooling/minimize/src/constant.rs
essickmango: Option<NonZeroU8>::None uses this for its constant somehow. I rewrote it to write the bytes into a global allocation, however we get the bits without endianness information (at least we know that "the first `size` bytes of `data` are the value)

tooling/minimize/tests/pass/iter.rs
```diff
@@ -0,0 +1,31 @@
...
+    }
+}
+
+fn main() {
+    let iter = RepeatN {
+        val: true,
+        repetitions: 3,
+    };
+    for i in iter {
+        print(i as u8);
```
RalfJung, line 29: Oh, I thought this was meant to print 0, 1, 2?

tooling/minimize/tests/pass/iter.rs
```diff
@@ -0,0 +1,31 @@
...
+    }
```
RalfJung, line 30:
```suggestion
    let mut iters = 0;
    for _ in iter {
        iters += 1;
    }
    assert(iters == 3);
```

tooling/minimize/src/constant.rs
essickmango: Thing is, we aren't given an allocation. We are given an `ConstValue::Scalar(Scalar::Int(ScalarInt))` where the Scalar comes from [here](https://doc.rust-lang.org/beta/nightly-rustc/rustc_middle/mir/interpret/value/enum.Scalar.html) and gives us the data as a `u128`.

The ConstValue documentation claims that Scalar and Slice could be used for enum discriminant optimizations and pattern matching simplification. However the type that this constant has is the entire enum, so I think we are literally byte-comparing or instantiating the enum from those bits. Now, I have read nowhere that those bits in the u128 have any endianness, which would be required for either getting a vector of bytes to move into a global allocation (because I don't know if the u128 is in any particular endianness and whether that matches the machine's), or to translate it into a MiniRust Value because for decoding I'd once again have to know the endianness.

With this said, the smallest thing to make it work is this way of moving it into a global allocation, though as I said, getting the bytes vector is a risk because of the endianness and I have yet to encounter this with a value other than 0u128.

The smallest thing to make this work on all 0u128 in a safe manner was the previous way, to ensure that the variant that has the tag 0 has no data and return this as `Value::Variant`. This is safe because if all bits are 0 then the tag is guaranteed to be 0, but does no longer work if we don't get a 0u128 constant (except for `u128::MAX`).

tooling/minimize/tests/pass/iter.rs
essickmango: Funnily enough I found a bug when I converted the `bool` to `u8` as the cast is a `IntToInt`-cast internally. So I'd like to keep it this way to have this as a test somewhere. Also the iterator is called `RepeatN`, of course it repeats it's value. We can test that it repeats the right amount of times by checking that the value appears N times in the output.

tooling/minimize/src/constant.rs
essickmango: Okay, after a few more hours of digging through compiler code it seems like [constant propagation](https://github.com/rust-lang/rust/blob/master/compiler/rustc_mir_transform/src/dataflow_const_prop.rs#L685) might indeed write immediates of the entire enum with the bytes in target endianness. Interestingly enough the documentation of ConstValue tells us that a constant enum represents "enum discriminants and pattern matching logic". Other than the pattern matching logic (which writes the enum bits to test into a global allocation to compare it to) I don't find anything that would write the discriminant.

tooling/minimize/src/constant.rs
```diff
@@ -31,6 +32,38 @@ fn translate_const_val<'cx, 'tcx>(
         Type::Tuple { fields, .. } if fields.is_empty() => {
             return ValueExpr::Tuple(List::new(), ty);
         }
+        Type::Enum { size, align, .. } => {
+            assert!(size.bytes() <= 16);
+            let Some(bits) = val.try_to_bits_for_ty(fcx.cx.tcx, rs::ParamEnv::reveal_all(), rs_ty) else {
+                panic!("Can only create constant enum from bits. Got {:?}", val);
+            };
+            // According to the documentation the first size bytes are data.
+            // This suggests BigEndian but is unclear.
```
RalfJung, line 41: This is in target endianess, i.e., `fcx.cx.tcx.data_layout().endian`. But we shouldn't have to care about that, ideally...

tooling/minimize/tests/pass/iter.rs
RalfJung: > Funnily enough I found a bug when I converted the bool to u8 as the cast is a IntToInt-cast internally. So I'd like to keep it this way to have this as a test somewhere. 

If you want a bool2int cast test, that should be a dedicated test (in tooling/minimize/tests/pass/casts.rs), not an accidental side-effect of another test. Otherwise the chances are too high that this will be refactored away at some point.

RalfJung: Repeating a bool here seems a bit odd as well, why not repeat a `u8`?

> We can test that it repeats the right amount of times by checking that the value appears N times in the output.

Relying on output for testing is not great. When a test fails it's not easy to figure out where things even went wrong. You have to figure out which part of the output corresponds to which lines of code.

It's much better to have the assertions directly in the test.

tooling/minimize/src/constant.rs
RalfJung: I've pushed a commit adjusting this slightly -- good enough for now. It's not the proper thing yet, but that needs a small rustc change so that we can access the right APIs: https://github.com/rust-lang/rust/pull/121396.

essickmango: Since this behaviour only occurs for `Option<NonZeroXX>` we decided to remove this from the test and open an issue for this. `consts.rs` needs some rework, see above.

tooling/minimize/tests/pass/enums.rs
```diff
@@ -0,0 +1,112 @@
...
+        Zero = 0,
+        One = 1,
+    }
+
+    assert(Aligned::Zero as u8 == 0);
+    assert(Aligned::One as u8 == 1);
+}
+
+#[allow(unused)]
+fn test_recursive_enum() {
```
RalfJung, line 36: This isn't actually recursive, just big.

So what happens currently when you enable this?

essickmango: see this: https://github.com/minirust/minirust/issues/161

RalfJung: Okay, makes sense.

But please rename the test to `test_big_enum` or so.

essickmango: okay

RalfJung:
Great, thanks. :)

PR #166, opened 2024-03-19
Migrate rvalue to StableMIR

celinval:
- Methods that handle StableMIR objects have the `_stable` suffix for now.

### DIFF
--- a/tooling/minimize/src/enums.rs
+++ b/tooling/minimize/src/enums.rs
@@ -177,6 +177,10 @@ impl<'tcx> Ctxt<'tcx> {
             .collect()
     }
 
+    pub fn discriminant_for_variant_stable(&self, ty: smir::Ty, variant_idx: smir::VariantIdx) -> Int {
+        self.discriminant_for_variant(smir::internal(self.tcx, ty), smir::internal(self.tcx, variant_idx))
+    }
+
     pub fn discriminant_for_variant(&self, ty: rs::Ty<'tcx>, variant_idx: rs::VariantIdx) -> Int {
         let rs::TyKind::Adt(adt_def, _) = ty.kind() else {
             panic!("Getting discriminant for a variant of a non-enum type!")
--- a/tooling/minimize/src/enums.rs
+++ b/tooling/minimize/src/enums.rs
@@ -177,8 +177,15 @@ impl<'tcx> Ctxt<'tcx> {
             .collect()
     }
 
-    pub fn discriminant_for_variant_stable(&self, ty: smir::Ty, variant_idx: smir::VariantIdx) -> Int {
-        self.discriminant_for_variant(smir::internal(self.tcx, ty), smir::internal(self.tcx, variant_idx))
+    pub fn discriminant_for_variant_smir(
+        &self,
+        ty: smir::Ty,
+        variant_idx: smir::VariantIdx,
+    ) -> Int {
+        self.discriminant_for_variant(
+            smir::internal(self.tcx, ty),
+            smir::internal(self.tcx, variant_idx),
+        )
     }
 
     pub fn discriminant_for_variant(&self, ty: rs::Ty<'tcx>, variant_idx: rs::VariantIdx) -> Int {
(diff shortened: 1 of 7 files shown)


tooling/minimize/src/ty.rs
```diff
@@ -11,6 +11,15 @@ impl<'tcx> Ctxt<'tcx> {
         Layout { size, align, inhabited }
     }
 
+    pub fn layout_of_stable(&self, ty: smir::Ty) -> Layout {
```
RalfJung, line 14: I think "smir" would be more clear as a suffix.

tooling/minimize/src/enums.rs
```diff
@@ -177,6 +177,10 @@ impl<'tcx> Ctxt<'tcx> {
             .collect()
     }
 
+    pub fn discriminant_for_variant_stable(&self, ty: smir::Ty, variant_idx: smir::VariantIdx) -> Int {
+        self.discriminant_for_variant(smir::internal(self.tcx, ty), smir::internal(self.tcx, variant_idx))
```
RalfJung, line 181: Ah, there's a way to go from an smir type back to its nternal type? That's good to know.

tooling/minimize/src/get.rs
```diff
@@ -31,7 +31,9 @@ impl<F: FnOnce(Program) + Send + Copy> Callbacks for Cb<F> {
         queries: &'tcx Queries<'tcx>,
     ) -> Compilation {
         queries.global_ctxt().unwrap().enter(|arg| {
-            let prog = Ctxt::new(arg).translate();
+            let prog = smir::run(arg, || {
+                Ctxt::new(arg).translate()
+            }).unwrap();
```
RalfJung, line 36: Could you add a comment explaining why `smir::run` is needed?

tooling/minimize/src/program.rs
```diff
@@ -126,6 +130,8 @@ pub struct FnCtxt<'cx, 'tcx> {
 
     pub locals: Map<LocalName, Type>,
     pub blocks: Map<BbName, BasicBlock>,
+
+    pub locals_stable: Vec<smir::LocalDecl>,
```
RalfJung, line 134: This seems to have a role quite different from `locals`, it's really part of the body. Could you move it next to the body, and add a doc comment explaining its purpose?

tooling/minimize/src/enums.rs
celinval: Yes, many constructs in SMIR can be translated back. The main exception today is MIR body components, like statements, operand, rvalue.

tooling/minimize/src/main.rs
```diff
@@ -52,12 +63,15 @@ mod rvalue;
...
 mod get;
+
 use get::get_mini;
 
 mod chunks;
+
 use chunks::calc_chunks;
 
 mod enums;
+
```
RalfJung, line 74: Why did you reformat this?

tooling/minimize/src/rvalue.rs
```diff
@@ -5,32 +5,39 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
-                rs::Rvalue::Use(operand) => self.translate_operand(operand),
-                rs::Rvalue::CheckedBinaryOp(bin_op, box (l, r))
-                | rs::Rvalue::BinaryOp(bin_op, box (l, r)) => {
-                    let lty = l.ty(&self.body, self.tcx);
-                    let rty = r.ty(&self.body, self.tcx);
+                smir::Rvalue::Use(operand) => self.translate_operand_smir(operand),
+                smir::Rvalue::CheckedBinaryOp(bin_op, l, r)
+                | smir::Rvalue::BinaryOp(bin_op, l, r) => {
+                    let lty = l.ty(&self.locals_smir).unwrap();
+                    let rty = r.ty(&self.locals_smir).unwrap();
```
RalfJung, line 23: When can these fail?

tooling/minimize/src/main.rs
celinval: I just ran `cargo fmt` or it's possible that my IDE messed it up. let me fix

tooling/minimize/src/rvalue.rs
celinval: This is overly protective code if the locals passed are not compatible with the projection.

tooling/minimize/src/main.rs
celinval: Yeah, it was my IDE. I've been playing around with RustRover, and it has some rough edges. I reverted it.

RalfJung:
Looks good, thanks!

Issue #167, opened 2024-03-25
enums in unions trigger "unsupported" error


RalfJung:
Testcase:
```rust
fn enum_in_union() {
    #[derive(Clone, Copy)]
    enum MyBool { MyTrue, MyFalse }

    union Bools { my: MyBool, theirs: bool }

    let _val = Bools { theirs: true };
    let _val = Bools { my: MyBool::MyFalse };
}
```

This is caused by `mark_used_bytes` not being able to handle enums.

essickmango:
I have an implementation ready [here](https://github.com/essickmango/minirust/tree/enum_mark_used_bytes), and will open a PR after I thought a bit about how to test for this or to leave it at that.

essickmango:
I agree with adding the discriminator, I didn't even think of it as it really should be equivalent tbh.

I thought I would add some more tests for nested structs and arrays as well, and it seems like structs cause the union to not be well-formed.

Regarding the failing test: The gap in `Option<(i8, i16)>` would be used for the discriminant so that would work fine. And as soon as I produce the gap with `repr(i16)` then I get well-formedness issues again. They seem to be with ADTs that have gaps in general?

essickmango:
Oh, found the issue: Every field needs to fit into a single chunk at the moment. This seems wrong considering that the types in the fields may have inside padding.

RalfJung:
The gap in option<(u8, u16)> cannot be used for the discriminant. Padding of inner fields can be overwritten when taking a reference to the field and assigning to it, so it cannot be used to store the discriminat.

RalfJung:
Note that `Option<(u8, u16)>` is not the same as `enum E { V1(u8, u16), V2 }`. The fact that `(u8, u16)` are wrapped in a tuple makes a difference.

> Oh, found the issue: Every field needs to fit into a single chunk at the moment. This seems wrong considering that the types in the fields may have inside padding.

I don't know what you mean by this. `mark_used_bytes` computes a bitmask, it doesn't care about "fields". The part of `calc_chunks` that processes that bitmask does not need any changes for enums.

essickmango:
> Note that Option<(u8, u16)> is not the same as enum E { V1(u8, u16), V2 }. The fact that (u8, u16) are wrapped in a tuple makes a difference.

True, I confused the two.

> I don't know what you mean by this. mark_used_bytes computes a bitmask, it doesn't care about "fields". The part of calc_chunks that processes that bitmask does not need any changes for enums.

[...]

PR #168, opened 2024-03-28
Mark used bytes of enums for unions

essickmango:
Implements `mark_used_bytes` for enums.
Also adds test cases to ensure that relevant data in structs, enums and arrays is marked, and that padding is not marked in the case of enums.

Fixes https://github.com/minirust/minirust/issues/167

### DIFF
--- a/tooling/minimize/tests/pass/union.rs
+++ b/tooling/minimize/tests/pass/union.rs
@@ -12,6 +12,46 @@ union B {
     f2: u8,
 }
 
+
+#[derive(Clone, Copy)]
+union EnumUnion {
+    data: Option<i32>,
+}
+
+fn extract_some_int(u: EnumUnion) -> i32 {
+    unsafe {
+        match u.data {
+            Some(i) => i,
+            None => unreachable!(),
+        }
+    }
+}
+
+
+#[derive(Clone, Copy, PartialEq, Eq)]
+struct TestStruct(i8, i16);
+
+#[derive(Clone, Copy)]
+union StructUnion {
+    data: TestStruct,
+}
+
+fn extract_struct(u: StructUnion) -> TestStruct {
+    unsafe { u.data }
+}
+
+
+#[derive(Clone, Copy)]
+union ArrayUnion {
+    data: [[i8; 3]; 2]
+}
+
+fn extract_array(u: ArrayUnion) -> [[i8; 3]; 2] {
+    unsafe { u.data }
+}
+
+
+
 fn main() {
     let mut x = A { f2: ()};
     x.f1 = 20;
@@ -20,4 +60,22 @@ fn main() {
     }
 
     let _y = B { f2: 0 };
+
+    // Make sure the value in `data` is preserved for enums, structs and arrays.
+    let u = EnumUnion { data: Some(42) };
+    assert!(extract_some_int(u) == 42);
+
+    let u = StructUnion { data: TestStruct(12, 1200) };
+    let s = extract_struct(u);
+    assert!(s.0 == 12);
+    assert!(s.1 == 1200);
+    
+    let u = ArrayUnion { data: [[42;3], [12;3]] };
+    let a = extract_array(u);
+    assert!(a[0][0] == 42);
+    assert!(a[0][1] == 42);
+    assert!(a[0][2] == 42);
+    assert!(a[1][0] == 12);
+    assert!(a[1][1] == 12);
+    assert!(a[1][2] == 12);
 }
--- a/tooling/minimize/tests/ub/enum_mark_used_bytes.rs
+++ b/tooling/minimize/tests/ub/enum_mark_used_bytes.rs
@@ -1,0 +1,33 @@
+include!("../helper/transmute.rs");
+
+/// This test case ensures that enums correctly mark their used bytes for unions and don't mark padding.
+/// It creates a union with all initialized bytes, and then after a copy tries to read all bytes.
+/// The read should fail as padding will get uninitialized during the copy.
+/// For alignment reasons the bytes are read as `u16`, as the enum is 2-byte aligned.
+
+#[derive(Clone, Copy)]
+#[repr(u16)]
+#[allow(unused)]
+enum Inner {
+    WithData(u16, u8),
+    Empty,
+}
+
+#[derive(Clone, Copy)]
+union TestUnion {
+    _data: Inner
+}
+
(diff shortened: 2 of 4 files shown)


tooling/minimize/tests/pass/union_marks_bytes.rs
```diff
@@ -0,0 +1,62 @@
...
+    unsafe {
+        match u.data {
+            Some(i) => i,
+            None => 0
+        }
+    }
+}
+
+
+// FIXME: structs cause union to not be well-formed
```
RalfJung, line 17: What's that about? That should be fixed by now I think.

tooling/minimize/tests/pass/union_marks_bytes.rs
RalfJung, line 1: We already have `union.rs`, I think you can just add `EnumUnion` there instead of making a new file?

tooling/minimize/tests/ub/enum_mark_used_bytes.rs
```diff
@@ -0,0 +1,35 @@
...
+    Empty,
+}
+
+#[derive(Clone, Copy)]
+union TestUnion {
+    _data: Inner
+}
+
+// Shorthand for an array of `u16` with the same size of the union.
+type UnionShorts = [u16; std::mem::size_of::<TestUnion>() / 2];
```
RalfJung, line 21: Why "shorts"...? Please don't use C integer type terminology, it is way too confusing.^^

tooling/minimize/tests/ub/enum_mark_used_bytes.rs
```diff
@@ -0,0 +1,35 @@
...
+fn get_union_shorts(u: TestUnion) -> UnionShorts {
+    unsafe { *(&u as *const TestUnion as *const UnionShorts) }
+}
+
+
+fn main() {
+    let u = unsafe {
+        let zero = [0u16; std::mem::size_of::<TestUnion>() / 2];
+        *(&zero as *const UnionShorts as *const TestUnion)
+    };
```
RalfJung, line 32: I think this is easier when expressed as a transmute. Then you also don't have to worry about alignment.
```rust
let zero = [0u8; std::mem::size_of::<TestUnion>()];
let u: TestUnion = transmute(zero);
// This is UB because the padding bytes get de-initialized.
let _ints: [0u8; std::mem::size_of::<TestUnion>()] = transmute(u);
```

essickmango: good point

tooling/minimize/tests/ub/enum_mark_used_bytes.rs
```diff
@@ -0,0 +1,35 @@
...
+    get_union_shorts(u);
+    let _shorts = get_union_shorts(u);
```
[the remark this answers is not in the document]
essickmango: Because I wanted to force the copy compared to the move. But the UB occurs already without.

tooling/minimize/tests/pass/union_marks_bytes.rs
essickmango: It is

tooling/minimize/tests/pass/union.rs
```diff
@@ -20,4 +60,25 @@ fn main() {
     }
 
     let _y = B { f2: 0 };
+
+    // We don't care if the read value is correct, only whether reading from it is UB.
+    // Calling the function twice forces a copy.
```
RalfJung, line 65: Move vs Copy makes no difference here, I think I'd prefer to remove that -- it just makes the test more confusing.

tooling/minimize/tests/ub/enum_mark_used_bytes.rs
```diff
@@ -0,0 +1,33 @@
...
+    WithData(u16, u8),
+    Empty,
+}
+
+#[derive(Clone, Copy)]
+union TestUnion {
+    _data: Inner
+}
+
+// Shorthand for an array of `u16` with the same size of the union.
```
RalfJung, line 21:
```suggestion
// Shorthand for an array of `u8` with the same size of the union.
```

tooling/minimize/tests/pass/union.rs
essickmango: Okay. I am not fully aware of the details of move vs copy semantics for copy objects, so I wanted to make sure that the union padding gets removed due to a literal copy,

tooling/minimize/tests/pass/union.rs
```diff
@@ -20,4 +60,23 @@ fn main() {
     }
 
     let _y = B { f2: 0 };
+
+    let u = EnumUnion { data: Some(42) };
+    assert!(test_enum_union(u) == test_enum_union(u));
```
RalfJung, line 65:
```suggestion
    // Make sure the value in `data` is preserved.
    let u = EnumUnion { data: Some(42) };
    assert!(test_enum_union(u) == 42);
```

tooling/minimize/tests/pass/union.rs
```diff
@@ -20,4 +60,23 @@ fn main() {
...
+
+    let u = StructUnion { data: TestStruct(12, 1200) };
+    let v1 = test_struct_union(u);
+    let v2 = test_struct_union(u);
```
RalfJung, line 69: You're still calling all these functions twice, what's the point of that? Instead please just call it once and check that the result matches the original input.

tooling/minimize/tests/pass/union.rs
```diff
@@ -20,4 +60,23 @@ fn main() {
...
+    assert!(v1.0 == v2.0);
+    assert!(v1.1 == v2.1);
+
+    let u = ArrayUnion { data: [[42;3], [12;3]] };
+    let v1 = test_array_union(u);
+    let v2 = test_array_union(u);
```
RalfJung, line 75: Same here.

tooling/minimize/tests/pass/union.rs
```diff
@@ -12,6 +12,46 @@ union B {
...
+#[derive(Clone, Copy)]
+union EnumUnion {
+    data: Option<i32>,
+}
+
+fn test_enum_union(u: EnumUnion) -> i32 {
+    unsafe {
+        match u.data {
+            Some(i) => i,
+            None => 0
```
RalfJung, line 25:
```suggestion
            None => unreachable!(),
```

RalfJung:
Thanks. :)

PR #169, opened 2024-04-02
minimize: remove Option wrap from `translate_stmt`

bifbof:
Removed the `Option` in `translate_stmt` by returning an empty vector instead.
I tried to remove the option from `translate_rvalue` as well, this would require support for the `BitAnd` operation.
Adding this support for `Int` is easy, but support for `Bool` is also required (possibly by casting to `Int` and doing the operation there, but I don't know if hacking a solution together is the right thing.)

Edit: changed `BinAdd` to `BitAnd` sorry for the confustion

### DIFF
--- a/tooling/minimize/src/bb.rs
+++ b/tooling/minimize/src/bb.rs
@@ -24,7 +24,7 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
                     stmts.push(Statement::Assign { destination, source });
                     stmts
                 } else {
-                    vec![] // assign of unsupported rvalues is IGNORED!
+                    vec![] // FIXME: assign of unsupported rvalues is IGNORED!
                 }
             }
             rs::StatementKind::StorageLive(local) => {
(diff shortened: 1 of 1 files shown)


RalfJung:
> I tried to remove the option from translate_rvalue as well, this would require support for the BinAdd operation.
Adding this support for Int is easy, but support for Bool is also required (possibly by casting to Int and doing the operation there, but I don't know if hacking a solution together is the right thing.)

So we already have code that generates BinAdd but we currently just do nothing there?

Wow okay that's bad. But this means that removing the `Option` is blocked on at least figuring out what the BinAdd is for, and maybe implementing it.

RalfJung:
> this would entail that BinAdd is supported.

You mean `BinAnd`, I assume?

bifbof:
> So we already have code that generates BinAdd but we currently just do nothing there?

We have code that turns `BitAnd` into `None` and then ignores it.

RalfJung:
No I mean we have tests that need BitAnd. But then we ignore it.

That's pretty bad. Good catch!

tooling/minimize/src/bb.rs
```diff
@@ -7,25 +7,25 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
-
-                // this puts the extra statements before the evaluation of `destination`!
-                stmts.push(Statement::Assign { destination, source });
-                stmts
+                if let Some((mut stmts, source)) = self.translate_rvalue(rval) {
+                    // this puts the extra statements before the evaluation of `destination`!
+                    stmts.push(Statement::Assign { destination, source });
+                    stmts
+                } else {
+                    vec![] // assign of unsupported rvalues is IGNORED!
```
RalfJung, line 27:
```suggestion
                    vec![] // FIXME: assign of unsupported rvalues is IGNORED!
```

PR #170, opened 2024-04-03
Add `BitAnd` to MiniRust

bifbof:
Add `BitAnd` to the binary ops of `Int` and `Bool`. The `translate_rvalue` now returns the new operation instead of None &rarr; therefore the return type does not have to be `Option` wrapped anymore. 
Should I add some additional tests for this?

### DIFF
--- a/spec/lang/syntax.md
+++ b/spec/lang/syntax.md
@@ -113,7 +113,7 @@ pub enum UnOpBool {
 pub enum UnOp {
     /// An operation on integers, with the given output type.
     Int(UnOpInt, IntType),
-    /// An operation on booleans.
+    /// An operation on booleans; returns a boolean.
     Bool(UnOpBool),
     /// Integer-to-pointer cast (uses previously exposed provenance).
     PtrFromExposed(PtrType),
@@ -134,6 +134,8 @@ pub enum BinOpInt {
     /// Remainder of a division, the `%` operator.
     /// Throws UB, if the modulus (second operand) is zero.
     Rem,
+    /// Bitwise-and two integer values.
+    BitAnd
 }
 
 /// A relation between integers.
@@ -152,6 +154,11 @@ pub enum IntRel {
     Ne,
 }
 
+pub enum BinOpBool {
+    /// Bitwise-and on booleans.
+    BitAnd,
+} 
+
 pub enum BinOp {
     /// An operation on integers, with the given output type.
     Int(BinOpInt, IntType),
@@ -159,6 +166,8 @@ pub enum BinOp {
     IntRel(IntRel),
     /// Pointer arithmetic (with or without inbounds requirement).
     PtrOffset { inbounds: bool },
+    /// An operation on booleans
+    Bool(BinOpBool),
 }
 ```
 
--- a/tooling/minimize/tests/pass/ops.rs
+++ b/tooling/minimize/tests/pass/ops.rs
@@ -10,6 +10,7 @@ fn main() {
     print(black_box(7) * 6);
     print(black_box(504) / 12);
     print(black_box(112) % 70);
+    print(black_box(171) & 62);
 
     print(black_box(10) > 2);
     print(black_box(10) >= 2);
@@ -17,4 +18,6 @@ fn main() {
     print(black_box(10) <= 2);
     print(black_box(10) == 2);
     print(black_box(10) != 2);
+
+    print(black_box(true) & true);
 }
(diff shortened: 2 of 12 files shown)


tooling/minimize/tests/pass/ops.rs
RalfJung, line 1: This test really should use `assert!`, not `print`... please make a note of that, to do it in a future PR.

RalfJung:
Apart from the missing test, looks good. :)

tooling/minimize/tests/pass/ops.rs
bifbof: Does this also apply to all other tests in `tooling/minimize/tests/pass/` with a `name.stdout` file?

bifbof:
> Yes please add a test in tooling/minitest.

Added a new test in `tooling/minitest/src/tests/bit_and.rs` :)

tooling/minimize/tests/pass/ops.rs
RalfJung: All the ones that use stdout just to test whether some value is computed correctly.

We should have at least one test ensuring that printing actually works. :)

tooling/minitest/src/tests/bit_and.rs
```diff
@@ -0,0 +1,96 @@
...
+/// Test that bit_and works for ints
+#[test]
+fn bit_and_bool_works() {
+    let locals = [<bool>::get_type()];
+    let b0 = block!(
+        storage_live(0),
+        assign(local(0), bit_and::<bool>(const_bool(false), const_bool(false),)),
+        assign(local(0), bit_and::<bool>(const_bool(false), const_bool(true),)),
+        assign(local(0), bit_and::<bool>(const_bool(true), const_bool(false),)),
+        assign(local(0), bit_and::<bool>(const_bool(true), const_bool(true),)),
```
RalfJung, line 32: This will not check that the correct value was actually computed, will it?

The new tests for bools should be in `bool.rs`, and for the ints we probably want an `int.rs` to put the tests there.

bifbof: No it doesn't. So far `minitest` `run_program` does not expose any computed values.

 I split the code into `bool.rs` and `int.rs`.

RalfJung: Yeah but you can add an `if` in the code that calls `unreachable` if anything goes wrong, and then the test would fail if we ever hit the `unreachable`.

There should be existing tests that demonstrate this technique.

RalfJung: See for instance here:

https://github.com/minirust/minirust/blob/f4098857543ac6f72d3c441a597dbd7e12736e75/tooling/minitest/src/tests/enum_discriminant.rs#L88-L101

bifbof: Ah yes that is quite neat.

spec/lang/syntax.md
```diff
@@ -152,13 +154,20 @@ pub enum IntRel {
...
+} 
+
 pub enum BinOp {
     /// An operation on integers, with the given output type.
     Int(BinOpInt, IntType),
     /// A relation between integers.
     IntRel(IntRel),
     /// Pointer arithmetic (with or without inbounds requirement).
     PtrOffset { inbounds: bool },
+    /// An operation on booleans
```
RalfJung, line 169:
```suggestion
    /// An operation on booleans.
```

RalfJung:
GitHub doesn't reliably send email notifications on force-pushes, so after you rebased and pushed please leave a message to ensure I see that.

bifbof:
@RalfJung branch is updated.

spec/lang/syntax.md
RalfJung: You ended up adding this comment at a different place than where I suggested it. :)

tooling/miniutil/src/build/expr.rs
```diff
@@ -122,6 +122,13 @@ pub fn mul<T: TypeConv>(l: ValueExpr, r: ValueExpr) -> ValueExpr {
 pub fn div<T: TypeConv>(l: ValueExpr, r: ValueExpr) -> ValueExpr {
     int_binop::<T>(BinOpInt::Div, l, r)
 }
+pub fn bit_and<T: TypeConv>(l: ValueExpr, r: ValueExpr) -> ValueExpr {
```
RalfJung, line 125: FYI: I ended up splitting this into 2 functions -- when something is 2 completely different operations in MiniRust, I don't think we want to hide that behind a single builder here.

PR #174, opened 2024-04-18
minimize: call `span_bug!`

bifbof:
Added `span_bug!` to all places where failing code seemed possible.
But it often points to some internal Rust code that generates that generates that error.


RalfJung:
Looks good, thanks!

> But it often points to some internal Rust code that generates that generates that error.

Internal code in the standard library, I assume? That's still a lot better than pointing to somewhere in the compiler. :)

PR #180, opened 2024-04-22
minimize: add `span_bug!` to `translate_ty`

bifbof:
This PR proposes two approaches handling `span` (and adding `span_bug!` to `translate_ty`)
1. **Simple approach (first commit)**: Pass `span` as an argument to each function as before.
2. **Experimental approach (second commit)**: Store `span` in `Ctxt` (change `TyCtxt` &rarr; `TyCtxtAt`)

The benefits of the experimental approach is that we get concise code and easy access to `span` whenever we need it, the disadvantage is that we have to remember to update `span`.

### DIFF
--- a/tooling/minimize/src/bb.rs
+++ b/tooling/minimize/src/bb.rs
@@ -50,7 +50,7 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
         let span = stmt.source_info.span;
         StatementResult::Statement(match &stmt.kind {
             rs::StatementKind::Assign(box (place, rval)) => {
-                let destination = self.translate_place(place);
+                let destination = self.translate_place(place, span);
                 // Some things that MIR handles as rvalues are non-deterministic,
                 // so MiniRust treats them differently.
                 match rval {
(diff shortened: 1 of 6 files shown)


RalfJung:
> The benefits of the experimental approach is that we get concise code and easy access to span whenever we need it, the disadvantage is that we have to remember to update span.

Yeah, the usual problems with global state...

I don't like storing this in the global context. That's just "too global". But I could be convinced to store the span in `FnCtxt`. How much code would that still save?

bifbof:
> Yeah, the usual problems with global state...
> 
> I don't like storing this in the global context. That's just "too global". But I could be convinced to store the span in `FnCtxt`. How much code would that still save?

We would save the code that is implemented in `FnCtxt` (basically all of PR #174). The code that was changed in the first commit of this PR would remain, as `translate_ty` is implemented in `Ctxt` thus we need to pass it in as argument.

bifbof:
I removed the second commit for the time being :)

RalfJung:
Right, so let's land the uncontroversial part first, makes sense.

RalfJung:
> We would save the code that is implemented in FnCtxt (basically all of PR https://github.com/minirust/minirust/pull/174). The code that was changed in the first commit of this PR would remain, as translate_ty is implemented in Ctxt thus we need to pass it in as argument.

Yes, that sounds right. I don't have a strong opinion either way here. But I do feel fairly strongly that we don't want to track the span of the current *function* in the *global* context -- that's a layering violation.

PR #181, opened 2024-04-26
minimize: workaround to get `--dump` to minimize

bifbof:


RalfJung:
Thanks!

With this, the flag is duplicated in two places, which can get out of sync in the future. Can you restructure this to avoid that? Basically, we need a function `parse_args` that returns `Vec` of arguments to pass to rustc, as well as the information about the flags for minimize itself.

bifbof:
Updated the PR @RalfJung, with a new function `split_args` that splits off all arguments that start with `--minimize-`.

RalfJung:
Looks reasonable, thanks!

Issue #158, opened 2024-02-22
Minimize: restructure `ConstValue` minimization


essickmango:
Currently the minimizer only supports a few const values, and it assumes they are all represented as scalars.
However what should actually happen is that it should recursively translate the const value using the API from https://github.com/rust-lang/rust/pull/121396, and therefore support arbitrary types.

RalfJung:
So to make this a bit more concrete, this is about `translate_const_val`. That should start by calling `let (ecx, val) = mk_eval_cx_for_const_val`, and then match on the type (which it already does) and then use methods like `ecx.read_scalar` and `ecx.project_field` to recurse over the `val`.

Also, we should not be calling `eval_to_allocation_raw`. That has `raw` in its name for a reason. Quoting from its doc comment:
> **Do not use this** directly, use the `eval_to_const_value` or `eval_to_valtree` instead.

RalfJung:
Fixed by https://github.com/minirust/minirust/pull/182

PR #182, opened 2024-05-02
minimize: restructe const evaluation

bifbof:
I changed the constant evaluation according to #158.
I wanted to find a cleaner way to deal with the pointer translation, but in the end kept the old system.


tooling/minimize/src/constant.rs
```diff
@@ -2,87 +2,82 @@ use crate::*;
 
 impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
     pub fn translate_const(&mut self, c: &rs::mir::Const<'tcx>, span: rs::Span) -> ValueExpr {
-        match c {
-            rs::mir::Const::Ty(_) => rs::span_bug!(span, "Const::Ty not supported"),
-            rs::mir::Const::Unevaluated(uneval, ty) =>
-                self.translate_const_uneval(uneval, *ty, span),
-            rs::mir::Const::Val(val, ty) => self.translate_const_val(val, *ty, span),
-        }
+        let val = c.eval(self.tcx, rs::ParamEnv::reveal_all(), None).expect("const-eval failed");
```
RalfJung, line 5: Would be good to use span_bug when const eval fails

tooling/minimize/src/constant.rs
```diff
@@ -2,87 +2,82 @@ use crate::*;
...
-        let constant = match ty {
+        let ty = self.translate_ty(val.layout.ty, span);
+        match ty {
             Type::Int(int_ty) => {
-                let val = val.try_to_scalar_int().unwrap();
-                let int: Int = match int_ty.signed {
-                    Signed => val.try_to_int(val.size()).unwrap().into(),
-                    Unsigned => val.try_to_uint(val.size()).unwrap().into(),
+                let scalar = ecx.read_scalar(&val).unwrap();
+                let val = match int_ty.signed {
```
RalfJung, line 26:
```suggestion
                let val: Int = match int_ty.signed {
```

tooling/minimize/src/constant.rs
```diff
@@ -2,87 +2,82 @@ use crate::*;
...
-                let (alloc_id, offset) =
-                    val.try_to_scalar().unwrap().to_pointer(&self.tcx).unwrap().into_parts();
-                let alloc_id = alloc_id.expect("no alloc id?").alloc_id();
+            Type::Ptr(ptr_ty) => {
+                if let PtrType::FnPtr(_) = ptr_ty {
+                    rs::span_bug!(span, "Function pointers are currently not supported")
+                }
+                let ptr = ecx.read_pointer(&val).unwrap();
+                let (prov, offset) = ptr.into_parts();
+                let alloc_id = prov.unwrap().alloc_id();
```
RalfJung, line 42: This can be `None` if the constant is something like
```rust
const C: *const i32 = 4 as *const i32;
```

tooling/minimize/src/constant.rs
```diff
@@ -136,12 +131,10 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
             .map(|&(offset, alloc_id)| {
                 // "Note that the bytes of a pointer represent the offset of the pointer.", see https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/interpret/struct.Allocation.html
                 // Hence we have to decode them.
-                let inner_offset_bytes: &[Option<u8>] = &bytes[offset.bytes() as usize..]
-                    [..DefaultTarget::PTR_SIZE.bytes().try_to_usize().unwrap()];
-                let inner_offset_bytes: List<u8> =
-                    inner_offset_bytes.iter().map(|x| x.unwrap()).collect();
-                let inner_offset: Int =
-                    DefaultTarget::ENDIANNESS.decode(Unsigned, inner_offset_bytes);
+                let start = offset.bytes() as usize;
```
RalfJung, line 134:
```suggestion
                let start = offset.bytes_usize();
```

tooling/minimize/src/constant.rs
```diff
@@ -136,12 +131,10 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
+                let end = start + DefaultTarget::PTR_SIZE.bytes().try_to_usize().unwrap();
+                let inner_offset = bytes[start..end].iter().map(|x| x.unwrap()).collect();
```
RalfJung, line 136:
```suggestion
                // Pointer bytes are always initialized, so we can unwrap.
                let inner_offset = bytes[start..end].iter().map(|x| x.unwrap()).collect();
```

RalfJung (commented):
Thanks, looks good overall! However, please add a test file that ensures we can correct translate a bunch of constants (at least one for each case in translate_const_val).

tooling/minimize/src/constant.rs
bifbof: Now create a raw pointer if provenance is `None`.

bifbof:
Thanks for the review! I have added the requested changes @RalfJung.
There were also some additional changes where I fixed a bug where I was using the `VariantIdx` instead of the real discriminator in the enums.

RalfJung:
> There were also some additional changes where I fixed a bug where I was using the VariantIdx instead of the real discriminator in the enums.

Is that covered by the tests you added?

bifbof:
> > There were also some additional changes where I fixed a bug where I was using the VariantIdx instead of the real discriminator in the enums.
> 
> Is that covered by the tests you added?

Yes, the  `assert!(!(() > ()));` and the assert with `NEG_ONE` and `NEG_TWO` test for it.

RalfJung:
Great, thanks!

PR #184, opened 2024-05-12
Add better errors to well-formedness check

### LINKED ISSUE #164: Better errors when well-formedness check fails
Currently WF-check just returns "no" when it fails, which is not very friendly. It should say what went wrong, and ideally also where.

bifbof:
Closes #164
- Changed `check_wf` return type to `Result<(), TerminationInfo>` such that we can append error information.
- Added a String to `TerminationInfo::IllFormed` for the error message.
- Adapted the `throw_ill_formed` macro such that it adds the position of the error.
- Made `ensure` a macro as well such that it interacts nicely with `throw_ill_formed`.

The error messages are quite brief in the moment, but I am also happy to change that to more descriptive ones.

### DIFF
--- a/spec/lang/well-formed.md
+++ b/spec/lang/well-formed.md
@@ -7,13 +7,13 @@ Those requirements are defined in this file.
 We also define the idea of a "value being well-formed at a type".
 `decode` will only ever return well-formed values, and `encode` will never panic on a well-formed value.
 
-Note that `check_wf` functions for testing well-formedness return `Option<()>` rather than `bool` so that we can use `?`.
+Note that `check_wf` functions for testing well-formedness return `Result<()>` to pass information in case an error occured.
+
 We use the following helper function to convert Boolean checks into this form.
 
 ```rust
-fn ensure(b: bool) -> Option<()> {
-    if !b { throw!(); }
-
+fn ensure_wf(b: bool, msg: &str) -> Result<()> {
+    if !b { throw_ill_formed!("{}", msg); }
     ret(())
 }
 ```
@@ -22,25 +22,21 @@ fn ensure(b: bool) -> Option<()> {
 
 ```rust
 impl IntType {
-    fn check_wf(self) -> Option<()> {
+    fn check_wf(self) -> Result<()> {
         // In particular, this checks that the size is at least one byte.
-        ensure(self.size.bytes().is_power_of_two())?;
-
-        ret(())
+        ensure_wf(self.size.bytes().is_power_of_two(), "IntType: size is not power of two")
     }
 }
 
 impl Layout {
-    fn check_wf<T: Target>(self) -> Option<()> {
+    fn check_wf<T: Target>(self) -> Result<()> {
         // We do *not* require that size is a multiple of align!
-        ensure(T::valid_size(self.size))?;
-
-        ret(())
+        ensure_wf(T::valid_size(self.size), "Layout: size not valid")
     }
 }
 
 impl PtrType {
-    fn check_wf<T: Target>(self) -> Option<()> {
+    fn check_wf<T: Target>(self) -> Result<()> {
         match self {
             PtrType::Ref { pointee, mutbl: _ } | PtrType::Box { pointee } => {
                 pointee.check_wf::<T>()?;
@@ -53,14 +49,14 @@ impl PtrType {
 }
 
 impl Type {
-    fn check_wf<T: Target>(self) -> Option<()> {
+    fn check_wf<T: Target>(self) -> Result<()> {
         use Type::*;
 
         // Ensure that the size is valid and a multiple of the alignment.
         let size = self.size::<T>();
-        ensure(T::valid_size(size))?;
+        ensure_wf(T::valid_size(size), "Type: size not valid")?;
         let align = self.align::<T>();
-        ensure(size.bytes() % align.bytes() == 0)?;
+        ensure_wf(size.bytes() % align.bytes() == 0, "Type: size is not multiple of alignment")?;
 
         match self {
             Int(int_type) => {
@@ -79,22 +75,25 @@ impl Type {
                     // Recursively check the field type.
                     ty.check_wf::<T>()?;
                     // And ensure it fits after the one we previously checked.
-                    ensure(offset >= last_end)?;
+                    ensure_wf(offset >= last_end, "Type::Tuple: overlapping fields")?;
                     last_end = offset + ty.size::<T>();
                 }
                 // And they must all fit into the size.
                 // The size is in turn checked to be valid for `M`, and hence all offsets are valid, too.
-                ensure(size >= last_end)?;
+                ensure_wf(size >= last_end, "Type::Tuple: size of fields is bigger than total size")?;
             }
             Array { elem, count } => {
-                ensure(count >= 0)?;
+                ensure_wf(count >= 0, "Type::Array: negative amount of elements")?;
                 elem.check_wf::<T>()?;
             }
             Union { fields, size, chunks, align: _ } => {
                 // The fields may overlap, but they must all fit the size.
                 for (offset, ty) in fields {
                     ty.check_wf::<T>()?;
-                    ensure(size >= offset + ty.size::<T>())?;
+                    ensure_wf(
+                        size >= offset + ty.size::<T>(),
+                        "Type::Union: field size does not fit union",
+                    )?;
                     // This field may overlap with gaps between the chunks. That's perfectly normal
                     // when there is padding inside the field.
                     // FIXME: should we check that all the non-padding bytes of the field are in some chunk?
@@ -106,11 +105,14 @@ impl Type {
                 // FIXME: should we relax this and allow arbitrary chunk order?
                 let mut last_end = Size::ZERO;
                 for (offset, size) in chunks {
-                    ensure(offset >= last_end)?;
+                    ensure_wf(
+                        offset >= last_end,
+                        "Type::Union: chunks are not stored in ascending order",
+                    )?;
                     last_end = offset + size;
                 }
                 // And they must all fit into the size.
-                ensure(size >= last_end)?;
+                ensure_wf(size >= last_end, "Type::Union: chunks do not fit union")?;
             }
             Enum { variants, size, discriminator, discriminant_ty, .. } => {
                 // All the variants need to be well-formed and be the size of the enum so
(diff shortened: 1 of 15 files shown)


spec/prelude/main.md
```diff
@@ -53,17 +53,35 @@ macro_rules! throw_memory_leak {
...
 macro_rules! throw_ill_formed {
-    () => {
-        do yeet TerminationInfo::IllFormed
+    ($($tt:tt)*) => {
+        do yeet TerminationInfo::IllFormed(format!(
+            "{} at {}:{}:{}",
+            format!($($tt)*),
+            file!(),
+            line!(),
+            column!()
```
RalfJung, line 64: I don't think I am a fan of this -- and it's not super useful either as it will point to the specr-transpile produced code, not the original sources.

Please make this behave just like `throw_ub`.

spec/prelude/main.md
```diff
@@ -53,17 +53,35 @@ macro_rules! throw_memory_leak {
...
     };
 }
+
 macro_rules! throw_deadlock {
     () => {
         do yeet TerminationInfo::Deadlock
     };
 }
 
+macro_rules! ensure {
```
RalfJung, line 75: The macro name is a bit too general to live globally. Maybe call it `ensure_wf`?

tooling/minitest/src/lib.rs
```diff
@@ -57,8 +57,11 @@ pub fn assert_ub_eventually(prog: Program, attempts: usize, msg: &str) {
 }
 
 #[track_caller]
-pub fn assert_ill_formed(prog: Program) {
-    assert_eq!(run_program(prog), TerminationInfo::IllFormed);
+pub fn assert_ill_formed(prog: Program, msg: &str) {
+    let TerminationInfo::IllFormed(info) = run_program(prog) else {
+        panic!("program is not ill formed!")
+    };
+    assert!(info.get_internal().starts_with(msg));
```
RalfJung, line 64: You can make this `==` after removing the file name and line number, I think?

spec/lang/well-formed.md
```diff
@@ -7,40 +7,27 @@ Those requirements are defined in this file.
...
 
 ```rust
 impl IntType {
-    fn check_wf(self) -> Option<()> {
+    fn check_wf(self) -> Result<()> {
         // In particular, this checks that the size is at least one byte.
-        ensure(self.size.bytes().is_power_of_two())?;
-
-        ret(())
+        ensure!(self.size.bytes().is_power_of_two(), "size of int is not power of two")
```
RalfJung, line 18:
```suggestion
        ensure!(self.size.bytes().is_power_of_two(), "IntType: size is not power of two")
```

spec/lang/well-formed.md
```diff
@@ -7,40 +7,27 @@ Those requirements are defined in this file.
...
 }
 
 impl Layout {
-    fn check_wf<T: Target>(self) -> Option<()> {
+    fn check_wf<T: Target>(self) -> Result<()> {
         // We do *not* require that size is a multiple of align!
-        ensure(T::valid_size(self.size))?;
-
-        ret(())
+        ensure!(T::valid_size(self.size), "size not valid")
```
RalfJung, line 25:
```suggestion
        ensure!(T::valid_size(self.size), "Layout: size not valid")
```

spec/lang/well-formed.md
```diff
@@ -53,14 +40,14 @@ impl PtrType {
...
 
 impl Type {
-    fn check_wf<T: Target>(self) -> Option<()> {
+    fn check_wf<T: Target>(self) -> Result<()> {
         use Type::*;
 
         // Ensure that the size is valid and a multiple of the alignment.
         let size = self.size::<T>();
-        ensure(T::valid_size(size))?;
+        ensure!(T::valid_size(size), "size not valid")?;
```
RalfJung, line 48:
```suggestion
        ensure!(T::valid_size(size), "Type: size not valid")?;
```

spec/lang/well-formed.md
```diff
@@ -53,14 +40,14 @@ impl PtrType {
...
         let align = self.align::<T>();
-        ensure(size.bytes() % align.bytes() == 0)?;
+        ensure!(size.bytes() % align.bytes() == 0, "size is not multiple of alignment")?;
```
RalfJung, line 50:
```suggestion
        ensure!(size.bytes() % align.bytes() == 0, "Type: size is not multiple of alignment")?;
```

spec/lang/well-formed.md
```diff
@@ -174,24 +162,24 @@ impl Discriminator {
...
             (Constant::FnPointer(fn_name), Type::Ptr(_)) => {
-                ensure(prog.functions.contains_key(fn_name))?;
+                ensure!(prog.functions.contains_key(fn_name), "invalid function name")?;
             }
             (Constant::PointerWithoutProvenance(addr), Type::Ptr(_)) => {
-                ensure(addr.in_bounds(Signedness::Unsigned, T::PTR_SIZE))?;
+                ensure!(addr.in_bounds(Signedness::Unsigned, T::PTR_SIZE), "pointer out-of-bounds")?;
             }
-            _ => throw!(),
+            _ => throw_ill_formed!("invalid constant type"),
```
RalfJung, line 182:
```suggestion
            _ => throw_ill_formed!("Constant: value does not match type"),
```

spec/lang/well-formed.md
```diff
@@ -213,49 +201,51 @@ impl ValueExpr {
 
                 match t {
                     Type::Tuple { fields, .. } => {
-                        ensure(exprs.len() == fields.len())?;
+                        ensure!(exprs.len() == fields.len(), "invalid number of fields")?;
```
RalfJung, line 204:
```suggestion
                        ensure!(exprs.len() == fields.len(), "ValueExpr::Tuple: invalid number of fields")?;
```

spec/lang/well-formed.md
```diff
@@ -274,23 +264,23 @@ impl ValueExpr {
...
                     Int(_int_op) => {
                         let Type::Int(int_ty) = operand else {
-                            throw!();
+                            throw_ill_formed!("invalid type");
                         };
                         Type::Int(int_ty)
                     }
                     Bool(_bool_op) => {
-                        ensure(matches!(operand, Type::Bool))?;
+                        ensure!(matches!(operand, Type::Bool), "invalid operand in boolean unop")?;
```
RalfJung, line 272:
```suggestion
                        ensure!(matches!(operand, Type::Bool), "UnOp::Bool: invalid operand")?;
```

spec/lang/well-formed.md
```diff
@@ -308,26 +298,26 @@ impl ValueExpr {
                 match operator {
                     Int(_int_op) => {
                         let Type::Int(int_ty) = left else {
-                            throw!();
+                            throw_ill_formed!("invalid type");
```
RalfJung, line 301:
```suggestion
                            throw_ill_formed!("BinOp::Int: invalid type");
```

spec/lang/well-formed.md
```diff
@@ -595,26 +620,33 @@ impl Function {
...
         // Ensure the start function exists, has the right ABI, takes no arguments, and returns a 1-ZST.
-        let func = self.functions.get(self.start)?;
-        ensure(func.calling_convention == CallingConvention::C);
-        let ret_layout = func.locals.get(func.ret)?.layout::<T>();
-        ensure(ret_layout.size == Size::ZERO && ret_layout.align == Align::ONE);
-        ensure(func.args.is_empty())?;
+        let Some(func) = self.functions.get(self.start) else {
+            throw_ill_formed!("invalid function name");
+        };
+        ensure!(func.calling_convention == CallingConvention::C, "invalid calling convention")?;
```
RalfJung, line 643:
```suggestion
        ensure!(func.calling_convention == CallingConvention::C, "Program: invalid calling convention")?;
```

RalfJung (commented):
Thanks! I left some comments. For the error message, I assume you get the pattern: they should always be of the form `conext: message`, where `context` is the name of the MiniRust construct that is ill-formed. I think that makes it easier to figure out what is wrong, and it also avoids having the exact same message in more than one place.

spec/prelude/main.md
bifbof: With the changes to `throw_ill_formed` this does not have to be a global macro anymore.
I made a function out of it (`ensure_wf`) and moved it to `well-formed.md`.

bifbof:
@RalfJung I updated the PR :)

tooling/minitest/src/lib.rs
```diff
@@ -57,8 +57,14 @@ pub fn assert_ub_eventually(prog: Program, attempts: usize, msg: &str) {
...
-pub fn assert_ill_formed(prog: Program) {
-    assert_eq!(run_program(prog), TerminationInfo::IllFormed);
+pub fn assert_ill_formed(prog: Program, msg: &str) {
+    let TerminationInfo::IllFormed(info) = run_program(prog) else {
+        panic!("program is not ill formed!")
+    };
+    if !(info.get_internal() == msg) {
+        panic!("{} vs {}", info.get_internal(), msg)
+    }
+    assert_eq!(info.get_internal(), msg);
```
RalfJung, line 67: What is this about, why are you comparing twice?
```suggestion
    assert_eq!(info.get_internal(), msg, "program is ill-formed with a different error message");
```
`assert_eq!` will already print the two values if there is a discrepancy.

spec/lang/well-formed.md
```diff
@@ -213,49 +228,55 @@ impl ValueExpr {
...
+                            ensure_wf(checked == ty, "ValueExpr::Tuple: invalid field type")?;
                         }
                     },
                     Type::Array { elem, count } => {
-                        ensure(exprs.len() == count)?;
+                        ensure_wf(exprs.len() == count, "ValueExpr::Tuple: invalid number of elements")?;
                         for e in exprs {
                             let checked = e.check_wf::<T>(locals, prog)?;
-                            ensure(checked == elem)?;
+                            ensure_wf(checked == elem, "ValueExpr::Tuple::Array: invalid element type")?;
```
RalfJung, line 241:
```suggestion
                            ensure_wf(checked == elem, "ValueExpr::Array: invalid element type")?;
```
`Tuple` seems wrong?

spec/lang/well-formed.md
```diff
@@ -274,23 +295,23 @@ impl ValueExpr {
                 match operator {
                     Int(_int_op) => {
                         let Type::Int(int_ty) = operand else {
-                            throw!();
+                            throw_ill_formed!("ValueExpr::UnOp::Int: invalid operand");
```
RalfJung, line 298:
```suggestion
                            throw_ill_formed!("UnOp::Int: invalid operand");
```
`UnOp` already implies it's a value expression, let's not be unnecessarily verbose. (Applies everywhere. I specifically an example with UnOp in my previous review to avoid this. ;)

spec/lang/well-formed.md
```diff
@@ -308,26 +329,26 @@ impl ValueExpr {
                 match operator {
                     Int(_int_op) => {
                         let Type::Int(int_ty) = left else {
-                            throw!();
+                            throw_ill_formed!("ValueExpr::BinOp::Int: invalid left type");
```
RalfJung, line 332:
```suggestion
                            throw_ill_formed!("BinOp::Int: invalid left type");
```

spec/lang/well-formed.md
```diff
@@ -523,7 +567,7 @@ impl Terminator {
                 match intrinsic {
                     IntrinsicOp::AtomicFetchAndOp(op) => {
                         if !is_atomic_binop(op) {
-                            throw!();
+                            throw_ill_formed!("Terminator::Intrinsic::IntrinsicOp::AtomicFetchAndOp: non atomic op");
```
RalfJung, line 570:
```suggestion
                            throw_ill_formed!("IntrinsicOp::AtomicFetchAndOp: non atomic op");
```

RalfJung (commented):
Just some nits regarding the error messages. :)

tooling/minitest/src/lib.rs
RalfJung: You didn't quite apply the patch I suggested. Having a custom message is still good courtesy. :)

bifbof: ups, that was something I forgot to remove.

spec/lang/well-formed.md
bifbof: Removed the verbosity everywhere. Now the maximum depth is `a::b:` and it works out fine. 👍🏼

spec/lang/well-formed.md
RalfJung: Ah, good point. What about "ValueExpr::Tuple: invalid array element type"?
And then below, `"ValueExpr::Tuple: invalid number of array elements"?

bifbof: That works fine I think. :)

PR #185, opened 2024-05-14
add tests for NonZeroI32

RalfJung:
This used to be broken due do gaps in our const translation. That doesn't seem to happen any more (a compiler update a while ago made the problematic const go away, and also we can now handle such consts).


RalfJung:
Ah, we need to normalize after getting the field on an ADT.

Issue #186, opened 2024-05-14
Add missing operators


RalfJung:
Rust has a bunch of operators that we do not support yet:
- [x] BitOr, BitXor (on integers and Booleans)
- [x] Shl, Shr
- [x] Cmp
- [x] checked operators (in MIR encoded via a separate `Rvalue::CheckedBinaryOp`) -- AFAIK this is only allowed for Add, Sub, Mul; they should become separate BinOp for us
- [x] some intrinsics that are lowered to binops: AddUnchecked, SubUnchecked, MulUnchecked, ShlUnchecked, ShrUnchecked
- [x] `UnOp::Not`, which we support for Booleans but not for integers
- [x] Comparison operators on `bool` (here it seems best to not add anything new to MiniRust, and teach minimize to translate this to operations on integers)
- [x] Comparison operators on pointers (probably also best done via casts)

bifbof:
The boolean comparison operators are also not supported yet:
```rust
fn main() {
    let _ = false < true;
    let _ = false <= true;
    let _ = false == true;
    let _ = false >= true;
    let _ = false > true;
}
```

RalfJung:
We could handle those in minimize, by just casting to int and comparing there.

bifbof:
On a similar note the comparison for pointers is also not supported:
```rust
fn main() {
    let a = &42 as *const i32;
    let b = &21 as *const i32;
    let _ = a == b;
}
```

RalfJung:
That too should IMO be done via casts to integers; we'll have to add new casts for that.

RalfJung:
> That too should IMO be done via casts to integers; we'll have to add new casts for that.

In fact we do not, we can just transmute to integers. Under our current provenance model, that has exactly the right behavior.

PR #187, opened 2024-05-15
Add operations: `BitXor`, `BitOr`, and `Not` to MiniRust

bifbof:
Part of https://github.com/minirust/minirust/issues/186
Should be pretty straight for the most part.

~~But for the shift operation I am unsure if i created the correct semantics.~~
~~The [rustc doc](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/enum.BinOp.html#variant.Shl) states:~~
```
Shl: The << operator (shift left)

The offset is truncated to the size of the first operand and made unsigned before shifting.
```
~~For made unsigned I see two possibilities:~~
~~1. C-style: interpret the bits as unsigned integers now (e.g. `-1i8` &rarr; `255u8`)~~
~~2. Throw Error (but not UB?) if trying to shift with negative value.~~
~~So far the 2nd version is implemented by `libspecr` (via BigInt).~~
~~


RalfJung:
Thanks!
Please restrict this to BitXor, BitOr, Not so that we can discuss the shifts separately.

spec/lang/syntax.md
```diff
@@ -100,6 +100,8 @@ pub enum Constant {
 pub enum UnOpInt {
     /// Negate an integer value arithmetically (`x` becomes `-x`).
     Neg,
+    /// Bitwise-not integer value
```
RalfJung, line 103:
```suggestion
    /// Bitwise-negate an integer value
```

tooling/minimize/src/rvalue.rs
```diff
@@ -20,60 +20,54 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
-                            operand: GcCow::new(operand),
-                        }
-                    }
-                },
+                use smir::UnOp::*;
+                let operator = match (unop, ty) {
+                    (Neg, Type::Int(_)) => UnOp::Int(UnOpInt::Neg),
+                    (Not, Type::Int(_)) => UnOp::Int(UnOpInt::Not),
+                    (Not, Type::Bool) => UnOp::Bool(UnOpBool::Not),
+                    (op, _) => rs::span_bug!(span, "UnOp {op:?} called with unsupported type."),
```
RalfJung, line 67:
```suggestion
                    (op, ty) => rs::span_bug!(span, "UnOp {op:?} called with unsupported type {ty}"),
```

tooling/minimize/src/rvalue.rs
```diff
@@ -20,60 +20,54 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
-                            ty => rs::span_bug!(span, "BitAnd not supported for: {ty:?}"),
-                        },
-                    x => rs::span_bug!(span, "Binary Op not supported: {x:?}"),
+                    // all bool ops
+                    (BitAnd, Type::Bool) => BinOp::Bool(BinOpBool::BitAnd),
+                    (BitOr, Type::Bool) => BinOp::Bool(BinOpBool::BitOr),
+                    (BitXor, Type::Bool) => BinOp::Bool(BinOpBool::BitXor),
+
+                    (op, ty) =>
+                        rs::span_bug!(span, "Binary Op {op:?} not supported for type {ty:?}"),
```
RalfJung, line 52:
```suggestion
                        rs::span_bug!(span, "Binary Op {op:?} not supported for type {ty}"),
```
The debug rendering of types can be quite verbose

tooling/minimize/tests/pass/ops.rs
```diff
@@ -5,12 +5,17 @@ fn black_box<T>(t: T) -> T { t }
 
 fn main() {
     print(-black_box(-42));
+    print(!black_box(-43));
```
RalfJung, line 8: Please don't add more `print` tests -- it's so hard to figure out which line of output corresponds to which part of the test.

Instead, use `assert!`.

RalfJung, line 8: Please also check an unsigned number, like `!1u32` becoming `(u32::MAX-1)`

tooling/minimize/src/rvalue.rs
bifbof: Sadly I cannot do this as types do not implement `std::fmt::Display`. So for now I dropped printing the types at all.

tooling/minimize/tests/pass/ops.rs
bifbof: Added `assert!(!black_box(i32::MIN) == i32::MAX);` as well for good measure :)

bifbof:
I removed the shift operations.

tooling/minimize/src/rvalue.rs
RalfJung: Rust types implement `Display`... can we print the Rust type instead?

RalfJung:
> I removed the shift operations.

Thanks! Please also remove them from the PR title and commit message.

RalfJung:
> [BitXor, BitOr, and Ǹot to MiniRust](https://github.com/minirust/minirust/pull/187/commits/5f447cdf289914f02ba7593613261d9a93cc4953)

Your commit message has a strange `Ǹ` character in it still. ;)

bifbof:
> > [BitXor, BitOr, and Ǹot to MiniRust](https://github.com/minirust/minirust/pull/187/commits/5f447cdf289914f02ba7593613261d9a93cc4953)
> 
> Your commit message has a strange `Ǹ` character in it still. ;)

oh man, thanks for the reminder.

RalfJung:
Looks good, thanks!

PR #188, opened 2024-05-21
bump ui_test to 0.23

oli-obk:


tooling/minimize/tests/ui.rs
```diff
@@ -1,20 +1,30 @@
...
+    };
+
+    config.comment_defaults.base().mode = Spanned::dummy(mode).into();
+    config.comment_defaults.base().set_custom(
+        "dependencies",
+        DependencyBuilder {
+            crate_manifest_path: "./tests/deps/Cargo.toml".into(),
+            ..Default::default()
+        },
+    );
```
RalfJung, line 26: I am a bit surprised to see the dependency builder set via `comment_defaults`... and then `set_custom`? It says that's for custom flags but we don't have a custom flag, do we? Is `dependencies`  a magic flag name that is recognized by ui_test?

RalfJung (commented):
Thanks! That's a lot less tricky than I expected (given how there seems to be trouble updating the version for Miri).

FWIW, I'd prefer to avoid calling `run_tests_generic` and stick to the high-level APIs but it seems not possible to run multiple `Config` in parallel without that.

tooling/minimize/tests/ui.rs
oli-obk: the name doesn't matter, you can name it anything you want. The name only matters for flags that include their own comment parsers that may need to modify an existing item.

RalfJung: Now I'm even more confused. So we add a custom flag that's not a flag and we give it a name that's not used for parsing and then somehow that adds a dependency builder...?

oli-obk: maybe I should name them "settings". these are the same things as the explicitly named fields on `Revisioned`. E.g. `edition` is not a field anymore, but just the `"edition"` map key. Ideally we wouldn't have a key at all, and just use a set instead of a map, but that makes lookup for overwriting harder.

RalfJung: Ah so I guess ideally this would be one of these type-indexed `Any` maps where one can insert and looking things using their type as the key?

oli-obk: yes, I may give this another attempt in the future, but for now, a stringly keyed map is just simpler

tooling/minimize/tests/ui.rs
```diff
@@ -1,20 +1,30 @@
...
+    let mut config = Config {
         program,
-        dependencies_crate_manifest_path: Some(PathBuf::from("./tests/deps/Cargo.toml")),
         out_dir: PathBuf::from(env!("CARGO_TARGET_TMPDIR")).join("ui"),
         ..Config::rustc(path)
-    }
+    };
+
+    config.comment_defaults.base().mode = Spanned::dummy(mode).into();
+    config.comment_defaults.base().set_custom(
```
RalfJung, line 20:
```suggestion
    // To let tests use dependencies, we have to add a `DependencyBuilder`
    // custom "comment" (with arbitrary name), which will then take care
    // of building the dependencies and making them available in the test.
    config.comment_defaults.base().set_custom(
```

tooling/minimize/tests/ui.rs
RalfJung: Okay I see, thanks for explaining!

RalfJung:
I have reverted this PR in https://github.com/minirust/minirust/pull/194; https://github.com/oli-obk/ui_test/issues/238 made debugging test failures near impossible.

PR #189, opened 2024-05-27
miniutil: new test api

### LINKED ISSUE #160: Minitest: make writing tests more ergonomic
Currently, the tests in `minitest` do all management of local names / basic block IDs / function names "by hand". That's not very ergonomic. We should come up with APIs (somewhat similar to rustc codegen) that let us write tests more ergonomically.

I am imagining something roughly like this:
```rust
let mut p = ProgramBuilder::new();

let mut f = p.function();
let var = f.local::<u32>();
f.storage_live(var);
f.atomic_store(addr_of_constptr(var), const_int(1u32));
f.exit();
p.finish_function(f);
```

For straight-line code (including function pointers), we don't have to talk about basic blocks at all. For if-then-else we need something slightly lower-level, details TBD.

bifbof:
closes #160 
This PR adds a new miniutil test API. The implementation follows the interface proposed in #160 relatively closely. (The only thing I didn't add from issue #160 was `addr_of_constptr(var)`.)

We add two structures:
- `ProgramBuilder`, which takes care of the program, functions and global variables
- `FunctionBuilder`, which takes care of args, ret, locals, blocks, statements and terminators

New test file `builder_api.rs` tests all new public methods that have been added.
I also rewrote some tests in `locks.rs` and `packed.rs` to show what the interface looks like in the wild.

### DIFF
--- a/tooling/minitest/src/tests/builder_api.rs
+++ b/tooling/minitest/src/tests/builder_api.rs
@@ -1,0 +1,172 @@
+use std::u32;
+
+use crate::*;
+
+#[test]
+fn global_var() {
+    let mut p = ProgramBuilder::new();
+    let var = p.declare_global_zero_initialized::<u32>();
+
+    let mut f = p.declare_function();
+    f.assign(var, const_int(42u32));
+    f.if_(eq(load(var), const_int(42u32)), |f| f.exit(), |f| f.unreachable());
+    let f = p.finish_function(f);
+
+    let p = p.finish_program(f);
+    assert_stop(p);
+}
+
+#[test]
+fn local_var() {
+    let mut p = ProgramBuilder::new();
+
+    let mut f = p.declare_function();
+    let var = f.declare_local::<u32>();
+    f.storage_live(var);
+    f.assign(var, const_int(42u32));
+    f.if_(eq(load(var), const_int(42u32)), |_| {}, |f| f.unreachable());
+    f.exit();
+    let f = p.finish_function(f);
+
+    let p = p.finish_program(f);
+    assert_stop(p);
+}
+
+#[test]
+fn arg_and_ret_var() {
+    let mut p = ProgramBuilder::new();
+
+    let add_two_if_42: FnName = {
+        let mut f = p.declare_function();
+        let var = f.declare_arg::<u32>();
+        let ret = f.declare_ret::<u32>();
+        f.if_(eq(load(var), const_int(42u32)), |_| {}, |f| f.unreachable());
+        f.assign(ret, add(load(var), const_int(2u32)));
+        f.return_();
+        p.finish_function(f)
+    };
+
+    let start: FnName = {
+        let mut start = p.declare_function();
+        let ret_place = start.declare_local::<u32>();
+        start.storage_live(ret_place);
+        start.call(ret_place, add_two_if_42, &[by_value(const_int(42u32))]);
+        start.if_(eq(load(ret_place), const_int(44u32)), |f| f.exit(), |f| f.unreachable());
+        p.finish_function(start)
+    };
+
+    let p = p.finish_program(start);
+    assert_stop(p);
+}
+
+#[test]
+fn switch_int() {
+    let mut p = ProgramBuilder::new();
+
+    let mut f = p.declare_function();
+    let var = f.declare_local::<u32>();
+    f.storage_live(var);
+    f.assign(var, const_int(42_u32));
+    f.switch_int(
+        load(var),
+        &[
+            (41, &|f| f.assign(var, add(load(var), const_int(0u32)))),
+            (42, &|f| f.assign(var, add(load(var), const_int(1u32)))),
+            (43, &|f| f.assign(var, add(load(var), const_int(0u32)))),
+        ],
+        |f| f.assign(var, add(load(var), const_int(0u32))),
+    );
+    f.if_(eq(load(var), const_int(43_u32)), |_| {}, |f| f.unreachable());
+    f.storage_dead(var);
+    f.exit();
+    let f = p.finish_function(f);
+
+    let p = p.finish_program(f);
+    assert_stop(p);
+}
+
+#[test]
+fn while_() {
+    let mut p = ProgramBuilder::new();
+    let mut f = p.declare_function();
+    let var = f.declare_local::<u32>();
+    let counter = f.declare_local::<u32>();
+    f.storage_live(var);
+    f.storage_live(counter);
+    f.assign(var, const_int(0u32));
+    f.assign(counter, const_int(0u32));
+    f.while_(lt(load(counter), const_int(42u32)), |f| {
+        f.assign(counter, add(load(counter), const_int(1u32)));
+        f.assign(var, add(load(var), const_int(2u32)));
+    });
+    f.if_(eq(load(var), const_int(84u32)), |f| f.exit(), |f| f.unreachable());
+    let f = p.finish_function(f);
+    let p = p.finish_program(f);
+    assert_stop(p);
+}
+
+#[test]
+#[should_panic(expected = "PlaceExpr is not a local")]
+fn storage_live_with_non_local() {
+    let mut p = ProgramBuilder::new();
+    let g = p.declare_global_zero_initialized::<i32>();
+    let mut f = p.declare_function();
+    f.storage_live(g);
+}
+
+#[test]
+#[should_panic(expected = "PlaceExpr is not a local")]
+fn storage_dead_with_non_local() {
+    let mut p = ProgramBuilder::new();
+    let g = p.declare_global_zero_initialized::<i32>();
+
+    let mut f = p.declare_function();
+    f.storage_dead(g);
+}
+
+#[test]
+#[should_panic(expected = "finish_block: there is no block to finish")]
+fn double_exit() {
+    let mut p = ProgramBuilder::new();
+
+    let mut f = p.declare_function();
+    f.exit();
+    f.exit(); // this is wrong, we already finished the block
+    let f = p.finish_function(f);
+
+    let p = p.finish_program(f);
+    assert_stop(p);
+}
+
+#[test]
+#[should_panic(expected = "There is no current block. Cannot insert statement/terminator.")]
+fn statement_after_exit() {
+    let mut p = ProgramBuilder::new();
+
+    let mut f = p.declare_function();
+    let var = f.declare_local::<u32>();
+    f.exit();
+    f.assign(var, const_int(42_u32)); // this is wrong, we already finished the block
+    f.exit();
+    let f = p.finish_function(f);
+
+    let p = p.finish_program(f);
+    assert_stop(p);
+}
+
+#[test]
+#[should_panic(
+    expected = "Function has an unfinished block. You need to return or exit from the last block."
+)]
+fn no_exit() {
+    let mut p = ProgramBuilder::new();
+
+    let mut f = p.declare_function();
+    let var = f.declare_local::<u32>();
+    f.storage_live(var);
+    f.assign(var, const_int(42_u32));
+    let f = p.finish_function(f);
+
+    let p = p.finish_program(f);
+    assert_stop(p);
+}
--- a/tooling/minitest/src/tests/locks.rs
+++ b/tooling/minitest/src/tests/locks.rs
@@ -12,37 +12,50 @@ use crate::*;
 /// By making the critical section large (256 times around a loop)
 /// we get a high probability that the handover happened.
 fn lock_handover() {
-    let locals = [<()>::get_type(), <u32>::get_type()];
+    let mut p = ProgramBuilder::new();
+    let lock = p.declare_global_zero_initialized::<u32>();
 
-    let b0 = block!(
-        storage_live(1),
-        assign(local(1), const_int::<u32>(0)),
-        lock_acquire(load(global::<u32>(0)), 1)
-    );
-    let b1 = block!(if_(eq(load(local(1)), const_int::<u32>(256)), 3, 2));
-    let b2 = block!(assign(local(1), add(load(local(1)), const_int::<u32>(1))), goto(1));
-    let b3 = block!(lock_release(load(global::<u32>(0)), 4));
-    let b4 = block!(return_());
-    let critical = function(Ret::Yes, 0, &locals, &[b0, b1, b2, b3, b4]);
+    let critical: FnName = {
+        let mut critical = p.declare_function();
+        let val = critical.declare_local::<u32>();
 
-    let locals = [<u32>::get_type(), <()>::get_type()];
+        critical.storage_live(val);
+        critical.assign(val, const_int(0u32));
+        critical.lock_acquire(load(lock));
+        critical.while_(ne(load(val), const_int(256_u32)), |f| {
+            f.assign(val, add(load(val), const_int(1u32)));
+        });
+        critical.lock_release(load(lock));
+        critical.return_();
 
-    let b0 = block!(storage_live(0), storage_live(1), lock_create(global::<u32>(0), 1),);
-    let b1 = block!(spawn(fn_ptr(1), null(), local(0), 2));
-    let b2 = block!(call(2, &[], local(1), Some(3)));
-    let b3 = block!(join(load(local(0)), 4));
-    let b4 = block!(exit());
-    let main = function(Ret::No, 0, &locals, &[b0, b1, b2, b3, b4]);
+        p.finish_function(critical)
+    };
 
-    let locals = [<()>::get_type(), <*const ()>::get_type()];
+    let mut second = p.declare_function();
 
-    let b0 = block!(call(2, &[], local(0), Some(1)));
-    let b1 = block!(return_());
-    let second = function(Ret::Yes, 1, &locals, &[b0, b1]);
+    let main: FnName = {
+        let mut main = p.declare_function();
+        let thread_id = main.declare_local::<u32>();
 
-    let globals = [global_int::<u32>()];
+        main.storage_live(thread_id);
+        main.lock_create(lock);
+        main.spawn(second.name(), null(), thread_id);
+        main.call_ignoreret(critical, &[]);
+        main.join(load(thread_id));
+        main.exit();
 
-    let p = program_with_globals(&[main, second, critical], &globals);
+        p.finish_function(main)
+    };
+
+    // implement function `second`
+    {
+        second.declare_arg::<*const ()>();
+        second.call_ignoreret(critical, &[]);
+        second.return_();
+        p.finish_function(second);
+    }
+
+    let p = p.finish_program(main);
     assert_stop(p);
 }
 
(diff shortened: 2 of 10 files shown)


tooling/minitest/src/tests/builder_api.rs
```diff
@@ -0,0 +1,115 @@
...
+    p.finish_function(f);
+    let p = p.finish_program();
+    assert_stop(p);
+}
+
+#[test]
+fn local_var() {
+    let mut p = ProgramBuilder::new();
+    let mut f = p.function();
+    let var = f.local::<u32>();
```
RalfJung, line 20:
```suggestion
    let mut f = p.declare_function();
    let var = f.declare_local::<u32>();
```
I think it'd be good to have some pattern for all the functions that declare a fresh name. `f.ret()` could also mean "return from the function", after all.

tooling/minitest/src/tests/builder_api.rs
```diff
@@ -0,0 +1,115 @@
...
+    f.assign(ret, add(load(var), const_int(2u32)));
+    f.exit();
+    let add_two_if_42 = p.finish_function(f);
+
+    let mut start = p.function();
+
+    let l = start.local::<u32>();
+    start.storage_live(l);
+    start.call(add_two_if_42, &[by_value(const_int(42u32))], l, Ret::Yes);
+    start.if_(eq(load(l), const_int(44u32)), |f| f.exit(), |f| f.unreachable());
```
RalfJung, line 46:
```suggestion
    start.if_(eq(load(l), const_int(44u32)), |f| f.return_(), |f| f.unreachable());
```

tooling/minitest/src/tests/builder_api.rs
```diff
@@ -0,0 +1,115 @@
...
+    let start = p.finish_function(start);
+
+    p.set_start(start);
+    let p = p.finish_program();
+    assert_stop(p);
+}
+#[test]
```
RalfJung, line 53:
```suggestion

#[test]
```

tooling/minitest/src/tests/builder_api.rs
```diff
@@ -0,0 +1,115 @@
...
+        &[
+            (41, &|f| f.assign(var, add(load(var), const_int(0u32)))),
+            (42, &|f| f.assign(var, add(load(var), const_int(1u32)))),
+            (43, &|f| f.assign(var, add(load(var), const_int(0u32)))),
+        ],
+        |f| f.assign(var, add(load(var), const_int(0u32))),
+    );
+    f.if_(eq(load(var), const_int(43_u32)), |_| {}, |f| f.unreachable());
+    f.storage_dead(var);
+    f.exit();
```
RalfJung, line 90: What happens if you don't exit here?

tooling/miniutil/src/build/mod.rs
```diff
@@ -39,6 +39,136 @@ pub use ty::*;
...
+    }
+
+    pub fn local<T: TypeConv>(&mut self) -> PlaceExpr {
+        let name = LocalName(Name::from_internal(self.next_local));
+        self.next_local += 1;
+        self.locals.try_insert(name, T::get_type()).unwrap();
+        local_by_name(name)
+    }
+
+    pub fn local_non_const_ty(&mut self, t: Type) -> PlaceExpr {
```
RalfJung, line 150: The name is confusing. This has nothing to do with `const`, does it?

Maybe `local_with_ty`?

tooling/miniutil/src/build/terminator.rs
```diff
@@ -1,5 +1,243 @@
...
+    pub fn return_(&mut self) {
+        self.terminator = Some(Terminator::Return);
+        self.finish_block();
+    }
+
+    // terminators with 0 or 1 following blocks
+    pub fn call(&mut self, f: FnName, args: &[ArgumentExpr], ret: PlaceExpr, can_return: Ret) {
+        let next_block = match can_return {
+            Ret::No => None,
+            Ret::Yes => Some(BbName(Name::from_internal(self.next_block))),
```
RalfJung, line 49: Instead of directly fiddling with `next_block`, what about having a `declare_block` function that just returns a fresh block name, and never accessing `self.next_block` outside that function? Then it is less likely that you will accidentally use a block name twice.

Below then instead of `prepare_next_block()` it would be something like `self.set_current_block(next_block)` where `next_block` is what got returned by `declare_block`.

tooling/miniutil/src/build/terminator.rs
```diff
@@ -1,5 +1,243 @@
...
+    fn prepare_next_block(&mut self) {
+        self.new_block(self.next_block);
+        self.next_block += 1;
+    }
+
+    // terminators with 0 following blocks
+    pub fn exit(&mut self) {
+        self.terminator = Some(exit());
+        self.finish_block();
+    }
```
RalfJung, line 28: It would be good if something like this led to an error:
```rust
f.exit();
f.exit();
```
So, `finish_block` should leave the builder in a state where there is *no* current block, and then any attempt at adding a statement or terminator should fail.

In `FunctionBuilder`, this could look like
```rust
cur_block: Option<CurBlock>,
```
with
```rust
struct CurBlock {
    statements: List<Statement>,
    block_name: BbName,
}
```
`CurBlock` has no terminator as adding the terminator finishes the block and makes it not the current block any more.

You probably want a `fn cur_block(&mut self) -> &mut CurBlock` to easily get access to the current block with a nice panic when there is no such block.

tooling/miniutil/src/build/statement.rs
```diff
@@ -1,5 +1,29 @@
...
+    pub fn validate(&mut self, place: PlaceExpr, fn_entry: bool) {
+        self.statements.push(Statement::Validate { place, fn_entry });
+    }
+
+    pub fn storage_live(&mut self, local: PlaceExpr) {
+        let PlaceExpr::Local(name) = local else { panic!("PlaceExpr is not a local") };
+        self.statements.push(Statement::StorageLive(name));
+    }
+
+    pub fn storage_dead(&mut self, local: PlaceExpr) {
```
RalfJung, line 21:
```suggestion
    #[track_caller]
    pub fn storage_live(&mut self, local: PlaceExpr) {
        let PlaceExpr::Local(name) = local else { panic!("PlaceExpr is not a local") };
        self.statements.push(Statement::StorageLive(name));
    }

    #[track_caller]
    pub fn storage_dead(&mut self, local: PlaceExpr) {
```
That will make the panics a bit more informative

tooling/miniutil/src/build/terminator.rs
```diff
@@ -1,5 +1,243 @@
 use crate::build::*;
 
+impl FunctionBuilder {
+    fn finish_block(&mut self) {
```
RalfJung, line 4:
```suggestion
    #[track_caller]
    fn finish_block(&mut self) {
```

tooling/miniutil/src/build/terminator.rs
```diff
@@ -1,5 +1,243 @@
...
+        let cases: Vec<(&(T, &dyn Fn(&mut Self)), u32)> =
+            cases.iter().zip(self.next_block..).collect();
+        let cases_nr: Vec<(T, u32)> =
+            cases.iter().map(|((case, _), successor)| (case.clone(), successor.clone())).collect();
+
+        self.next_block += cases.len() as u32;
+        let fallback_nr = self.next_block;
+        // reserve next block nr -> if not used, function ends anyway
+        let next = self.next_block + 1;
+        self.next_block += 2;
```
RalfJung, line 199: This seems like pretty fragile reasoning for block number freshness, the `declare_block` API should help simplify that I hope.

tooling/miniutil/src/build/terminator.rs
```diff
@@ -1,5 +1,243 @@
...
+        cases: &[(T, &dyn Fn(&mut Self))],
+        fallback: G,
+    ) where
+        T: Clone + Into<Int>,
+        G: Fn(&mut Self),
+    {
+        let cases: Vec<(&(T, &dyn Fn(&mut Self)), u32)> =
+            cases.iter().zip(self.next_block..).collect();
+        let cases_nr: Vec<(T, u32)> =
+            cases.iter().map(|((case, _), successor)| (case.clone(), successor.clone())).collect();
```
RalfJung, line 193: Please add comments explaining what these are

tooling/minitest/src/tests/builder_api.rs
```diff
@@ -0,0 +1,115 @@
...
+    f.if_(eq(load(var), const_int(42u32)), |_| {}, |f| f.unreachable());
+    f.assign(ret, add(load(var), const_int(2u32)));
+    f.exit();
+    let add_two_if_42 = p.finish_function(f);
+
+    let mut start = p.function();
+
+    let l = start.local::<u32>();
+    start.storage_live(l);
+    start.call(add_two_if_42, &[by_value(const_int(42u32))], l, Ret::Yes);
```
RalfJung, line 45: I think the order of arguments for `call` should be like in `x = fun(arg1, arg2)`, i.e. with the return place first.

Also, we shouldn't have the `Ret` argument, it's annoying to repeat that all the time. Instead we can have a `call_noret` function for the rare case where a function we call cannot return.

tooling/minitest/src/tests/locks.rs
```diff
@@ -12,37 +12,45 @@ use crate::*;
...
+    let mut second = p.function();
+
+    let mut main = p.function();
+    let thread_id = main.local::<u32>();
+    let unit = main.local::<()>();
+    main.storage_live(thread_id);
+    main.storage_live(unit);
+    main.lock_create(lock);
+    main.spawn(second.get_function_name(), null(), thread_id);
+    main.call(critical, &[], unit, Ret::Yes);
```
RalfJung, line 41: Having to declare the `unit` here is annoying. What about a `call_ignoreret` function that
- declares a new local for the return value
- makes it live just before the call
- makes it dead just after the call

tooling/minitest/src/tests/locks.rs
```diff
@@ -73,40 +81,41 @@ fn lock_handover() {
...
+    main.join(load(thread_id));
+    main.exit();
+    let main = p.finish_function(main);
+
+    let ret = second.ret::<()>();
+    second.arg::<*const ()>();
+    second.call(critical, &[], ret, Ret::Yes);
+    second.return_();
+    p.finish_function(second);
+    p.set_start(main);
```
RalfJung, line 117: Please have empty lines before and after each function, to make it easier to spot where they begin/end (applies in all the tests)

tooling/miniutil/src/build/global.rs
```diff
@@ -1,5 +1,23 @@
...
+        global_by_name::<T>(name)
+    }
+
+    pub fn global_ptr<T: TypeConv>(&mut self) -> PlaceExpr {
+        let global = global_ptr::<T>();
+        let name = GlobalName(Name::from_internal(self.next_global));
+        self.next_global += 1;
+        self.globals.try_insert(name, global).unwrap();
+        global_by_name::<*const T>(name)
+    }
```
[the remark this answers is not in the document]
bifbof: agree

tooling/miniutil/src/build/mod.rs
```diff
@@ -39,6 +39,136 @@ pub use ty::*;
...
+        let name = LocalName(Name::from_internal(self.next_local));
+        self.next_local += 1;
+        self.locals.try_insert(name, t).unwrap();
+        local_by_name(name)
+    }
+
+    pub fn ret<T: TypeConv>(&mut self) -> PlaceExpr {
+        let name = LocalName(Name::from_internal(0));
```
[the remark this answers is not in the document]
bifbof: agree this is clearer, added a `ret` field for `FunctionBuilder`.

tooling/miniutil/src/build/mod.rs
```diff
@@ -39,6 +39,136 @@ pub use ty::*;
...
+            block_name: BbName(Name::from_internal(0)),
+            next_block: 1,
+            // local 0 is always ret
+            next_local: 1,
+        }
+    }
+
+    fn finish_function(mut self) -> Function {
+        // start block and return value are always 0!
+        let start = BbName(Name::from_internal(0));
```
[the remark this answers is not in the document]
bifbof: I added `start` field to `FunctionBuilder` to hard code it there instead.

tooling/minitest/src/tests/builder_api.rs
```diff
@@ -0,0 +1,115 @@
...
+    f.exit();
+    p.finish_function(f);
+    let p = p.finish_program();
+    assert_stop(p);
+}
+
+#[test]
+fn arg_and_ret_var() {
+    let mut p = ProgramBuilder::new();
+    let mut f = p.function();
```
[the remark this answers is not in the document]
bifbof: If the test has one function I opted for the whitespace. With multiple functions I opted for the block variant :)

bifbof:
I incorporated all the the review remarks into the changes @RalfJung . :)
Most important change: no more manual handling of `self.next_block`. This is now behind the `declare_block` method. The only negative point here is that I have to use `block.name_into_u32()` to use the old terminator constructor functions, and that feels a bit messy.

tooling/miniutil/src/build/mod.rs
```diff
@@ -39,6 +39,136 @@ pub use ty::*;
...
+            start: None,
+            next_fn: 0,
+            next_global: 0,
+        }
+    }
+
+    pub fn finish_program(self) -> Program {
+        let start = match self.start {
+            // TODO: decide if default start function 0 is good idea
+            None => FnName(Name::from_internal(0)),
```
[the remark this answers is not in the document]
bifbof: `finish_program` takes now `start_function` argument.

tooling/minitest/src/tests/builder_api.rs
bifbof: As discussed `return_()` would lead to UB (as start function cannot return)

tooling/minitest/src/tests/builder_api.rs
bifbof: Added a test for this case &rarr; should now throw a nice error msg.

tooling/minitest/src/tests/locks.rs
bifbof: Added `call_ignoreret`. It only works for functions with `()` return type (otherwise we would have to pass the return type), but this fits all current use cases.

tooling/miniutil/src/build/terminator.rs
```diff
@@ -1,5 +1,253 @@
...
+
+    fn goto(&mut self, dest: BbName) {
+        self.finish_block(Terminator::Goto(dest));
+    }
+
+    pub fn return_(&mut self) {
+        self.finish_block(Terminator::Return);
+    }
+
+    pub fn call_noret(&mut self, ret: PlaceExpr, f: FnName, args: &[ArgumentExpr]) {
```
RalfJung, line 29:
```suggestion
    /// Call a function that does not return.
    pub fn call_noret(&mut self, ret: PlaceExpr, f: FnName, args: &[ArgumentExpr]) {
```

tooling/miniutil/src/build/terminator.rs
```diff
@@ -1,5 +1,253 @@
...
+            let new_block = self.declare_block();
+            branch_map.try_insert(case.clone().into(), new_block.name).unwrap();
+            branches.push((branch, new_block));
+        }
+
+        let fallback_block = self.declare_block();
+        let switch = Terminator::Switch { value, cases: branch_map, fallback: fallback_block.name };
+        self.finish_block(switch);
+
+        // next block if one of the branches does not exit/return
```
RalfJung, line 209:
```suggestion
        // next block after the switch_int.
```
I think "next" is a bit ambiguous here, but I don't have a better idea either...

tooling/miniutil/src/build/terminator.rs
```diff
@@ -1,5 +1,253 @@
...
+        let mut next_block: Option<CurBlock> = None;
```
RalfJung, line 210: Ah this is tricky, you only make a next block if it is needed. Why that?

RalfJung (commented):
Thanks!

I have to run now, here are some first comments.

tooling/miniutil/src/build/mod.rs
```diff
@@ -39,6 +39,172 @@ pub use ty::*;
...
+            cur_block: None,
+            next_block: 0,
+            next_local: 0,
+        };
+        // prepare the starting block
+        fb.cur_block = Some(fb.declare_block());
+        fb
+    }
+
+    fn declare_block(&mut self) -> CurBlock {
```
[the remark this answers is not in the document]
bifbof: We don't expose the `CurBlock` to the user, blocks are so far totally internal. I needed the `CurBlock` to later set the `self.cur_block` with `self.set_block` and also to store it for the branches in `switch.`

tooling/miniutil/src/build/terminator.rs
bifbof: I use `None` to indicate that none of the branches go to the `next_block`. If we always create a `next_block`, we would get a dangling block at the end of the function.

tooling/miniutil/src/build/mod.rs
```diff
@@ -39,6 +39,172 @@ pub use ty::*;
...
+            blocks: Default::default(),
+            args: Default::default(),
+            start: BbName(Name::from_internal(0)),
+            ret: None,
+            cur_block: None,
+            next_block: 0,
+            next_local: 0,
+        };
+        // prepare the starting block
+        fb.cur_block = Some(fb.declare_block());
```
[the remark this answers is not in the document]
bifbof: I agree that relies on some implementation details. I'll add an `assert!` to make that at least explicit.

tooling/miniutil/src/build/terminator.rs
```diff
@@ -1,5 +1,253 @@
 use crate::build::*;
 
+impl FunctionBuilder {
+    #[track_caller]
+    fn finish_block(&mut self, terminator: Terminator) {
+        let bb = BasicBlock { statements: self.cur_block().statements, terminator };
```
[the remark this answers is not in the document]
bifbof: `finish_block` isn't the only function using `cur_block()`, the statements in `statements.rs` are using it as well. 
The approach with `take` would allow access to the block in two different ways.

tooling/miniutil/src/build/terminator.rs
bifbof: I just took the comment and called the block `after_switch_block` :)

tooling/miniutil/src/build/terminator.rs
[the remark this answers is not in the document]
RalfJung: > finish_block isn't the only function using cur_block(), the statements in statements.rs are using it as well.

Ah, makes sense.

I think using `take` here would still be more idiomatic Rust though.

tooling/miniutil/src/build/mod.rs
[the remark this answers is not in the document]
RalfJung: But why does that need the statements already?
What I expected is
- declare_block returns a BbName
- set_block (set_cur_block, maybe) takes a BbName

tooling/miniutil/src/build/terminator.rs
```diff
@@ -1,5 +1,255 @@
...
+
+        for (branch, block) in branches {
+            self.set_block(block);
+            branch(self);
+            // if current block not finished jump to next_block
+            if self.cur_block.is_some() {
+                if after_switch_block.is_none() {
+                    after_switch_block = Some(self.declare_block());
+                }
+                self.goto(after_switch_block.as_ref().unwrap().name);
```
RalfJung, line 225: A more idiomatic way to do this, without `unwrap`, would be something like
```rust
let jump_to_block = after_switch_block.get_or_insert_with(|| self.declare_block());
self.goto(jump_to_block);
```

tooling/miniutil/src/build/mod.rs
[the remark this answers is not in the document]
bifbof: I see your point. I changed the function accordingly &rarr; not `set_cur_block` creates the `CurBlock` and stores it into `self.cur_block`

tooling/miniutil/src/build/terminator.rs
bifbof: Thank you for this advice. Was struggling here a bit in how to do that in Rust. :)

tooling/miniutil/src/build/mod.rs
```diff
@@ -39,6 +39,177 @@ pub use ty::*;
...
+        }
+        self.cur_block = match self.cur_block {
+            None => Some(CurBlock::new(name)),
+            Some(_) => panic!("Tries to overwrite non-finished CurBlock."),
+        };
+    }
+
+    fn cur_block(&mut self) -> &mut CurBlock {
+        self.cur_block.as_mut().expect(
+            "Current Block is None. Use after terminator statement without succeeding block.",
```
RalfJung, line 131:
```suggestion
            "There is no current block. Cannot insert statement/terminator.",
```

tooling/miniutil/src/build/mod.rs
```diff
@@ -39,6 +39,177 @@ pub use ty::*;
...
+        name
+    }
+
+    fn set_cur_block(&mut self, name: BbName) {
+        if self.blocks.contains_key(name) {
+            panic!("Already inserted a block with this name.")
+        }
+        self.cur_block = match self.cur_block {
+            None => Some(CurBlock::new(name)),
+            Some(_) => panic!("Tries to overwrite non-finished CurBlock."),
```
RalfJung, line 125:
```suggestion
            Some(_) => panic!("There is an unfinished current block. Cannot set a new current block."),
```

tooling/miniutil/src/build/terminator.rs
```diff
@@ -1,5 +1,250 @@
...
+        }
+
+        let fallback_block = self.declare_block();
+        let switch = Terminator::Switch { value, cases: branch_map, fallback: fallback_block };
+        self.finish_block(switch);
+
+        // None means that every branch finished on its own and we don't need a after_switch_block
+        let mut after_switch_block: Option<BbName> = None;
+
+        // for the execution the fallback can be handled as any other branch
```
RalfJung, line 212:
```suggestion
        // Add the fallback block to the list of blocks to build.
```

tooling/miniutil/src/build/terminator.rs
```diff
@@ -1,5 +1,250 @@
...
+        branches.push((&fallback, fallback_block));
+
+        for (branch, block) in branches {
+            self.set_cur_block(block);
+            branch(self);
+            // if current block not finished jump to after_switch_block
```
RalfJung, line 218:
```suggestion
            // If the current block not finished, jump to `after_switch_block`.
```

tooling/miniutil/src/build/mod.rs
```diff
@@ -39,6 +39,177 @@ pub use ty::*;
...
+            args: Default::default(),
+            start: BbName(Name::from_internal(0)),
+            ret: None,
+            cur_block: None,
+            next_block: 0,
+            next_local: 0,
+        };
+        // prepare the starting block
+        let start_block = fb.declare_block();
+        assert_eq!(start_block, BbName(Name::from_internal(0)));
```
RalfJung, line 108:
```suggestion
        // Make sure we set `start` correctly above.
        assert_eq!(start_block, fb.start);
```

tooling/minitest/src/tests/builder_api.rs
```diff
@@ -0,0 +1,172 @@
...
+}
+
+#[test]
+#[should_panic(expected = "finish_block: there is no block to finish")]
+fn double_exit() {
+    let mut p = ProgramBuilder::new();
+
+    let mut f = p.declare_function();
+    f.exit();
+    f.exit();
```
RalfJung, line 134:
```suggestion
    f.exit(); // this is wrong, we already finished the block
```

tooling/minitest/src/tests/builder_api.rs
```diff
@@ -0,0 +1,172 @@
...
+#[should_panic(
+    expected = "Current Block is None. Use after terminator statement without succeeding block."
+)]
+fn statement_after_exit() {
+    let mut p = ProgramBuilder::new();
+
+    let mut f = p.declare_function();
+    let var = f.declare_local::<u32>();
+    f.exit();
+    f.assign(var, const_int(42_u32));
```
RalfJung, line 151:
```suggestion
    f.assign(var, const_int(42_u32)); // this is wrong, we already finished the block
```

tooling/miniutil/src/build/mod.rs
```diff
@@ -39,6 +39,177 @@ pub use ty::*;
...
+    fn cur_block(&mut self) -> &mut CurBlock {
+        self.cur_block.as_mut().expect(
+            "Current Block is None. Use after terminator statement without succeeding block.",
+        )
+    }
+
+    #[track_caller]
+    fn finish_function(mut self) -> Function {
+        if self.cur_block.is_some() {
+            panic!("Function has a dangling block. Return or exit from function.")
```
RalfJung, line 138:
```suggestion
            panic!("Function has an unfinished block. You need to return or exit from the last block.")
```

RalfJung (commented):
Just a few nits, this is almost ready to land. :)

RalfJung:
In the future please avoid these force-pushes. Github has pretty bad UI here, making it basically impossible for me to see the diff since my last review. We can always squash the commits at the end when review is done.

PR #190, opened 2024-06-13
Add shift operators

bifbof:
part of #186 

Implemented the shift operator according to [mir specification](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/enum.BinOp.html#variant.Shl).
Adapted well-formedness check, such that for shift ops the left and right type don't have to be equivalent.

### DIFF
--- a/spec/lang/syntax.md
+++ b/spec/lang/syntax.md
@@ -140,6 +140,10 @@ pub enum IntBinOp {
     /// Remainder of a division, the `%` operator.
     /// Throws UB, if the modulus (second operand) is zero.
     Rem,
+    /// Shift left `<<`
+    Shl,
+    /// Shift right `>>` (arithmetic shift for unsigned integers, logical shift for signed integers)
+    Shr,
     /// Bitwise-and two integer values.
     BitAnd,
     /// Bitwise-or two integer values.
--- a/tooling/minitest/src/tests/int.rs
+++ b/tooling/minitest/src/tests/int.rs
@@ -106,3 +106,68 @@ fn bit_int_not_works() {
     let prog = program(&[function(Ret::No, 0, &locals, &blocks)]);
     assert_stop(prog);
 }
+
+#[test]
+fn shl_works() {
+    let mut p = ProgramBuilder::new();
+
+    let mut f = p.declare_function();
+
+    f.assume(eq(shl(const_int(1u8), const_int(7u8)), const_int(128u8)));
+    f.assume(eq(shl(const_int(1u8), const_int(0u8)), const_int(1u8)));
+    f.assume(eq(shl(const_int(-1i32), const_int(1i32)), const_int(-2i32)));
+    f.assume(eq(shl(const_int(i32::MAX), const_int(1)), const_int(-2i32)));
+
+    // Shl should allow for different integer types for left and right operands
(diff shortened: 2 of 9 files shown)


RalfJung:
Blocked on https://github.com/minirust/specr/pull/10

bifbof:
Instead of adding the `assert` statement to MiniRust, I used for now a closure to deal with the `assert` statement in a separate PR.

RalfJung:
I think this shouldn't be a closure, it should be a method on `FunctionBuilder` so it can be used by all tests.

Maybe call it `assert_unchecked`? That matches the name in Rust: https://doc.rust-lang.org/nightly/std/hint/fn.assert_unchecked.html.

spec/lang/syntax.md
```diff
@@ -140,6 +140,10 @@ pub enum IntBinOp {
     /// Remainder of a division, the `%` operator.
     /// Throws UB, if the modulus (second operand) is zero.
     Rem,
+    /// Shift left `<<`
+    Shl,
+    /// Shift right `>>`
```
RalfJung, line 145:
```suggestion
    /// Shift right `>>` (arithmetic shift for unsigned integers, logical shift for signed integers)
```

RalfJung:
Can you make "bump specr version" a separate PR? Would be better to keep that apart from functional changes.

bifbof:
> I think this shouldn't be a closure, it should be a method on `FunctionBuilder` so it can be used by all tests.
> 
> Maybe call it `assert_unchecked`? That matches the name in Rust: [doc.rust-lang.org/nightly/std/hint/fn.assert_unchecked.html](https://doc.rust-lang.org/nightly/std/hint/fn.assert_unchecked.html).

What about using the intrinsic `assume`? It should do the same as `assert_unchecked` but the naming is maybe not so clear for tests.

RalfJung:
Yeah, that also works, good point. Please use `assume`.

tooling/Cargo.lock
RalfJung, line 1: Please rebase, the specr update diff is still part of this PR.

tooling/minitest/src/tests/int.rs
```diff
@@ -106,3 +106,67 @@ fn bit_int_not_works() {
...
+    let mut f = p.declare_function();
+
+    // Logical shr for unsigned integers
+    f.assume(eq(shr(const_int(u8::MAX), const_int(7u8)), const_int(1u8)));
+    f.assume(eq(shr(const_int(1u8), const_int(0)), const_int(1u8)));
+
+    // Arithmetic shr for signed integers
+    f.assume(eq(shr(const_int(-4i16), const_int(1u16)), const_int(-2i16)));
+    f.assume(eq(shr(const_int(-1i32), const_int(1)), const_int(-1i32)));
+    f.assume(eq(shr(const_int(i32::MAX), const_int(1)), const_int(i32::MAX / 2)));
```
RalfJung, line 152:
```suggestion
    f.assume(eq(shr(const_int(1i32), const_int(1)), const_int(0i32)));
    f.assume(eq(shr(const_int(i32::MAX), const_int(1)), const_int(i32::MAX / 2)));
```
Just for completeness' sake.

tooling/Cargo.lock
bifbof: done :)

RalfJung:
Looks good, thanks. :)

PR #194, opened 2024-06-19
Revert "bump ui_test to 0.23"

RalfJung:
Reverts minirust/minirust#188. Due to https://github.com/oli-obk/ui_test/issues/238, test failures are near impossible to debug with ui_test 0.23, so let's go back to the previous version.


oli-obk:
0.24 has been released with a fix for these issues

PR #200, opened 2024-06-21
minimize: add boolean comparison operators

bifbof:
Just because it was easy to do :)

part of #186

Added helper function`bin_eq` to `miniutil` because I found it useful.

### DIFF
--- a/tooling/minimize/src/rvalue.rs
+++ b/tooling/minimize/src/rvalue.rs
@@ -38,12 +38,14 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
                     (ShlUnchecked, Type::Int(_)) => BinOp::Int(IntBinOp::ShlUnchecked),
                     (ShrUnchecked, Type::Int(_)) => BinOp::Int(IntBinOp::ShrUnchecked),
 
-                    (Lt, Type::Int(_)) => BinOp::IntRel(IntRel::Lt),
-                    (Le, Type::Int(_)) => BinOp::IntRel(IntRel::Le),
-                    (Gt, Type::Int(_)) => BinOp::IntRel(IntRel::Gt),
-                    (Ge, Type::Int(_)) => BinOp::IntRel(IntRel::Ge),
-                    (Eq, Type::Int(_)) => BinOp::IntRel(IntRel::Eq),
-                    (Ne, Type::Int(_)) => BinOp::IntRel(IntRel::Ne),
+                    // We use the int relations also for bools
+                    // (and insert a cast, see below).
+                    (Lt, Type::Int(_) | Type::Bool) => BinOp::IntRel(IntRel::Lt),
+                    (Le, Type::Int(_) | Type::Bool) => BinOp::IntRel(IntRel::Le),
+                    (Gt, Type::Int(_) | Type::Bool) => BinOp::IntRel(IntRel::Gt),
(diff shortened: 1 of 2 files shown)


tooling/miniutil/src/build/expr.rs
```diff
@@ -189,6 +189,9 @@ pub fn bool_or(l: ValueExpr, r: ValueExpr) -> ValueExpr {
 pub fn bool_xor(l: ValueExpr, r: ValueExpr) -> ValueExpr {
     bool_binop(BoolBinOp::BitXor, l, r)
 }
+pub fn bool_eq(l: ValueExpr, r: ValueExpr) -> ValueExpr {
```
RalfJung, line 192: I don't quite see the point of this function... it uses a different encoding than minimize, and it's not even used by anything?

tooling/minimize/src/rvalue.rs
```diff
@@ -38,12 +38,12 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
                     (ShlUnchecked, Type::Int(_)) => BinOp::Int(IntBinOp::ShlUnchecked),
                     (ShrUnchecked, Type::Int(_)) => BinOp::Int(IntBinOp::ShrUnchecked),
 
-                    (Lt, Type::Int(_)) => BinOp::IntRel(IntRel::Lt),
-                    (Le, Type::Int(_)) => BinOp::IntRel(IntRel::Le),
-                    (Gt, Type::Int(_)) => BinOp::IntRel(IntRel::Gt),
-                    (Ge, Type::Int(_)) => BinOp::IntRel(IntRel::Ge),
-                    (Eq, Type::Int(_)) => BinOp::IntRel(IntRel::Eq),
-                    (Ne, Type::Int(_)) => BinOp::IntRel(IntRel::Ne),
+                    (Lt, Type::Int(_)) | (Lt, Type::Bool) => BinOp::IntRel(IntRel::Lt),
```
RalfJung, line 41:
```suggestion
                    // We use the int relations als for bools
                    // (and insert a cast, see below).
                    (Lt, Type::Int(_) | Type::Bool) => BinOp::IntRel(IntRel::Lt),
```

tooling/minimize/src/rvalue.rs
```diff
@@ -54,6 +54,23 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
                         rs::span_bug!(span, "Binary Op {op:?} not supported for type {lty_smir}."),
                 };
 
+                // cast boolean used in comparisons to Int to use them in IntRel ops
+                let (l, r) = match (bin_op, lty) {
```
RalfJung, line 58: I think it's easier to check `(op, lty)` against `(BinOp::IntRel(_), Type::Bool)`.

tooling/miniutil/src/build/expr.rs
bifbof: I thought it can be useful when writing programs with miniutil. I removed it.

tooling/minimize/src/rvalue.rs
RalfJung: Note that I didn't just add a comment, I also suggested a way to simplify the pattern. ;)

bifbof: ah nice :)

RalfJung:
Looks good, thanks!

PR #204, opened 2024-06-23
Add `IntWithOverflow` to `BinOp`

bifbof:
part of #186 

Couldn't test the translation of `minimize` as for that the Rust intrinsic `unlikely` is missing.
Added some commented out tests until that is fixed.

### DIFF
--- a/spec/lang/types.md
+++ b/spec/lang/types.md
@@ -189,5 +189,17 @@ impl IntType {
     pub fn bring_in_bounds(&self, i: Int) -> Int {
         i.bring_in_bounds(self.signed, self.size)
     }
+
+    /// Generate the return type for IntWithOverflow
+    pub fn with_overflow<T: Target>(&self) -> Type {
+        // Define a tuple type with two fields: An integer followed directly by a boolean.
+        let fields = list![(Size::ZERO, Type::Int(*self)), (self.size, Type::Bool)];
+        // Alignment is always the one of the integer as boolean has align requirement 1.
+        let align = self.align::<T>();
+        // The total size is `self.size + 1` rounded up to the next multiple of `align`.
+        // Since `self.size` is already a multiple of `align`, we can compute this as follows:
+        let size = self.size + Size::from_bytes(align.bytes()).unwrap();
+        Type::Tuple { fields, size, align }
+    }
 }
 ```
--- a/spec/lang/well-formed.md
+++ b/spec/lang/well-formed.md
@@ -341,6 +341,13 @@ impl ValueExpr {
                         }
                         Type::Int(left)
                     }
+                    IntWithOverflow(_int_op) => {
+                        let Type::Int(int_ty) = left else {
+                            throw_ill_formed!("BinOp::IntWithOverflow: invalid left type");
+                        };
+                        ensure_wf(right == Type::Int(int_ty), "BinOp::IntWithOverflow: invalid right type")?;
+                        int_ty.with_overflow::<T>()
+                    }
                     IntRel(_int_rel) => {
                         let Type::Int(int_ty) = left else {
                             throw_ill_formed!("BinOp::IntRel: invalid left type");
--- a/tooling/minimize/src/bb.rs
+++ b/tooling/minimize/src/bb.rs
@@ -182,7 +182,7 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
         if let rs::InstanceDef::Intrinsic(def_id) = instance.def {
             // A Rust intrinsic.
             let intrinsic_name = self.tcx.item_name(def_id);
-            return match intrinsic_name.as_str() {
+            match intrinsic_name.as_str() {
                 "assert_inhabited" | "assert_zero_valid" | "assert_mem_uninitialized_valid" => {
                     let ty = instance.args.type_at(0);
                     let requirement =
@@ -194,10 +194,48 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
 
                     if should_panic {
                         // FIXME trigger a panic instead of UB
-                        Terminator::Unreachable
+                        return Terminator::Unreachable;
                     } else {
-                        Terminator::Goto(self.bb_name_map[&target.unwrap()])
+                        return Terminator::Goto(self.bb_name_map[&target.unwrap()]);
                     }
+                }
+                "unlikely" | "likely" => {
+                    // FIXME: use the "fallback body" provided in the standard library.
+                    // translate `unlikely`, `likely` intrinsic into functions that can be called
+                    let function_name = self.cx.get_fn_name(instance);
+                    if !self.functions.contains_key(function_name) {
+                        let ret = LocalName(Name::from_internal(0));
+                        let arg = LocalName(Name::from_internal(1));
+
+                        let b0_name = BbName(Name::from_internal(0));
+                        let assign = Statement::Assign {
+                            destination: PlaceExpr::Local(ret),
+                            source: ValueExpr::Load { source: GcCow::new(PlaceExpr::Local(arg)) },
+                        };
+                        let b0 = BasicBlock {
+                            statements: list![assign],
+                            terminator: Terminator::Return,
+                        };
+
+                        let mut blocks = Map::new();
+                        blocks.insert(b0_name, b0);
+
+                        let mut locals = Map::new();
+                        locals.insert(ret, <bool>::get_type());
+                        locals.insert(arg, <bool>::get_type());
+
+                        let function = Function {
+                            locals,
+                            args: list![arg],
+                            ret,
+                            blocks,
+                            start: b0_name,
+                            calling_convention: CallingConvention::Rust,
+                        };
+                        // We can unwrap because we already checked that `function_name` is not in the `functions` map.
+                        self.cx.functions.try_insert(function_name, function).unwrap();
+                    }
+                    // Fall through to the regular function call handling below.
                 }
                 name => panic!("unsupported Rust intrinsic `{}`", name),
             };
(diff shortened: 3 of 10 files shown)


RalfJung:
FWIW that intrinsic should be translated to a NOP -- evaluating the boolean argument but ignoring the result.

spec/lang/types.md
```diff
@@ -187,5 +187,14 @@ impl IntType {
     pub fn bring_in_bounds(&self, i: Int) -> Int {
         i.bring_in_bounds(self.signed, self.size)
     }
+
+    /// Generate the return type for IntWithOverflow
+    pub fn overflow_type<T: Target>(&self) -> Type {
+        let self_ty = Type::Int(*self); 
+        let fields = list![(Size::ZERO, self_ty), (self_ty.size::<T>(), Type::Bool)];
+        let size = self_ty.size::<T>() + self_ty.size::<T>();
```
RalfJung, line 195: This does not seem right -- if `self_ty` is `u16`, size must be 4 or else it is not a multiple of the alignment. Currently you are generating a not-well-formed type in that case.

spec/lang/types.md
```diff
@@ -187,5 +187,14 @@ impl IntType {
...
+        let align = self_ty.align::<T>();
+        Type::Tuple{ fields, size, align }
```
RalfJung, line 197:
```suggestion
        Type::Tuple { fields, size, align }
```

spec/lang/well-formed.md
```diff
@@ -341,6 +341,19 @@ impl ValueExpr {
...
+                    IntWithOverflow(int_op) => {
+                        let Type::Int(int_ty) = left else {
+                            throw_ill_formed!("BinOp::IntWithOverflow: invalid left type");
+                        };
+                        ensure_wf(right == Type::Int(int_ty), "BinOp::IntWithOverflow: invalid right type")?;
+                        match int_op {
+                            IntBinOp::Add | IntBinOp::Sub | IntBinOp::Mul => {}
+                            _ => throw_ill_formed!(
+                                "BinOp::IntWithOverflow: op is not available as overflow variant"
+                            ),
```
RalfJung, line 353: Instead of checking this dynamically, please use a separate enum so that it is impossible to even use a wrong operation here.

bifbof:
> FWIW that intrinsic should be translated to a NOP -- evaluating the boolean argument but ignoring the result.

But is the `unlikely`/`likely` not [behaving like an identity function?](https://doc.rust-lang.org/src/core/intrinsics.rs.html#1009).
I have a (ugly) version running with a new MiniRust `Identity` intrinsic, as I could not use an existing terminator/intrinsic such that it acts like the identity function.

spec/lang/types.md
bifbof: Maybe I'm missing something but `self_ty.size::<T>()  = 2` and `self_ty.size::<T>() + self_ty.size::<T>() = 4`.
And double the size is always a multiple of the alignment.

spec/lang/well-formed.md
bifbof: Added the `IntBinOpWithOverflow`

spec/lang/types.md
bifbof: tried to increase the clarity by changing the line to:
```rust
let size = self_ty.size::<T>() * Int::from(2); 
```

RalfJung: d'oh, I misread.
Please add a comment explaining all that is going on here -- all this only makes sense because the 2nd field has alignment 1 and size 1.

bifbof:
Added review changes. Now `unlikely`/`likely` intrinsic are translated to functions that can be called.
Plus explaining comments for the `overflow_ty`.

spec/lang/types.md
```diff
@@ -189,5 +189,17 @@ impl IntType {
     pub fn bring_in_bounds(&self, i: Int) -> Int {
         i.bring_in_bounds(self.signed, self.size)
     }
+
+    /// Generate the return type for IntWithOverflow
+    pub fn overflow_type<T: Target>(&self) -> Type {
```
RalfJung, line 194:
```suggestion
    pub fn with_overflow<T: Target>(&self) -> Type {
```

spec/lang/types.md
```diff
@@ -189,5 +189,17 @@ impl IntType {
...
+        // Define a tuple type with two fields: An integer followed directly by a boolean.
+        let self_ty = Type::Int(*self);
+        let fields = list![(Size::ZERO, self_ty), (self_ty.size::<T>(), Type::Bool)];
+        // Alignment is always the one of the integer as boolean has align requirement 1.
+        let align = self_ty.align::<T>();
+        // As size is multiple of alignment boolean occupies space of integer size.
+        let size = self_ty.size::<T>() * Int::from(2);
```
RalfJung, line 201: This is actually wrong when the integer has size != align, isn't it? For instance, `u64` on a 32-bit-target works like that.

tooling/minimize/src/bb.rs
```diff
@@ -194,9 +194,46 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
+
+                        let function = Function {
+                            locals,
+                            args: list![arg],
+                            ret,
+                            blocks,
+                            start: b0_name,
+                            calling_convention: CallingConvention::Rust,
+                        };
+                        // can unwrap as checked if `function_name` already in functions map
```
RalfJung, line 235:
```suggestion
                        // We can unwrap because we already checked that `function_name` is not in the `functions` map.
```

tooling/minimize/src/bb.rs
```diff
@@ -194,9 +194,46 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
+                        self.cx.functions.try_insert(function_name, function).unwrap();
                     }
```
RalfJung, line 237:
```suggestion
                    }
                    // Fall through to the regular function call handling below.
```

tooling/minimize/tests/pass/ops.rs
```diff
@@ -29,6 +29,8 @@ fn main() {
     assert!(unsafe { black_box(i32::MIN).unchecked_shl(1u32) } == 0);
     assert!(unsafe { black_box(-1_i32).unchecked_shr(1u32) } == -1);
     assert!(unsafe { black_box(84_i32).unchecked_shr(1u32) } == 42);
+    assert!(black_box(41_i32).checked_add(1) == Some(42));
+    assert!(black_box(i32::MAX).checked_add(1) == None);
```
RalfJung, line 33: Can you also add calls to `overflowing_add`, as well as sub and mul?

spec/lang/types.md
bifbof: Oh, I did not know this is possible. I changed it such that we now add size and alignment.

tooling/minimize/src/bb.rs
bifbof: Proper sentences! :)

spec/lang/types.md
```diff
@@ -189,5 +189,17 @@ impl IntType {
...
+
+    /// Generate the return type for IntWithOverflow
+    pub fn with_overflow<T: Target>(&self) -> Type {
+        // Define a tuple type with two fields: An integer followed directly by a boolean.
+        let self_ty = Type::Int(*self);
+        let fields = list![(Size::ZERO, self_ty), (self_ty.size::<T>(), Type::Bool)];
+        // Alignment is always the one of the integer as boolean has align requirement 1.
+        let align = self_ty.align::<T>();
+        // As size is multiple of alignment boolean occupies space of integer size.
+        let size = self_ty.size::<T>() + Size::from_bytes(align.bytes()).unwrap();
```
RalfJung, line 201:
```suggestion
        let size = self.size + Size::from_bytes(align.bytes()).unwrap();
```

spec/lang/types.md
```diff
@@ -189,5 +189,17 @@ impl IntType {
     pub fn bring_in_bounds(&self, i: Int) -> Int {
         i.bring_in_bounds(self.signed, self.size)
     }
+
+    /// Generate the return type for IntWithOverflow
+    pub fn with_overflow<T: Target>(&self) -> Type {
+        // Define a tuple type with two fields: An integer followed directly by a boolean.
+        let self_ty = Type::Int(*self);
+        let fields = list![(Size::ZERO, self_ty), (self_ty.size::<T>(), Type::Bool)];
```
RalfJung, line 197:
```suggestion
        let fields = list![(Size::ZERO, self_ty), (self.size, Type::Bool)];
```

spec/lang/types.md
```diff
@@ -189,5 +189,17 @@ impl IntType {
...
+        // Alignment is always the one of the integer as boolean has align requirement 1.
+        let align = self_ty.align::<T>();
+        // As size is multiple of alignment boolean occupies space of integer size.
```
RalfJung, line 200:
```suggestion
        // The total size is `self.size + 1` rounded up to the next multiple of `align`.
        // Since `self.size` is already a multiple of `align`, we can compute this as follows:
```

spec/lang/types.md
bifbof: removed `self_ty` altogether.

tooling/minimize/src/bb.rs
RalfJung: Yes please. :) I had to read your comment 3 times before I understood it...

tooling/minimize/src/bb.rs
```diff
@@ -194,11 +194,49 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
                         // FIXME trigger a panic instead of UB
-                        Terminator::Unreachable
+                        return Terminator::Unreachable;
                     } else {
-                        Terminator::Goto(self.bb_name_map[&target.unwrap()])
+                        return Terminator::Goto(self.bb_name_map[&target.unwrap()]);
                     }
                 }
+                "unlikely" | "likely" => {
+                    // FIXME: translate this into an assignment statement + goto terminator
```
RalfJung, line 203:
```suggestion
                    // FIXME: use the "fallback body" provided in the standard library.
```

RalfJung:
Looks good, thanks. :) Just one final comment nit. Please squash the commits.

bifbof:
> Looks good, thanks. :) Just one final comment nit. Please squash the commits.

done 👍🏼

PR #206, opened 2024-06-26
minimize: add caching of types

### LINKED ISSUE #161: Minimize: cache type minimization
The type minimization currently doesn't reuse the result, which can lead to exponential blow up. `tests/pass/enums.rs` has an example of a type which currently cannot be minimized with 32GB of ram due to its construction.

bifbof:
closes #161 

Now minimize can do the translation fast but now the type explodes in MiniRust instead.
Thus if we run the test mentioned in #161 we still run out of memory.

### DIFF
--- a/tooling/minimize/src/ty.rs
+++ b/tooling/minimize/src/ty.rs
@@ -15,12 +15,16 @@ impl<'tcx> Ctxt<'tcx> {
         self.layout_of(smir::internal(self.tcx, ty))
     }
 
-    pub fn translate_ty_smir(&self, ty: smir::Ty, span: rs::Span) -> Type {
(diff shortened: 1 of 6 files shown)


tooling/minimize/src/ty.rs
```diff
@@ -1,5 +1,11 @@
+use std::cell::RefCell;
+
 use crate::*;
 
+thread_local! {
+    static TY_CACHE: RefCell<HashMap<smir::Ty, Type>> = RefCell::new(HashMap::new())
+}
```
RalfJung, line 7: The cache could live in `Ctxt`, no need to make it a thread-local

bifbof: moved it to `Ctxt`

RalfJung:
Looks good, thank!

PR #207, opened 2024-06-30
miniutil: add assume test to `mod.rs`

bifbof:
Small bugfix, `assume.rs` wasn't added to `mod.rs` and was therefore never tested.


RalfJung:
Thanks! That makes me wonder whether there are any other files we forgot to add...

Issue #172, opened 2024-04-05
Some TerminatorKind are silently ignored


RalfJung:
We should find better ways to translate these:
```rust
            // those are IGNORED currently.
            rs::TerminatorKind::Drop { target, .. } | rs::TerminatorKind::Assert { target, .. } =>
                Terminator::Goto(self.bb_name_map[&target]),
```

PR #208, opened 2024-07-01
add `panic` and `assert`

bifbof:
part of #172 

Add `panic` as intrinsic and `assert` as Terminator.
For now just added strings `"exited"` or `"panicked"` to `MachineStop` to differentiate between the two.

### DIFF
--- a/spec/lang/step/intrinsics.md
+++ b/spec/lang/step/intrinsics.md
@@ -96,6 +96,22 @@ impl<M: Memory> Machine<M> {
 }
 ```
 
+Currently `Panic` carries no message and aborts directly.
+
+```rust
+impl<M: Memory> Machine<M> {
+    fn eval_intrinsic(
+        &mut self,
+        IntrinsicOp::Panic: IntrinsicOp,
+        arguments: List<(Value<M>, Type)>,
+        ret_ty: Type,
+    ) -> NdResult<Value<M>> {
+        // Stop machine immediatly without any additional checks.
+        throw_abort!("we panicked");
+    }
+}
+```
+
 ## UB control
 
 ```rust
--- a/spec/lang/syntax.md
+++ b/spec/lang/syntax.md
@@ -349,6 +349,7 @@ pub enum IntrinsicLockOp {
 pub enum IntrinsicOp {
     Assume,
     Exit,
+    Panic,
     PrintStdout,
     PrintStderr,
     Allocate,
--- a/spec/prelude/main.md
+++ b/spec/prelude/main.md
@@ -23,6 +23,8 @@ pub enum TerminationInfo {
     Ub(String),
     /// The program was executed and the machine stopped without error.
     MachineStop,
+    /// The program terminated with a panic
+    Abort(String),
     /// The program was ill-formed.
     IllFormed(String),
     /// The program did not terminate but no thread can make progress.
@@ -42,6 +44,11 @@ macro_rules! throw_ub {
     ($($tt:tt)*) => {
         do yeet TerminationInfo::Ub(format!($($tt)*))
     };
+}
+macro_rules! throw_abort {
+    ($($tt:tt)*) => {
+        do yeet TerminationInfo::Abort(format!($($tt)*))
+    }
 }
 macro_rules! throw_machine_stop {
     () => {
--- a/tooling/minimize/src/bb.rs
+++ b/tooling/minimize/src/bb.rs
@@ -158,9 +158,29 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
                 Terminator::Switch { value, cases, fallback }
             }
             rs::TerminatorKind::Unreachable => Terminator::Unreachable,
+            rs::TerminatorKind::Assert { cond, expected, target, .. } => {
+                let mut condition = self.translate_operand(cond, span);
+                // Check equality of `condition` and `expected`.
+                // We do this by inverting `condition` if `expected` is false
+                // and then checking if `condition` is true.
+                if !expected {
+                    condition = build::not(condition);
+                }
+
+                // Create panic block in case of `expected != condition`
+                let panic_bb = self.fresh_bb_name();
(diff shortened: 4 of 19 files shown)


spec/lang/syntax.md
```diff
@@ -326,6 +326,13 @@ pub enum Terminator {
     },
     /// Return from the current function.
     Return,
+    /// Assert the condition if false panic else go `next_block`.
+    Assert {
```
RalfJung, line 330: This should not be a primitive of the language, since it can be encoded with if-then-else and the "panic" intrinsic.

tooling/minimize/src/bb.rs
```diff
@@ -230,8 +245,13 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
                 next_block: target.as_ref().map(|t| self.bb_name_map[t]),
             }
         } else if is_panic_fn(&instance.to_string()) {
-            // We can't translate this call, it takes a string. So as a special hack we just make this `Unreachable`.
-            Terminator::Unreachable
+            // We can't translate this call, it takes a string. As a hack we just ignore the argument.
```
RalfJung, line 248: Nice, much better than the previous hack :)

tooling/minimize/src/bb.rs
```diff
@@ -230,8 +245,13 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
+            Terminator::Intrinsic {
+                intrinsic: IntrinsicOp::Panic,
+                arguments: list!(),
```
RalfJung, line 251:
```suggestion
                arguments: list![],
```

tooling/minimize/tests/pass/assert.rs
```diff
@@ -0,0 +1,8 @@
+fn main() {
+    // An assert is put before divisions, so this should panic but not create UB.
+    let _ = 42 / blackbox(0);
+}
+
+fn blackbox<T>(i: T) -> T {
+    i
+}
```
RalfJung, line 8: The blackbox stuff shouldn't be needed.

spec/prelude/main.md
```diff
@@ -22,7 +22,7 @@ pub enum TerminationInfo {
     /// The execution encountered undefined behaviour.
     Ub(String),
     /// The program was executed and the machine stopped without error.
-    MachineStop,
+    MachineStop(String),
```
RalfJung, line 25: As the comment says, this was meant for stopping *without* an error. But now you are using it for stopping with an error as well. That's also why panics now silently work as "pass" tests, which is not great.

Instead, maybe add a new `TerminationInfo` variant called `MachineAbort(String)`, and then minimize's `main.rs` can print the abort message and exit with a non-zero exit code?

tooling/minimize/tests/pass/array_out_of_bounds.rs
RalfJung, line 1: This program panics, right? That should be somehow covered by the test... and also the name of the test should be adjusted to indicate that. (and the FIXME removed)

spec/lang/syntax.md
bifbof: No minimize translates the program with some if-then-else blocks. It is a bit hacky.

tooling/minimize/tests/pass/assert.rs
bifbof: Added `#[allow(unconditional_panic)]` instead.

tooling/minimize/tests/pass/array_out_of_bounds.rs
bifbof: Panic now raises an error. Moved this and other test to `tests/panic` and check for the error msg.

spec/lang/step/intrinsics.md
```diff
@@ -96,6 +96,22 @@ impl<M: Memory> Machine<M> {
...
+```rust
+impl<M: Memory> Machine<M> {
+    fn eval_intrinsic(
+        &mut self,
+        IntrinsicOp::Panic: IntrinsicOp,
+        arguments: List<(Value<M>, Type)>,
+        ret_ty: Type,
+    ) -> NdResult<Value<M>> {
+        // Stop machine immediatly without any additional checks.
+        throw_abort!("called panic intrinsic");
```
RalfJung, line 110:
```suggestion
        throw_abort!("we panicked");
```

tooling/minimize/tests/ub/div_by_zero.rs
```diff
@@ -1,8 +1,11 @@
+#![allow(internal_features)] 
+#![feature(core_intrinsics)]
 extern crate intrinsics;
 use intrinsics::*;
 
 fn main() {
-    print(1 / black_box(0));
+    // use `unchecked_div` to avoid `assert` != 0 before division
```
RalfJung, line 7:
```suggestion
    // Use `unchecked_div` to make div-by-zero UB (rather than panic).
```

tooling/minimize/src/bb.rs
```diff
@@ -158,9 +158,38 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
+                        ret: zst_place(),
+                        next_block: None,
+                    },
+                };
+                self.blocks.try_insert(panic_bb, panic_block).unwrap();
+
+                let nop_bb = self.fresh_bb_name();
+                let nop_block =
+                    BasicBlock { statements: list![], terminator: Terminator::Goto(next_block) };
+                self.blocks.try_insert(nop_bb, nop_block).unwrap();
```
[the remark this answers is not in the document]
bifbof: Yes that works better.

RalfJung:
Looks good, thanks. :)

Please squash the commits (without changing the base commit)

bifbof:
> Please squash the commits (without changing the base commit)

I squashed the commits and kept the commit they are based from, but for this I still need the merge from master to deal with the conflict.
Is that what you wanted?

RalfJung:
Ah if there's a conflict then rebase please.

bifbof:
> Ah if there's a conflict then rebase please.

Done :)

Issue #176, opened 2024-04-19
raw_eq intrinsic is not supported


RalfJung:
This occurs for instance [here](https://github.com/minirust/minirust/blob/c91adf155212f6197631cb560247ec604b1f003d/tooling/minimize/tests/pass/union.rs#L73), via the PartialEq impl for arrays of integers.

bifbof:
Here is a minimal testcase for this issue: 

```rust
fn main() {
    let _ = [0] == [0];
}
```

outputs:
```
thread 'rustc' panicked at minimize/src/bb.rs:205:25:
unsupported Rust intrinsic `raw_eq`
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

PR #211, opened 2024-07-06
Add RawEq Support.

sun-jacobi:
This patch fixes https://github.com/minirust/minirust/issues/176.

Specifically, this patch adds support for an operation to determine whether the raw bytes of the two values are equal, 
corresponding to the [`std::intrinsics::raw_eq`](https://doc.rust-lang.org/std/intrinsics/fn.raw_eq.html) in rust standard library.

### DIFF
--- a/spec/lang/step/intrinsics.md
+++ b/spec/lang/step/intrinsics.md
@@ -342,6 +342,68 @@ impl<M: Memory> Machine<M> {
     }
 }
 ```
+## Raw equality
+```rust
+impl<M: Memory> Machine<M> {
+    fn load_raw_data(&mut self, ptr : Pointer<<M as Memory>::Provenance>, ptr_ty : PtrType) -> Result<List<u8>> {
+        // We need the pointee layout to determine how many bytes to load.
+        let PtrType::Ref { pointee, .. } = ptr_ty else {
+            throw_ub!("invalid argument to `RawEq` intrinsic: not a reference");
+        };
+        let Layout { size, align, .. } = pointee;
+        let bytes = self.mem.load(ptr, size, align, Atomicity::None)?;
+
+        // FIXME: This violates provenance monotonicity
+        if bytes.any(|byte| byte.provenance() != None) {
+            throw_ub!("invalid argument to `RawEq` intrinsic: byte has provenance");
+        }
+
+        let Some(data) =  bytes.try_map(|byte| byte.data()) else {
+            throw_ub!("invalid argument to `RawEq` intrinsic: byte is uninitialized");
+        };
+
+        Ok(data)
+    }
+
+    fn eval_intrinsic(
+        &mut self,
+        IntrinsicOp::RawEq: IntrinsicOp,
+        arguments: List<(Value<M>, Type)>,
+        ret_ty: Type,
+    ) -> NdResult<Value<M>> {
+        if arguments.len() != 2 {
+            throw_ub!("invalid number of arguments for `RawEq` intrinsic");
+        }
+        if ret_ty != Type::Bool {
+            throw_ub!("invalid return type for `RawEq` intrinsic")
+        }
+
+        let (left, l_ty) = (arguments).index_at(0);
+        let (right, r_ty) = (arguments).index_at(1);
+
+        if l_ty != r_ty {
+            throw_ub!("invalid arguments to `RawEq` intrinsic: types of arguments are not identical");
+        }
+
+        let Value::Ptr(left) = left else {
+            throw_ub!("invalid first argument to `RawEq` intrinsic: not a pointer");
+        };
+
+        let Value::Ptr(right) = right else {
+            throw_ub!("invalid second argument to `RawEq` intrinsic: not a pointer");
+        };
+
+        let Type::Ptr(l_ty) = l_ty else {
+            throw_ub!("invalid argument type to `RawEq` intrinsic: not a pointer");
+        };
+
+        let left_data = self.load_raw_data(left, l_ty)?;
+        let right_data = self.load_raw_data(right, l_ty)?;
+
+        ret(Value::Bool(left_data == right_data))
+    }
+}
+```
 
 ## Atomic accesses
 
--- a/tooling/minimize/src/bb.rs
+++ b/tooling/minimize/src/bb.rs
@@ -200,6 +200,17 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
                         return Terminator::Goto(self.bb_name_map[&target.unwrap()]);
                     }
                 }
+                "raw_eq" =>
+                    return Terminator::Intrinsic {
+                        intrinsic: IntrinsicOp::RawEq,
+                        arguments: args
+                            .iter()
+                            .map(|x| self.translate_operand(&x.node, x.span))
+                            .collect(),
+                        ret: self.translate_place(&destination, span),
+                        next_block: target.as_ref().map(|t| self.bb_name_map[t]),
+                    },
+
                 "unlikely" | "likely" => {
                     // FIXME: use the "fallback body" provided in the standard library.
                     // translate `unlikely`, `likely` intrinsic into functions that can be called
--- a/tooling/minimize/tests/pass/union.rs
+++ b/tooling/minimize/tests/pass/union.rs
@@ -69,10 +69,5 @@ fn main() {
     assert!(extract_struct(u) == TestStruct(12, 1200));
     
     let u = ArrayUnion { data: [[42;3], [12;3]] };
-    let a = extract_array(u);
-    assert!(a[0][1] == 42);
-    assert!(a[0][2] == 42);
-    assert!(a[1][0] == 12);
-    assert!(a[1][1] == 12);
-    assert!(a[1][2] == 12);
+    assert!(extract_array(u) == [[42;3], [12;3]]);
 }
--- a/tooling/minitest/src/tests/raw_eq.rs
+++ b/tooling/minitest/src/tests/raw_eq.rs
@@ -1,0 +1,215 @@
+use crate::*;
+
+#[test]
+fn true_raw_eq() {
+    let mut p = ProgramBuilder::new();
+    let mut f = p.declare_function();
+    let dest = f.declare_local::<bool>();
+    let left = f.declare_local::<[u8; 2]>();
+    let right = f.declare_local::<[u8; 2]>();
+
+    let pointee = layout(size(2), align(1));
+    let ptr_ty = ref_ty(pointee);
+
+    let left_ptr = addr_of(left, ptr_ty);
+    let right_ptr = addr_of(right, ptr_ty);
+
+    f.storage_live(dest);
+    f.storage_live(left);
+    f.storage_live(right);
+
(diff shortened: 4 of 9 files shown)


spec/lang/step/operators.md
```diff
@@ -70,7 +70,7 @@ impl<M: Memory> Machine<M> {
 impl<M: Memory> Machine<M> {
     #[specr::argmatch(operator)]
     fn eval_bin_op(
-        &self,
+        &mut self,
```
RalfJung, line 73: These functions deliberately took `&self`, please don't change that.

You can make RawEq an intrinsic instead of a BinOp.

spec/lang/step/operators.md
```diff
@@ -250,6 +250,54 @@ impl<M: Memory> Machine<M> {
...
+### Raw Equality
+```rust
+impl<M: Memory> Machine<M> {
+    fn load_raw_data(&mut self, ptr : Pointer<<M as Memory>::Provenance>, ptr_ty : PtrType) -> Result<List<u8>> {
+        let PtrType::Ref { mutbl, pointee } = ptr_ty else {
+            panic!("non-reference input to raw equality");
+        };
+        if mutbl == Mutability::Mutable {
+            panic!("mutable reference to raw equality");
+        }
```
RalfJung, line 262: I don't see why we should reject this? Seems fine if people give us mutable references.

spec/lang/step/operators.md
```diff
@@ -250,6 +250,54 @@ impl<M: Memory> Machine<M> {
...
+        let Layout { size, align, ..} = pointee;
+        let bytes = self.mem.load(ptr, size, align, Atomicity::None)?;
+        Ok(bytes.map(|byte| {
+            let Some(data) = byte.data() else { panic!("byte is uninitialized"); };
+            if let Some(_) = byte.provenance() { panic!("byte has provenance"); }
```
RalfJung, line 267: These should not panic. A panic indicates a bug in the interpreter, but actual user programs can call raw_eq on uninit bytes or those with provenance. It is UB to do so, but that's a bug in the program, not the interpreter.

tooling/minimize/src/bb.rs
```diff
@@ -200,6 +200,57 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
                         return Terminator::Goto(self.bb_name_map[&target.unwrap()]);
                     }
                 }
+                "raw_eq" => {
+                    let function_name = self.cx.get_fn_name(instance);
+                    if !self.functions.contains_key(function_name) {
```
RalfJung, line 205: No need to synthesize a function here. `likely`/`unlikely` do that but it's kind of a hack.

If you make RawEq an intrinsic this will anyway be easier, you can just create a call to the intrinsic here.

tooling/minimize/tests/pass/union.rs
RalfJung, line 1: Please also add a test in tooling/minimize/tests/pass/array.rs; after all this is more about arrays than about unions.

RalfJung (commented):
Thanks, that's a good start! I have some comments though.

On the test side, please also add a `raw_eq` file in `minitest/src/tests` that tests the intrinsic directly. Cruciall it should also cover the UB cases (uninit / provenance).

tooling/minimize/tests/pass/union.rs
sun-jacobi: Do you think [tooling/minimize/tests/pass/raweq.rs](https://github.com/minirust/minirust/pull/211/files#diff-685695c636dfce66b404f488c9ad4d092ecf9bcbd2e2c9ab3f1c6701b9833735) is necessary?

RalfJung: I'd say merge that into `array.rs`

spec/lang/step/intrinsics.md
```diff
@@ -342,6 +342,70 @@ impl<M: Memory> Machine<M> {
...
+```rust
+impl<M: Memory> Machine<M> {
+    fn load_raw_data(&mut self, ptr : Pointer<<M as Memory>::Provenance>, ptr_ty : PtrType) -> Result<List<u8>> {
+        let PtrType::Ref { pointee, .. } = ptr_ty else {
+            throw_ub!("invalid argument to `RawEq` intrinsic: not a reference");
+        };
+        let Layout { size, align, .. } = pointee;
+        let bytes = self.mem.load(ptr, size, align, Atomicity::None)?;
+
+        if bytes.any(|byte| byte.provenance() != None) {
```
RalfJung, line 355:
```suggestion
        // FIXME: This violates provenance monotonicity.
        if bytes.any(|byte| byte.provenance() != None) {
```

spec/lang/step/intrinsics.md
```diff
@@ -342,6 +342,70 @@ impl<M: Memory> Machine<M> {
...
+        let Value::Ptr(left) = left else {
+            throw_ub!("invalid first argument to `RawEq` intrinsic: not a pointer");
+        };
+
+        let Value::Ptr(right) = right else {
+            throw_ub!("invalid second argument to `RawEq` intrinsic: not a pointer");
+        };
+
+        let Type::Ptr(l_ty) = l_ty else {
+            throw_ub!("invalid first argument to `RawEq` intrinsic: not a pointer");
```
RalfJung, line 395:
```suggestion
            throw_ub!("invalid first argument type to `RawEq` intrinsic: not a pointer");
```

spec/lang/step/intrinsics.md
```diff
@@ -342,6 +342,70 @@ impl<M: Memory> Machine<M> {
...
+        };
+
+        let Type::Ptr(r_ty) = r_ty else {
+            throw_ub!("invalid second argument to `RawEq` intrinsic: not a pointer");
+        };
+
+        let left_datas = self.load_raw_data(left, l_ty)?;
+        let right_datas = self.load_raw_data(right, r_ty)?;
```
RalfJung, line 403: "datas" -> "data"

spec/lang/step/intrinsics.md
```diff
@@ -342,6 +342,70 @@ impl<M: Memory> Machine<M> {
...
+            throw_ub!("invalid second argument to `RawEq` intrinsic: not a pointer");
+        };
+
+        let Type::Ptr(l_ty) = l_ty else {
+            throw_ub!("invalid first argument to `RawEq` intrinsic: not a pointer");
+        };
+
+        let Type::Ptr(r_ty) = r_ty else {
+            throw_ub!("invalid second argument to `RawEq` intrinsic: not a pointer");
+        };
```
RalfJung, line 400: This should not be needed since r_ty == l_ty

tooling/minitest/src/tests/raw_eq.rs
RalfJung, line 1: For new tests please use the builder API. Grep the tests for `ProgramBuilder` to find examples.

spec/lang/step/intrinsics.md
```diff
@@ -342,6 +342,70 @@ impl<M: Memory> Machine<M> {
     }
 }
 ```
+## Raw equality
+```rust
+impl<M: Memory> Machine<M> {
+    fn load_raw_data(&mut self, ptr : Pointer<<M as Memory>::Provenance>, ptr_ty : PtrType) -> Result<List<u8>> {
+        let PtrType::Ref { pointee, .. } = ptr_ty else {
```
RalfJung, line 349:
```suggestion
        // We need the pointee layout to determine how many bytes to load.
        let PtrType::Ref { pointee, .. } = ptr_ty else {
```

tooling/minitest/src/tests/raw_eq.rs
RalfJung, line 1: Please also add a test where it actually works. It should call the intrinsic twice, once where it returns `true` and once with `false`.

tooling/miniutil/src/fmt/function.rs
```diff
@@ -188,6 +188,7 @@ fn fmt_terminator(t: Terminator, comptypes: &mut Vec<CompType>) -> String {
                 IntrinsicOp::Deallocate => "deallocate",
                 IntrinsicOp::Spawn => "spawn",
                 IntrinsicOp::Join => "join",
+                IntrinsicOp::RawEq => "raweq",
```
RalfJung, line 191:
```suggestion
                IntrinsicOp::RawEq => "raw_eq",
```

RalfJung (commented):
Looks good, thanks! The comments are all minor.

tooling/minitest/src/tests/raw_eq.rs
RalfJung: This is still unresolved I think? Please use the `ProgramBuilder` API for these tests.

RalfJung: This is also still unresolved.

tooling/minitest/src/tests/raw_eq.rs
```diff
@@ -1,181 +1,214 @@
...
-    );
-    let b1 = block!(exit());
-
-    let f = function(Ret::No, 0, &locals, &[b0, b1]);
-    let p = program(&[f]);
+    let left_ptr = addr_of(left, ptr_ty);
+    let right_ptr = addr_of(right, ptr_ty);
+
+    f.raw_eq(dest, left_ptr, right_ptr);
+    f.return_();
```
RalfJung, line 84: Main should `exit`, not `return`. (It makes no difference here as the code is unreachable, but it still seems better to do this consistently.)

tooling/minitest/src/tests/raw_eq.rs
```diff
@@ -0,0 +1,214 @@
...
+    let f = p.finish_function(f);
+    let p = p.finish_program(f);
+    assert_ub(p, "invalid argument to `RawEq` intrinsic: not a reference");
+}
+
+#[test]
+fn invalid_ret_ty_raw_eq() {
+    let mut p = ProgramBuilder::new();
+    let mut f = p.declare_function();
+    let dest = f.declare_local::<i64>();
```
RalfJung, line 150:
```suggestion
    let dest = f.declare_local::<i64>(); // this is the invalid return type (should be `bool`)
```

tooling/minitest/src/tests/raw_eq.rs
```diff
@@ -0,0 +1,214 @@
...
+    assert_ub(p, "invalid return type for `RawEq` intrinsic");
+}
+
+#[test]
+fn unequal_args_ty_raw_eq() {
+    let mut p = ProgramBuilder::new();
+    let mut f = p.declare_function();
+    let dest = f.declare_local::<bool>();
+    let left = f.declare_local::<[u8; 2]>();
+    let right = f.declare_local::<[u8; 3]>();
```
RalfJung, line 178:
```suggestion
    let right = f.declare_local::<[u8; 3]>(); // not the same type as `left`
```

sun-jacobi: Fixed the above 3.

RalfJung:
Looks good, thanks. :)
Please squash the commits.

sun-jacobi:
Squashed :)

RalfJung:
Thanks! Let's get this landed. :)

PR #212, opened 2024-07-09
Modified APIs for running programs to support selecting memory models. 

sun-jacobi:
This patch modified APIs for running programs to support selecting memory models. Currently NFC, but might be useful for implementing different memory models besides the `BasicMemory`.

### DIFF
--- a/tooling/minimize/src/main.rs
+++ b/tooling/minimize/src/main.rs
@@ -53,6 +53,7 @@ pub use minirust_rs::prelude::*;
 pub use miniutil::build::{self, zst_place, TypeConv as _};
 pub use miniutil::fmt::dump_program;
 pub use miniutil::run::*;
+pub use miniutil::BasicMem;
 pub use miniutil::DefaultTarget;
 
 // Get back some `std` items
@@ -111,7 +112,7 @@ fn main() {
         if dump {
             dump_program(prog);
         } else {
-            match run_program(prog) {
+            match run_program::<BasicMem>(prog) {
                 // We can't use tcx.dcx().fatal due to <https://github.com/oli-obk/ui_test/issues/226>
                 TerminationInfo::IllFormed(err) =>
                     show_error!(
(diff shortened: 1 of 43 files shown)


spec/mem/models.md
RalfJung, line 1: I don't see why this would be an enum. The intention is that you can use everything that implements the `Memory` trait.

tooling/minitest/src/lib.rs
```diff
@@ -19,30 +19,30 @@ pub use std::string::String;
 mod tests;
 
 #[track_caller]
-pub fn assert_stop(prog: Program) {
-    assert_eq!(run_program(prog), TerminationInfo::MachineStop);
+pub fn assert_stop(prog: Program, model: MemoryModel) {
```
RalfJung, line 22: The signature I was looking for is `fn assert_stop<M: Memory>`. No need for a new enum.

To avoid having to repeat `BasicMemory<DefaultTarget>` everywhere, you can introduce `type BasicMem = BasicMemory<DefaultTarget>;` or something like that.

sun-jacobi: Thank you for the advice. I will fix it.

tooling/minimize/src/main.rs
```diff
@@ -30,6 +30,7 @@ mod rs {
     pub use rustc_target::abi::{self, call::*, Align, FieldIdx, Layout, Size};
     pub use rustc_target::abi::{FieldsShape, TagEncoding, VariantIdx, Variants};
 }
+use miniutil::BasicMem;
```
RalfJung, line 33: Please move this together with the other `miniutil` imports.

RalfJung (commented):
Looks good, just one small nit. :)

RalfJung:
Looks good! Please squash the commits.

sun-jacobi:
Squashed :)

RalfJung:
Thanks :)

PR #213, opened 2024-07-15
Initial support for Tree Borrows (without protector)

sun-jacobi:
This patch implements the Tree Borrows without Protector.


spec/mem/tree.md
```diff
@@ -0,0 +1,562 @@
...
+    fn new() -> Self {
+        Self { allocations: List::new(), tags_num: Int::ZERO, _phantom: std::marker::PhantomData }
+    }
+}
+```
+
+We track each tag's permissions on specific memory locations (bytes)
+
+```rust
+pub type Location = Address;
```
RalfJung, line 39: It seems unnecessarily confusing to have two different words for the same thing

spec/mem/tree.md
```diff
@@ -0,0 +1,562 @@
...
+            Permission::Disabled => ret(Permission::Disabled),
+        }
+    }
+
+     fn foreign_write(permission: Permission) -> Result<Permission> {
+        match permission {
+            Permission::Reserved => ret(Permission::Disabled),
+            Permission::Active => ret(Permission::Disabled), 
+            Permission::Frozen => ret(Permission::Disabled), 
+            Permission::Disabled => ret(Permission::Disabled),
```
RalfJung, line 115: This code can be simplified

spec/mem/tree.md
```diff
@@ -0,0 +1,562 @@
...
+    }
+
+    fn permission_transition(curr: Permission, access_pattern: AccessPattern, node_relation: NodeRelation) -> Result<Permission> {
+        let transition = match (access_pattern, node_relation) {
+            (AccessPattern::Read, NodeRelation::Child) => Self::child_read,
+            (AccessPattern::Write, NodeRelation::Child) => Self::child_write,
+            (AccessPattern::Read, NodeRelation::Foreign) => Self::foreign_read,
+            (AccessPattern::Write, NodeRelation::Foreign) => Self::foreign_write,
+        };
+        transition(curr)
```
RalfJung, line 126: Please avoid function pointers

spec/mem/tree.md
```diff
@@ -0,0 +1,562 @@
...
+
+pub enum Permission {
+    // Represents a two-phase borrow during its reservation phase
+    Reserved, 
+    // Represents an activated (written to) mutable reference
+    Active,
+    // Represents a shared (immutable) reference
+    Frozen,
+    // Represents a dead reference
+    Disabled,
```
RalfJung, line 54: Why does `Disabled` exist? Can't this just be "not in the map"?

spec/mem/tree.md
```diff
@@ -0,0 +1,562 @@
...
+    fn retag_ptr(&mut self, ptr: Pointer<Self::Provenance>, ptr_type: PtrType, _fn_entry: bool) -> Result<Pointer<Self::Provenance>> {
+        match ptr_type {
+            PtrType::Ref { mutbl, pointee } => {
+                let permission = match mutbl {
+                    Mutability::Mutable => Permission::Reserved,
+                    Mutability::Immutable => Permission::Frozen,
+                };
+                self.new_child(ptr, pointee.size, permission)
+            },
+            _ => ret(ptr),
```
RalfJung, line 470: `Box` should work like mutable references here.

spec/mem/tree.md
```diff
@@ -0,0 +1,562 @@
...
+        let child_permissions = Self::init_range_permissions(permission, ptr.addr, alloc_size);
+        let child_tag = self.next_tag();
+        let child_node = Node {
+            tag: child_tag,
+            parent: Some(parent_tag),
+            children: List::new(),
+            permissions: child_permissions,
+        };
+
+        // Perform one child read towards the parent node
```
RalfJung, line 258: The entire tree needs to know that there was a read, not just the parent node.

spec/mem/tree.md
```diff
@@ -0,0 +1,562 @@
...
+
+```rust
+pub struct Tree {
+    root_tag: BorTag, 
+    /// Map a borrow tag to a node
+    nodes: Map<BorTag, Node>,
+}
+
+pub struct Node {
+    tag: BorTag,
```
RalfJung, line 141: This is technically redundant, can you try removing it?

spec/mem/tree.md
```diff
@@ -0,0 +1,562 @@
...
+    }
+}
+```
+
+We track each tag's permissions on specific memory locations (bytes)
+
+```rust
+pub type Location = Address;
+
+pub enum AccessPattern {
```
RalfJung, line 41: Maybe `AccessKind`?

spec/mem/tree.md
```diff
@@ -0,0 +1,562 @@
...
+
+    /// Create a new node for a pointer (reborrow)
+    fn new_child(
+        &mut self, 
+        ptr: Pointer<TreeBorrowsProvenance>,
+        alloc_size: Size, 
+        permission: Permission
+    ) -> Result<Pointer<TreeBorrowsProvenance>> {
+        let Some((parent_tag, alloc_id)) = ptr.provenance else {
+            throw_ub!("Tree Borrows: Pointer does not contain provenance");
```
RalfJung, line 241: If the pointee is zero, then this should not be UB.

spec/mem/tree.md
```diff
@@ -0,0 +1,562 @@
...
+        };
+
+        let Some(mut parent_node) = self.allocations[alloc_id.0].tree.nodes.get(parent_tag) else {
+            throw_ub!("Tree Borrows: Parent pointer does not exist in the tree");
```
RalfJung, line 245: This case should be impossible, so it can panic.

spec/mem/tree.md
```diff
@@ -0,0 +1,562 @@
...
+        };
+
+        // Create the new child node
+        let child_permissions = Self::init_range_permissions(permission, ptr.addr, alloc_size);
```
RalfJung, line 249: This need adjusting for lazy init: the entire range of the allocation should get `permission` here, not just the pointee size. But then there's a read access next and that is only done for the pointee size.

spec/mem/tree.md
```diff
@@ -0,0 +1,562 @@
...
+```rust
+pub type Location = Address;
+
+pub enum AccessPattern {
+    Read, 
+    Write,
+}
+
+pub enum Permission {
+    // Represents a two-phase borrow during its reservation phase
```
RalfJung, line 47: These should also all be doc comments

sun-jacobi:
Rebased since #212 is merged.

spec/mem/tree.md
```diff
@@ -0,0 +1,561 @@
...
+            child_permissions.insert(location, permission);
+        }
+        child_permissions
+    }
+
+    /// Create a new node for a pointer (reborrow)
+    fn new_child(
+        &mut self, 
+        ptr: Pointer<TreeBorrowsProvenance>,
+        alloc_size: Size,
```
sun-jacobi, line 237: Maybe `alloc_size` -> `pointee_size`

spec/mem/tree.md
```diff
@@ -0,0 +1,564 @@
...
+
+            let next_permission = Self::permission_transition(curr_permission, access_pattern, node_relation)?;
+            node.permissions.insert(location, next_permission);
+        }
+
+        // Insert back the updated node
+        self.tree.nodes.insert(curr, node);
+
+        // For each descendant, the base node is a foreign.
+        let new_relation = if curr == base { NodeRelation::Foreign } else { node_relation };
```
sun-jacobi, line 198: This is actually incorrect. It can not differ strict transitive parents of `base` and nodes who do not share a branch with `base`

spec/mem/tree.md
RalfJung, line 1: Please find a way to split this in separate files. For instance, the tree and tree traversal code could be split from the rest pretty well, I think.

spec/mem/tree.md
```diff
@@ -0,0 +1,564 @@
+# MiniRust Tree Borrows
+
+To check the details on Tree Borrows
+
+1. [From Stacks to Trees: A new aliasing model for Rust](https://www.ralfj.de/blog/2023/06/02/tree-borrows.html)
+1. [Neven's posts on Tree Borrows](https://perso.crans.org/vanille/treebor)
+
+## Data structures
+
+In addition to the `BasicMemory`, `TreeBorrowsMemory` also track an ID for each pointer, called a *Borrow Tag*.
```
RalfJung, line 10:
```suggestion
In addition to the `BasicMemory`, `TreeBorrowsMemory` also tracks an ID for each pointer, called a *Borrow Tag*.
```

spec/mem/tree.md
```diff
@@ -0,0 +1,564 @@
...
+```rust
+pub struct BorTag(Int);
+pub type TreeBorrowsProvenance = (BorTag, AllocId);
+```
+
+```rust
+pub struct TreeBorrowsMemory<T: Target> {
+    allocations: List<TreeBorrowsAllocation>,
+    /// Next unused borrow tag.
+    next_tag: Int,
```
RalfJung, line 21: Why isn't this a `BorTag`?

spec/mem/tree.md
```diff
@@ -0,0 +1,564 @@
...
+
+     fn foreign_write(permission: Permission) -> Result<Permission> {
+        ret(
+            match permission {
+                Permission::Reserved => Permission::Disabled,
+                Permission::Active => Permission::Disabled,
+                Permission::Frozen => Permission::Disabled,
+                Permission::Disabled => Permission::Disabled,
+            }
+        )
```
RalfJung, line 118:
```suggestion
        ret(Permission::Disabled)
```

spec/mem/tree.md
```diff
@@ -0,0 +1,564 @@
...
+    /// `base` means the node that is actually memory accessed.
+    /// This method will throw UBs, representing the undefined behavior captured by Tree Borrows.
+    fn node_access(
+        &mut self, 
+        curr: BorTag,
+        access_pattern: AccessKind,
+        ptr: Pointer<TreeBorrowsProvenance>,
+        size: Size,
+        base: BorTag,
+    ) -> Result<bool> {
```
RalfJung, line 177: What is the return type for? The doc comment should explain that.

spec/mem/tree.md
```diff
@@ -0,0 +1,564 @@
...
+
+```rust
+impl TreeBorrowsAllocation {
+    /// Recusively do permission transition on the `curr` node and all its descendants.
+    /// `base` means the node that is actually memory accessed.
+    /// This method will throw UBs, representing the undefined behavior captured by Tree Borrows.
+    fn node_access(
+        &mut self, 
+        curr: BorTag,
+        access_pattern: AccessKind,
```
RalfJung, line 173:
```suggestion
        access_kind: AccessKind,
```

spec/mem/tree.md
```diff
@@ -0,0 +1,564 @@
...
+        let Some(mut node) = self.tree.nodes.get(curr) else { throw_ub!("Tree Borrows: node not existed"); };
+
+        // Flag to indicate whether `base`is a child of `curr`
+        let mut is_child = curr == base;
+
+        for child_tag in node.children {
+            is_child |= self.node_access(child_tag, access_pattern, ptr, size, base)?;
+        }
+
+        let node_relation = if is_child { NodeRelation::Child } else { NodeRelation::Foreign };
```
RalfJung, line 187: A comment should explain the direction of this relation -- i.e., who's the "self" here. It seems to be "curr"? I.e., "child" means that `base` is a child of `curr`.

spec/mem/tree.md
```diff
@@ -0,0 +1,564 @@
...
+    /// Recusively do permission transition on the `curr` node and all its descendants.
+    /// `base` means the node that is actually memory accessed.
+    /// This method will throw UBs, representing the undefined behavior captured by Tree Borrows.
+    fn node_access(
+        &mut self, 
+        curr: BorTag,
+        access_pattern: AccessKind,
+        ptr: Pointer<TreeBorrowsProvenance>,
+        size: Size,
+        base: BorTag,
```
RalfJung, line 176: Maybe call this "access_tag" or so, `base` is not very self-explaining.

"curr" would then become "curr_tag".

spec/mem/tree.md
```diff
@@ -0,0 +1,564 @@
...
+    /// The address where this allocation starts.
+    /// This is never 0, and `addr + data.len()` fits into a `usize`.
+    addr: Address,
+    /// The alignment that was requested for this allocation.
+    /// `addr` will be a multiple of this.
+    align: Align,
+    /// The kind of this allocation.
+    kind: AllocationKind,
+    /// Whether this allocation is still live.
+    live: bool,
```
RalfJung, line 71: Instead of repeating all the fields of a basic `Allocation`, would it make sense to just re-use the basic `Allocation` type here? And then also re-use some of the methods from basic.md for memory access etc?

tooling/minimize/tests/pass/tree_borrows/disjoint-reborrow.rs
```diff
@@ -0,0 +1,13 @@
...
+fn main() {
+    let parent = &mut 0u8; // (parent, Reserved)   
+    let x = &mut *parent; // (parent, Reserved) -> (x, Reserved)
+    let y = &mut *x; // (parent, Reserved) -> (x, Reserved) -> (y, Reserved)
+       
+    *y = 42; // (parent, Reserved) -> (x, Reserved) -> (y, Active)
+    *x = 31; // (parent, Reserved) -> (x, Active) -> (y, Disable)
+    
+    print(*x);
+}
```
RalfJung, line 13: There's a missing newline at the end of this file.

tooling/minimize/src/main.rs
```diff
@@ -143,6 +146,14 @@ fn split_args(args: Args) -> (Vec<String>, Vec<String>) {
     (minimize_args, rustc_args)
 }
 
+fn get_run_program(args: Vec<String>) -> fn(miniutil::Program) -> miniutil::TerminationInfo {
```
RalfJung, line 149: You really like function pointers, don't you? ;)
But I don't think making this higher-order is helpful. You can just do the case distinction later -- i.e., make this function actually run the program, according to the flags.

spec/mem/tree.md
RalfJung: Actually this will have to change -- `Reserved` needs to track whether there is interior mutability, and then allow foreign writes. Also see https://github.com/rust-lang/miri/pull/3754 and [this test](https://github.com/rust-lang/miri/blob/05640c48b11508a33b471d5e00f0f8072d8dba58/tests/pass/tree_borrows/2phase-interiormut.rs). You should also add [all of these tests](https://github.com/rust-lang/miri/blob/6680b2fa09496b60c40c6ce09449f46efbf253d5/tests/pass/tree_borrows/tree-borrows.rs) (they might need some adjusting to work in minirust).

spec/mem/tree.md
sun-jacobi: This is fixed by the above
```
// For zero-sized accesses, there is nothing to do.
if pointee_size.is_zero() {
    return ret(ptr);
}
```

spec/mem/tree.md
sun-jacobi: To support this, I think we may need to modify the interface of `retag_ptr` to add an argument like `is_freeze`. Then implement a function like `is_trivially_freeze` in the https://doc.rust-lang.org/nightly/nightly-rustc/src/rustc_middle/ty/util.rs.html#1271. Do you think this makes sense? @RalfJung

spec/mem/tree.md
JoJoDeveloping: Yes, please document this everywhere, and have one central place where you put lots of examples. In our Coq development, this was not documented, and we did it the wrong way around an _embarrasing_ number of times.

JoJoDeveloping (commented):
Why would you remove `Disabled`? This change spreads the logic for when accesses are UB around, it forces you to add `Option` everywhere, and it does not gain anything. Also, conceptually, `Disabled` very much is a state. It represents the absence of permissions and is the terminal state, but it _is a state_. So using `Option` is wrong conceptually I would argue.

Note that `foreign_write` does not always return `Disabled` in the extended model, so you will have to re-add it eventually.

RalfJung:
I told him to remove `Disabled`. It gains something, namely not having two different but equivalent ways to represent the same state.

spec/mem/tree.md
RalfJung: Hm... yeah we will need more information. Not just in `retag_ptr` but in `PtrType::Ref`.

sun-jacobi: Thank you for the advice. I will send another small PR for this :)

sun-jacobi:
@JoJoDeveloping Thank you for the comment:) I also think keeping the Disabled state might be conceptually easier.

spec/mem/tree.md
sun-jacobi: AFAIU, `UnsafeCell` is not currently supported. Maybe we should consider this after we support UnsafeCell?

spec/mem/tree_borrows/automata.md
```diff
@@ -0,0 +1,84 @@
...
+        let Some(mut node) = self.tree.nodes.get(tag) else { throw_ub!("Tree Borrows: node not existed"); };
+
+        let Some(curr_permission) = node.permissions.get(addr) else {
+            if node_relation == NodeRelation::Foreign {
+                return ret(());
+            }
+            match access_kind {
+                AccessKind::Read => throw_ub!("Tree Borrows: Child reading a pointer without permission"),
+                AccessKind::Write => throw_ub!("Tree Borrows: Child writing a pointer without permission"),
+            }
```
RalfJung, line 64: Hm, I can see how having this here isn't that great.

What if you change the argument type for `foreign_read` etc to `Option<Permission>` as well? Then this can be handled as before.

spec/mem/tree_borrows/automata.md
```diff
@@ -0,0 +1,84 @@
...
+        };
+
+        let next_permission = match (node_relation, access_kind) {
+            (NodeRelation::Foreign, AccessKind::Write) => None,
```
RalfJung, line 68: Please still call a `Self::foreign_write` function here, even if it always returns `None` for now.

spec/mem/tree_borrows/model.md
```diff
@@ -0,0 +1,371 @@
...
+# Memory Operations
+Then we implement the memory model interface for the Tree Borrow.
+
+### Allocate and Deallocate
+
+We create a new tree for one allocation
+
+```rust
+impl<T: Target> Memory for TreeBorrowsMemory<T> {
+    fn allocate(&mut self, kind: AllocationKind, size: Size, align: Align) -> NdResult<Pointer<Self::Provenance>>  {
```
RalfJung, line 127: This is mostly the same as what the basic memory does, right? Is there a way to share that?

The same goes for `read`, `write`, `deallocate`, and `check_ptr`.

JoJoDeveloping:
> namely not having two different but equivalent ways to represent the same state.

I don't think that `None` is ever going to appear otherwise? Since the state is set for all valid offsets within the allocation (i.e. `0..self.size()`), it's never gonna be accessed out-of-bounds, since this would already be UB before (since an OOB access should be identified earlier as it's in a different category compared to "expired pointer accesses").

RalfJung:
`None` in Miri means "has not yet been accessed, on first access please lazily put in the initial permission here". So there are two equivalent states: the permission being `None`, and the permission being whatever the initial permission for this tag is (with `initialized: false`).

The scheme used in this PR avoids that. It entirely avoids even having a notion of "initial permission" in the `struct Node`.

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,97 @@
...
+    nodes: Map<BorTag, Node>,
+}
+
+pub struct Node {
+    /// Borrow tag of the parent node
+    parent: Option<BorTag>, 
+    /// Borrow tags of the children node
+    children: List<BorTag>,
+    /// Permission for each location
+    permissions: Map<Address, Permission>,
```
RalfJung, line 18: This can't just be `Permission`, it has to be something like `(Accessed, Permission)` where
```rust
enum Accessed {
    /// This address has been accessed (read, written, or the initial implicit read upon retag)
    /// with this borrow tag.
    Yes,
    /// This address has not yet been accessed with this borrow tag. We still track how foreign
    /// accesses affect the current permission so that on the first access, we start in the right state.
    No,
}
```

JoJoDeveloping:
> `None` in Miri means "has not yet been accessed, on first access please lazily put in the initial permission here". So there are two equivalent states: the permission being `None`, and the permission being whatever the initial permission for this tag is (with `initialized: false`).
> 
> The scheme used in this PR avoids that. It entirely avoids even having a notion of "initial permission" in the `struct Node`.

Then I do not understand what you mean by 

> It gains something, namely not having two different but equivalent ways to represent the same state.
 
What are the two different but equivalent ways?

spec/mem/tree_borrows/tree.md
sun-jacobi: Could you please explain why this enum is necessary?

RalfJung:
@johannes The comment you replied two explained exactly that: the value not being in the map is equivalent to the value being in the map with the tag's initial permission and "initialized: false". Two different states, but equivalent.

RalfJung:
Lol, I was replying in my phone via mail so I didn't have auto-complete... sorry for the wrong ping.^^

spec/mem/tree_borrows/model.md
RalfJung, line 1: Maybe rename this file to "memory"?

spec/mem/tree_borrows/automata.md
```diff
@@ -0,0 +1,93 @@
+# State Machine For the Tree Borrows
+
+The core of the Tree Borrows is a state machine. We track a state machine for each node on the tree. We call the state as *Permission*.
```
RalfJung, line 3:
```suggestion
The core of Tree Borrows is a state machine for each node and each location.
We call the state a *Permission*.
```

tooling/minimize/tests/pass/tree_borrows.rs
```diff
@@ -0,0 +1,248 @@
+include!("../helper/transmute.rs");
+use std::ptr;
+
+//@ compile-flags: --minimize-tree-borrows
```
RalfJung, line 4: Please put flags like that at the top of the file

tooling/minimize/tests/ub/tree_borrows/array_access.rs
```diff
@@ -0,0 +1,20 @@
...
+
+//@ compile-flags: --minimize-tree-borrows
+fn main() {
+    unsafe {
+        let mut data = [0u8, 1, 2, 3];
+        let x = &mut data[0] as *mut u8; // (x, [R, R, R, R])
+        let y = &mut *x; // (x, [R, R, R, R]) -> (y, [R, R, R, R])
+        let yraw = y as *mut u8;
+
+        *x.add(1) = 42; // (x, [R, A, R, R]) -> (y, [R, D, R, R])
```
RalfJung, line 12:
```suggestion
        // Using x at offset 1 invalidates y.
        *x.add(1) = 42; // (x, [R, A, R, R]) -> (y, [R, D, R, R])
```

tooling/minimize/tests/ub/tree_borrows/array_access.rs
```diff
@@ -0,0 +1,20 @@
...
+
+        print(*yraw.add(0));
```
RalfJung, line 14:
```suggestion
        // y can still be used for all the other offsets
        print(*yraw.add(0));
```
In fact it could even be used to write, right? Maybe make these writes instead of reads.

tooling/minimize/tests/ub/tree_borrows/array_access.rs
```diff
@@ -0,0 +1,20 @@
...
+        print(*yraw.add(2));
+        print(*yraw.add(3));
+
+        *yraw.add(1) = 57;
```
RalfJung, line 18:
```suggestion
        // UB! The write to x has invalidated its child, y.
        *yraw.add(1) = 57;
```

tooling/minimize/tests/ub/tree_borrows/overlapping_read_write.rs
```diff
@@ -0,0 +1,12 @@
+extern crate intrinsics;
+use intrinsics::*;
+
+//@ compile-flags: --minimize-tree-borrows
+fn main() {
+    let parent = &mut 31; // (parent, Reserved)   
+    let x = parent as *mut i32; // (parent, Reserved)
```
RalfJung, line 7: Maybe "x" and "xraw" to make it more clear that these are the same?

tooling/minimize/tests/ub/tree_borrows/overlapping_read_write.rs
```diff
@@ -0,0 +1,12 @@
+extern crate intrinsics;
+use intrinsics::*;
+
+//@ compile-flags: --minimize-tree-borrows
+fn main() {
+    let parent = &mut 31; // (parent, Reserved)   
+    let x = parent as *mut i32; // (parent, Reserved) 
+    let y = unsafe { &mut *x }; // (parent, Reserved) -> (y, Reserved)
+    *y = 42;
```
RalfJung, line 9: This is the activation write, right? There should be a comment.

tooling/minimize/tests/ub/tree_borrows/overlapping_write_write.rs
```diff
@@ -0,0 +1,17 @@
...
+//@ compile-flags: --minimize-tree-borrows
+fn main() {
+    let parent = &mut 31; // (parent, Reserved)   
+    let x = parent as *mut i32; // (parent, Reserved) 
+    let y = unsafe { &mut *x }; // (parent, Reserved) -> (y, Reserved)
+    *y = 42;
+
+    unsafe { 
+        *x = 57; // (parent, Reserved) -> (y, Disabled)
+        print(*x); // (parent, Reserved) -> (y, Disabled)
```
RalfJung, line 13:
```suggestion
        assert!(*x == 57); // (parent, Reserved) -> (y, Disabled)
```
This is generally nicer than printing.

tooling/minimize/tests/ub/tree_borrows/overlapping_write_write.rs
```diff
@@ -0,0 +1,17 @@
...
+    } 
+
+    *y = 31; // UB! Child Write to Disabled 
+}
```
RalfJung, line 17: missing newline (maybe look into your editor settings to see whether you can make it automatically add trailing newlines)

tooling/minimize/tests/pass/tree_borrows.rs
RalfJung, line 1: Would be good to have a test here that creates two mutable references into the same array, `&mut arr[0]` and `&mut arr[1]`, and ensures we can use them both as long as the accesses don't overlap (and when they overlap, we can still do reads).

RalfJung, line 1: Please add a comment saying that this is copied from Miri.

Also, please also copy `tests/pass/tree_borrows/sb_fails.rs` from Miri.

tooling/minimize/tests/ub/tree_borrows/overlapping_write_write.rs
```diff
@@ -0,0 +1,17 @@
...
+    let x = parent as *mut i32; // (parent, Reserved) 
+    let y = unsafe { &mut *x }; // (parent, Reserved) -> (y, Reserved)
+    *y = 42;
+
+    unsafe { 
+        *x = 57; // (parent, Reserved) -> (y, Disabled)
+        print(*x); // (parent, Reserved) -> (y, Disabled)
+    } 
+
+    *y = 31; // UB! Child Write to Disabled
```
RalfJung, line 16: Here this could in fact just do a read from y, even that should be UB. So add a third test for that, "overlapping_write_read".

Also, the tests are about mutable references. So maybe rename them to `mutable_ref_conflicting_read_write` etc?

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,103 @@
...
+    nodes: Map<BorTag, Node>,
+}
+
+pub struct Node {
+    /// Borrow tag of the parent node
+    parent: Option<BorTag>, 
+    /// Borrow tags of the children node
+    children: List<BorTag>,
+    /// Permission for each location
+    permissions: List<(Accessed, Permission)>,
```
sun-jacobi, line 18: A name like `location_states` would be more appropriate here.

sun-jacobi:
Refactored some operations in `BasicMemory` to `Allocation`'s methods.

RalfJung:
Closing in favor of https://github.com/minirust/minirust/pull/220.

PR #217, opened 2024-07-18
[minimize] Enable emitting MIR Retag

sun-jacobi:
Current test cases in `tooling/minimize/tests` are running without MIR `Retag`, thus there is no `Statement::Validate` generated. So I think it might be helpful to enable this in `minimize` by default?


RalfJung:
yes, this is a good point, we should do this. :)  Thanks!

PR #218, opened 2024-07-22
Add support for unsized types

Indeximal:
Hi, this is my currently best idea to change the types and function signatures to allow for unsized types.

Of course it is very incomplete, specr-transpile only gives 74 errors xD.

I have drawn the abstraction of thin/fat pointers between the `typed_load` and `ConcurrentMemory::load` functions.

### DIFF
--- a/spec/lang/representation.md
+++ b/spec/lang/representation.md
@@ -4,6 +4,7 @@ The main purpose of [types](types.md) is to define how [values](values.md) are (
 This is the *[representation relation]*, which is defined in the following.
 `decode` converts a list of bytes into a value; this operation can fail if the byte list is not a valid encoding for the given type.
 `encode` inverts `decode`; it will always work when the value is [well-formed][well-formed-value] for the given type (which the specification must ensure, i.e. violating this property is a spec bug).
+Unsized types cannot be represented as values and thus a call to `encode` or `decode` in a well formed program is a spec bug.
 
 [representation relation]: https://github.com/rust-lang/unsafe-code-guidelines/blob/master/reference/src/glossary.md#representation-relation
 [well-formed-value]: well-formed.md#well-formed-values
@@ -93,7 +94,7 @@ This is required to achieve a "monotonicity" with respect to provenance (as disc
 ### Pointers
 
 ```rust
-fn decode_ptr<M: Memory>(bytes: List<AbstractByte<M::Provenance>>) -> Option<Pointer<M::Provenance>> {
+fn decode_ptr<M: Memory>(bytes: List<AbstractByte<M::Provenance>>) -> Option<ThinPointer<M::Provenance>> {
     if bytes.len() != M::T::PTR_SIZE.bytes() { throw!(); }
     // Convert into list of bytes; fail if any byte is uninitialized.
     let bytes_data = bytes.try_map(|b| b.data())?;
@@ -105,35 +106,38 @@ fn decode_ptr<M: Memory>(bytes: List<AbstractByte<M::Provenance>>) -> Option<Poi
             provenance = None;
         }
     }
-    ret(Pointer { addr, provenance })
+    ret(ThinPointer { addr, provenance })
 }
 
-fn encode_ptr<M: Memory>(ptr: Pointer<M::Provenance>) -> List<AbstractByte<M::Provenance>> {
-    let bytes_data = M::T::ENDIANNESS.encode(Unsigned, M::T::PTR_SIZE, ptr.data_pointer.addr).unwrap();
-    let bytes_data = bytes_data.map(|b| AbstractByte::Init(b, ptr.data_pointer.provenance));
-    let bytes_meta = match ptr.metadata {
+fn encode_ptr<M: Memory>(ptr: ThinPointer<M::Provenance>) -> List<AbstractByte<M::Provenance>> {
+    let bytes_data = M::T::ENDIANNESS.encode(Unsigned, M::T::PTR_SIZE, ptr.addr).unwrap();
+    bytes_data.map(|b| AbstractByte::Init(b, ptr.provenance))
+    
+}
+
+fn encode_ptr_meta<M: Memory>(meta: Option<PointerMeta>) -> List<AbstractByte<M::Provenance>> {
+    match meta {
         None => list!(),
         Some(PointerMeta::ElementCount(len)) => {
             let bytes = M::T::ENDIANNESS.encode(Unsigned, M::T::PTR_SIZE, len).unwrap();
             bytes.map(|b| AbstractByte::Init(b, None))
         }
-        Some(PointerMeta::VTable) => unimplemented!("Trait objects are not supported yet"),
-    };
-    // Is the order correct and target independent?
-    bytes_data.iter().chain(bytes_meta.iter()).collect()
+    }
 }
 
 impl Type {
     fn decode<M: Memory>(Type::Ptr(ptr_type): Self, bytes: List<AbstractByte<M::Provenance>>) -> Option<Value<M>> {
-        let ptr = decode_ptr::<M>(bytes)?;
+        if bytes.len() < M::T::PTR_SIZE.bytes() { throw!(); }
+        let ptr = decode_ptr::<M>(bytes.subslice_with_length(Int::ZERO, M::T::PTR_SIZE.bytes()))?;
         if !ptr_type.addr_valid(ptr.addr) {
             return None;
         }
-        ret(Value::Ptr(ptr))
+        // TODO: Decode wide pointers, but needs slight changes in PtrType & Layout
+        ret(Value::Ptr(ptr.widen(None)))
     }
     fn encode<M: Memory>(Type::Ptr(_): Self, val: Value<M>) -> List<AbstractByte<M::Provenance>> {
         let Value::Ptr(ptr) = val else { panic!() };
-        encode_ptr::<M>(ptr)
+        encode_ptr::<M>(ptr.thin_pointer).iter().chain(encode_ptr_meta::<M>(ptr.metadata).iter()).collect()
     }
 }
 ```
@@ -381,8 +385,11 @@ Similarly, on `Pointer` we say that adding provenance makes it more defined:
 ```rust
 impl<Provenance> DefinedRelation for Pointer<Provenance> {
     fn le_defined(self, other: Self) -> bool {
-        self.addr == other.addr &&
-            match (self.provenance, other.provenance) {
+        self.thin_pointer.addr == other.thin_pointer.addr &&
+        // Feedback wanted: Using `Option<PointerMeta> as DefinedRelation` would be okay aswell, 
+        // since all antecedents garantee that either both or none have metadata in the first place?
+        self.metadata == other.metadata && 
+            match (self.thin_pointer.provenance, other.thin_pointer.provenance) {
                 (None, _) => true,
                 (Some(prov1), Some(prov2)) => prov1 == prov2,
                 _ => false,
@@ -477,7 +484,7 @@ We also use this to lift retagging from pointers to compound values.
 
 ```rust
 impl<M: Memory> ConcurrentMemory<M> {
-    fn typed_store(&mut self, ptr: Pointer<M::Provenance>, val: Value<M>, ty: Type, align: Align, atomicity: Atomicity) -> Result {
+    fn typed_store(&mut self, ptr: ThinPointer<M::Provenance>, val: Value<M>, ty: Type, align: Align, atomicity: Atomicity) -> Result {
         assert!(val.check_wf(ty).is_ok(), "trying to store {val:?} which is ill-formed for {:#?}", ty);
         let bytes = ty.encode::<M>(val);
         self.store(ptr, bytes, align, atomicity)?;
@@ -485,8 +492,8 @@ impl<M: Memory> ConcurrentMemory<M> {
         ret(())
     }
 
-    fn typed_load(&mut self, ptr: Pointer<M::Provenance>, ty: Type, align: Align, atomicity: Atomicity) -> Result<Value<M>> {
-        let bytes = self.load(ptr.data_pointer, ty.size::<M::T>().resolve(ptr.metadata), align, atomicity)?;
+    fn typed_load(&mut self, ptr: ThinPointer<M::Provenance>, ty: Type, align: Align, atomicity: Atomicity) -> Result<Value<M>> {
+        let bytes = self.load(ptr, ty.size::<M::T>(), align, atomicity)?;
         ret(match ty.decode::<M>(bytes) {
             Some(val) => {
                 assert!(val.check_wf(ty).is_ok(), "decode returned {val:?} which is ill-formed for {:#?}", ty);
--- a/spec/lang/representation.md
+++ b/spec/lang/representation.md
@@ -4,7 +4,7 @@ The main purpose of [types](types.md) is to define how [values](values.md) are (
 This is the *[representation relation]*, which is defined in the following.
 `decode` converts a list of bytes into a value; this operation can fail if the byte list is not a valid encoding for the given type.
 `encode` inverts `decode`; it will always work when the value is [well-formed][well-formed-value] for the given type (which the specification must ensure, i.e. violating this property is a spec bug).
-Unsized types cannot be represented as values and thus a call to `encode` or `decode` in a well formed program is a spec bug.
+Unsized types cannot be represented as values and thus a call to `encode` or `decode` for such types can never occur in a well-formed program.
 
 [representation relation]: https://github.com/rust-lang/unsafe-code-guidelines/blob/master/reference/src/glossary.md#representation-relation
 [well-formed-value]: well-formed.md#well-formed-values
@@ -20,7 +20,7 @@ impl Type {
     /// Decode a list of bytes into a value. This can fail, which typically means Undefined Behavior.
     /// `decode` must satisfy the following property:
     /// ```
-    /// ty.decode(bytes) = Some(_) -> bytes.len() == ty.size() && ty.inhabited()`
+    /// ty.decode(bytes) = Some(_) -> bytes.len() == ty.size().unwrap_size() && ty.inhabited()`
     /// ```
     /// In other words, all valid low-level representations must have the length given by the size of the type,
     /// and the existence of a valid low-level representation implies that the type is inhabited.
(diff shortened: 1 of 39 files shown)


spec/lang/types.md
```diff
@@ -174,6 +175,41 @@ impl Type {
...
+
+    /// The size of the type is given by `min_size + element_size * len`,
+    /// where `len` is found in the wide pointer metadata.
+    FixPlusTail {
+        min_size: Size,
+        element_size: Size,
+    },
+
+    /// TODO
+    VTable,
```
RalfJung, line 191: For now let's just not have these

spec/lang/types.md
```diff
@@ -174,6 +175,41 @@ impl Type {
...
     }
 }
+
+pub enum SizeStrategy {
+    /// The type is statically `Sized`.
+    Sized(Size),
+
+    /// The size of the type is given by `min_size + element_size * len`,
+    /// where `len` is found in the wide pointer metadata.
+    FixPlusTail {
```
RalfJung, line 185: `SliceTail`?

spec/mem/pointer.md
```diff
@@ -20,25 +20,49 @@ As far as the interface is concerned, this is some opaque extra data that we car
 /// of the address space.
 pub type Address = Int;
 
-/// A "pointer" is an address together with its Provenance.
+/// A "data pointer" is an address together with its Provenance.
 /// Provenance can be absent; those pointers are
 /// invalid for all non-zero-sized accesses.
-pub struct Pointer<Provenance> {
+pub struct DataPointer<Provenance> {
```
RalfJung, line 26: `ThinPointer`

spec/mem/pointer.md
```diff
@@ -20,25 +20,49 @@ As far as the interface is concerned, this is some opaque extra data that we car
...
+// <https://doc.rust-lang.org/std/primitive.pointer.html>.
+// An alternative might be calling it `MemPointer`.
+// I believe it is more consistent to call the whole thing `Pointer`, rather than just the data part,
+// as this is how most of Rusts references are written (ignoring any metadata fields).
+
+/// A "pointer" is the data pointer with optionally some metadata.
+/// Corresponds to the rust primitive "pointer", as well as references and boxes.
+pub struct Pointer<Provenance> {
+    pub data_pointer: DataPointer<Provenance>,
+    pub metadata: Option<PointerMeta>
```
RalfJung, line 41:
```suggestion
    pub metadata: Option<PointerMeta>,
```

spec/mem/pointer.md
```diff
@@ -20,25 +20,49 @@ As far as the interface is concerned, this is some opaque extra data that we car
...
-        Pointer { addr, ..self }
+        DataPointer { addr, ..self }
     }
 }
 ```
 
+```rust
+// This doesn't make a lot of sense in this file, maybe rather values.md.
+// (Similarly the PtrType is only used for retagging in mem.rs, and it felt out of place when reading ?)
+/// The metadata that can be stored in a fat pointer.
```
RalfJung, line 58:
```suggestion
/// The metadata that can be stored in a wide pointer.
```

Indeximal:
At this revision, MiniRust compiles again and the tests pass (which I didn't expect, but I guess the Pointer Type truely is just internal). However the sizestrategy has been removed and I'll work on in a next commit.

spec/mem/pointer.md
```diff
@@ -6,40 +6,62 @@ This becomes even more prominent with aliasing models such as [Stacked Borrows].
 The memory model hence takes the stance that a pointer consists of the *address* (which truly is just an integer of appropriate size) and a *provenance*.
 What exactly [provenance] *is* is up to the memory model.
 As far as the interface is concerned, this is some opaque extra data that we carry around with our pointers and that places restrictions on which pointers may be used to do what when.
+Additionally, in rust the term *pointer* is used for the primitive that might contain metadata.
+We therefore use a separate *thin pointer* type when metadata is not needed.
```
RalfJung, line 10:
```suggestion

On top of this basic concept of a pointer, Rust also knows pointers with metadata (such as `*const [i32]`).
We therefore use the term *thin pointer* for what has been described above, and *pointer* for a pointer that optionally carries some metadata.
```

spec/mem/pointer.md
```diff
@@ -6,40 +6,62 @@ This becomes even more prominent with aliasing models such as [Stacked Borrows].
...
 /// invalid for all non-zero-sized accesses.
-pub struct Pointer<Provenance> {
+pub struct ThinPointer<Provenance> {
     pub addr: Address,
     pub provenance: Option<Provenance>,
 }
 
-impl<Provenance> Pointer<Provenance> {
+/// A "pointer" is the thin pointer with optionally some metadata, making it a wide pointer.
+/// Corresponds to the rust primitive "pointer", as well as references and boxes.
```
RalfJung, line 35:
```suggestion
/// This corresponds to the Rust raw pointer types, as well as references and boxes.
```

spec/mem/pointer.md
```diff
@@ -6,40 +6,62 @@ This becomes even more prominent with aliasing models such as [Stacked Borrows].
...
+Additionally, in rust the term *pointer* is used for the primitive that might contain metadata.
+We therefore use a separate *thin pointer* type when metadata is not needed. 
 
 [pointers-complicated]: https://www.ralfj.de/blog/2018/07/24/pointers-and-bytes.html
 [pointers-complicated-2]: https://www.ralfj.de/blog/2020/12/14/provenance.html
 [pointers-complicated-3]: https://www.ralfj.de/blog/2022/04/11/provenance-exposed.html
 [provenance]: https://github.com/rust-lang/unsafe-code-guidelines/blob/master/reference/src/glossary.md#pointer-provenance
 [Stacked Borrows]: https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md
 
+## Pointer Types
```
RalfJung, line 18:
```suggestion
## Pointer Types

```

spec/mem/pointer.md
```diff
@@ -6,40 +6,62 @@ This becomes even more prominent with aliasing models such as [Stacked Borrows].
...
+        Pointer {
+            thin_pointer: self,
+            metadata,
+        }
     }
 }
 ```
 
-
+## Layout
```
RalfJung, line 64:
```suggestion
## Layout

```

spec/lang/representation.md
```diff
@@ -4,6 +4,7 @@ The main purpose of [types](types.md) is to define how [values](values.md) are (
 This is the *[representation relation]*, which is defined in the following.
 `decode` converts a list of bytes into a value; this operation can fail if the byte list is not a valid encoding for the given type.
 `encode` inverts `decode`; it will always work when the value is [well-formed][well-formed-value] for the given type (which the specification must ensure, i.e. violating this property is a spec bug).
+Unsized types cannot be represented as values and thus a call to `encode` or `decode` in a well formed program is a spec bug.
```
RalfJung, line 7:
```suggestion
Unsized types cannot be represented as values and thus a call to `encode` or `decode` for such types can never occur in a well-formed program.
```

spec/lang/representation.md
```diff
@@ -105,25 +106,38 @@ fn decode_ptr<M: Memory>(bytes: List<AbstractByte<M::Provenance>>) -> Option<Poi
...
             return None;
         }
-        ret(Value::Ptr(ptr))
+        // TODO: Decode wide pointers, but needs slight changes in PtrType & Layout
+        ret(Value::Ptr(ptr.widen(None)))
     }
     fn encode<M: Memory>(Type::Ptr(_): Self, val: Value<M>) -> List<AbstractByte<M::Provenance>> {
         let Value::Ptr(ptr) = val else { panic!() };
-        encode_ptr::<M>(ptr)
+        encode_ptr::<M>(ptr.thin_pointer).iter().chain(encode_ptr_meta::<M>(ptr.metadata).iter()).collect()
```
RalfJung, line 140: Please let-bind the two components, so that it's not one line doing too many things.

spec/lang/representation.md
```diff
@@ -371,8 +385,11 @@ Similarly, on `Pointer` we say that adding provenance makes it more defined:
 ```rust
 impl<Provenance> DefinedRelation for Pointer<Provenance> {
     fn le_defined(self, other: Self) -> bool {
-        self.addr == other.addr &&
-            match (self.provenance, other.provenance) {
+        self.thin_pointer.addr == other.thin_pointer.addr &&
+        // Feedback wanted: Using `Option<PointerMeta> as DefinedRelation` would be okay aswell, 
+        // since all antecedents garantee that either both or none have metadata in the first place?
+        self.metadata == other.metadata &&
```
RalfJung, line 391: In the future PointerMeta will contain vtable pointers that can have provenance, so it's probably better to define a `DefinedRelation` on `PointerMeta`.

spec/lang/step/intrinsics.md
```diff
@@ -67,7 +67,7 @@ impl<M: Memory> Machine<M> {
         }
 
         let ptr = self.intptrcast.int2ptr(addr)?;
-        ret(Value::Ptr(ptr))
+        ret(Value::Ptr(ptr.widen(None)))
```
RalfJung, line 70: So here you are assuming that `ret_ty` is a pointer to a *sized* type. That would have to be checked above.

spec/lang/step/intrinsics.md
```diff
@@ -231,7 +231,7 @@ impl<M: Memory> Machine<M> {
 
         let alloc = self.mem.allocate(AllocationKind::Heap, size, align)?;
 
-        ret(Value::Ptr(alloc))
+        ret(Value::Ptr(alloc.widen(None)))
```
RalfJung, line 234: Same here, needs a check above that the pointee is sized.

spec/lang/step/intrinsics.md
```diff
@@ -244,7 +244,8 @@ impl<M: Memory> Machine<M> {
             throw_ub!("invalid number of arguments for `Deallocate` intrinsic");
         }
 
-        let Value::Ptr(ptr) = arguments[0].0 else {
+        // Is it UB to call this with a wide pointer?
```
RalfJung, line 247: Yeah let's reject that.

spec/lang/step/intrinsics.md
```diff
@@ -437,7 +438,7 @@ impl<M: Memory> Machine<M> {
             throw_ub!("invalid number of arguments for `AtomicStore` intrinsic");
         }
 
-        let Value::Ptr(ptr) = arguments[0].0 else {
+        let Value::Ptr(Pointer { thin_pointer: ptr, .. }) = arguments[0].0 else {
```
RalfJung, line 441: The atomic operations should not accept wide pointers.

spec/lang/step/operators.md
```diff
@@ -257,19 +257,20 @@ impl<M: Memory> Machine<M> {
...
     /// Perform a wrapping offset on the given pointer. (Can never fail.)
-    fn ptr_offset_wrapping(&self, ptr: Pointer<M::Provenance>, offset: Int) -> Pointer<M::Provenance> {
+    fn ptr_offset_wrapping(&self, ptr: ThinPointer<M::Provenance>, offset: Int) -> ThinPointer<M::Provenance> {
         ptr.wrapping_offset::<M::T>(offset)
     }
 
     /// Perform in-bounds arithmetic on the given pointer. This must not wrap,
     /// and the offset must stay in bounds of a single allocation.
-    fn ptr_offset_inbounds(&self, ptr: Pointer<M::Provenance>, offset: Int) -> Result<Pointer<M::Provenance>> {
+    // Question: Do we not care about the size here anymore? Or am I missing something?
```
RalfJung, line 266: The intrinsic requires a sized pointee.

spec/lang/syntax.md
```diff
@@ -201,6 +201,7 @@ pub enum BinOp {
     Rel(RelOp),
 
     /// Add a byte-offset to a pointer (with or without inbounds requirement).
+    /// For non-Sized pointees this leaves the metadata untouched.
```
RalfJung, line 204: This should require a sized pointee.

RalfJung (commented):
Some first comments :) Let me know if you need any other feedback to make further progress.

spec/lang/step/intrinsics.md
Indeximal: So I understand it correctly, the ways to allocate a slice is to call the allocate intrinsic with the needed size, but return some raw pointer and then manually (unsafe) cast it to a wide pointer, e.g. with `from_raw_parts`? Which also means supporting the intrinsic `core::intrinsics::aggregate_raw_ptr` is essential? I mean the alternative is to have the alloc & dealloc macros be able to deal with slices themselves.

spec/lang/syntax.md
Indeximal: Oh I guess I was looking at the new [byte_offset](https://doc.rust-lang.org/std/primitive.pointer.html#method.byte_offset-1) method, where it explicitly reassigns the metadata.

spec/lang/step/operators.md
Indeximal: Sure, but actually I am more confused about the actual signed_dereferencable function, because what it actually checks is that the range between the two pointers is within the allocation, not that the range from the new pointer to the end of the object is. To me it seems like this should instead be implemented as checking dereferencability from the newly offset pointer.

spec/lang/step/intrinsics.md
RalfJung: The usual way to create a slice is something like
```rust
let x = [0; 32];
let y: &[i32] = &x;
```
In other words, it is created via an unsizing coercion.

`from_raw_parts` can be left to a later PR.

spec/lang/syntax.md
RalfJung: Yeah but that's happening in Rust; the underlying MIR / MiniRust operation requires a sized pointee. If you look at the implementation of that function you can see that it casts the pointer to `u8`.

spec/lang/step/operators.md
RalfJung: When you offset a pointer by 10, then the old pointer needs to be dereferenceable for 10 bytes.

When you offset it by -10, it needs to be dereferenceable for -10 bytes.

We could equivalently check whether the newly offset pointer is dereferenceable for `-offset` bytes but that only seems more confusing to me.

Indeximal: I guess we don't care (yet) about the pointee object. So if the pointer points to a 16 byte struct, and it is offset, so that only 8 bytes fit in the same allocation, then this is not UB yet, only when loading from it?

spec/lang/step/intrinsics.md
RalfJung: Yes.

It will work for `Arc<[i32; N]>` to `Arc<[i32]>` though, that is also a legal unsizing coercion.

spec/lang/step/operators.md
RalfJung: Correct. Raw pointers do not have to point to valid memory, so a pointer to `[u8; 16]` that points to 4 bytes is entirely fine.

However, [`offset`](https://doc.rust-lang.org/nightly/std/primitive.pointer.html#method.offset) has the constraint that the offset operation itself must stay inside the bounds of the allocation, and that's what we are checking here.

Indeximal:
Oke, now the Sizestrategy has been readded, and well-formedness checks have been put into place, but not tested yet. All minimize tests pass, but 2 minitest tests fail due to stricter well formedness requirement, which might be correct?

spec/lang/well-formed.md
```diff
@@ -444,7 +510,9 @@ impl Statement {
             Assign { destination, source } => {
                 let left = destination.check_wf::<T>(func.locals, prog)?;
                 let right = source.check_wf::<T>(func.locals, prog)?;
+                // FIXME: to fix 2 tests, I would need an exception here for `Ptr(PtrType::Raw)` where the layout doesn't matter.
                 ensure_wf(left == right, "Statement::Assign: destination and source type differ")?;
```
Indeximal, line 514: This is the thing that now evaluates differently, since the raw pointer has to carry around its layout aswell (or at least the metadata kind, which is not a thing right now)

tooling/miniutil/src/build/ty.rs
```diff
@@ -1,9 +1,10 @@
 use crate::build::*;
 
-pub fn layout(size: Size, align: Align) -> Layout {
+pub fn layout(size: SizeStrategy, align: Align) -> Layout {
     Layout {
         size,
         align,
+        // FIXME: enums do exist, is this still good?
         inhabited: true, // currently everything is inhabited (enums don't exist yet).
```
Indeximal, line 8: Also I found this in the code, which seems outdated :shrug:

spec/lang/step/operators.md
```diff
@@ -48,9 +48,15 @@ impl<M: Memory> Machine<M> {
                 ret((Value::Int(result), Type::Int(int_ty)))
             }
             Transmute(new_ty) => {
+                // Question: Why is this not a well-formedness constraint?
+                // `core::intrinsics::transmute` states:
+                // "Both types must have the same size. Compilation will fail if this is not guaranteed."
                 if old_ty.size::<M::T>() != new_ty.size::<M::T>() {
                     throw_ub!("transmute between types of different size")
-                };
```
Indeximal, line 53: And here I am not entirely sure why this is UB and not a well formedness violation

spec/lang/well-formed.md
```diff
@@ -376,13 +425,30 @@ impl PlaceExpr {
...
                 let op_ty = operand.check_wf::<T>(locals, prog)?;
-                ensure_wf(matches!(op_ty, Type::Ptr(_)), "PlaceExpr::Deref: invalid type")?;
+                let Type::Ptr(op_ptr_ty) = op_ty else {
+                    throw_ill_formed!("PlaceExpr::Deref: invalid operand type");
+                };
+                let Some(pointee_layout) = op_ptr_ty.pointee() else {
+                    throw_ill_formed!("PlaceExpr::Deref: invalid operand type");
+                };
+                // This breaks some tests, so this is well formed, but just UB?
+                // ensure_wf(ty.size::<T>() == pointee_layout.size, "PlaceExpr::Deref: incompatible return type layout")?;
```
Indeximal, line 438: Similarly here, deref allows to create a place which isn't the same size as the pointer it dereferences, which actually breaks about 10 tests. I'm not sure if this is really what we want.

spec/lang/step/operators.md
Indeximal: Ah wait, I found the comment in MiniRust specifying that this is meant to be like that

RalfJung (commented):
Sorry, seems like I forgot to submit this comment.

Indeximal:
This will now be split into smallers PRs, starting with #225

PR #220, opened 2024-07-26
Tree Borrows without protector (Path Provenance)

sun-jacobi:
An alternative to #213, but use the path as the provenance.

### DIFF
--- a/spec/mem/tree_borrows/memory.md
+++ b/spec/mem/tree_borrows/memory.md
@@ -1,0 +1,290 @@
+# MiniRust Tree Borrows
+
+For background on Tree Borrows, see:
+
+1. [Neven's posts on Tree Borrows](https://perso.crans.org/vanille/treebor)
+2. [From Stacks to Trees: A new aliasing model for Rust](https://www.ralfj.de/blog/2023/06/02/tree-borrows.html)
+
+Similar to the [Basic Memory Model](../basic.md), we need to first define some basic data structures:
+the core date structure managing the tree is defined in [tree.md](tree.md), and the core state machine can be found in [state_machine.md](state_machine.md).
+
+The model then tracks a tree for each allocation:
+```rust
+struct TreeBorrowsAllocationExtra {
+    root: Node,
+}
+```
+
+We use a *path* to identify each node and track its location in the tree. A path is represented as a list of indices $[i_1, i_2, ..., i_k]$, where each index indicates which branch to take next.
+Below is an illustrated example:
+```
+Consider the following tree
+      A
+     / \
+    B   C
+   / \   \
+  D  E    F
+The path from A to A is represented as [].
+The path from A to B is represented as [0]
+The path from A to C is represented as [1].
+The path from A to D is represented as [0, 0].
+The path from A to E is represented as [0, 1].
+The path from A to F is represented as [1, 1].
+```
+
+```rust
+/// The index of a child in the list of child nodes.
+type ChildId = Int;
+/// A path from the root of a tree to some node inside the tree.
+type Path = List<ChildId>;
+```
+
+Then we can define the provenance of Tree Borrows as a pair consisting of the path and the allocation ID.
+
+```rust
+type TreeBorrowsProvenance = (AllocId, Path);
+type TreeBorrowsAllocation = Allocation<TreeBorrowsProvenance, TreeBorrowsAllocationExtra>;
+```
+
+
+```rust
+pub struct TreeBorrowsMemory<T: Target> {
+    allocations: List<TreeBorrowsAllocation>,
+    // FIXME: specr should add this automatically
+    _phantom: std::marker::PhantomData<T>,
+}
+
+impl<T: Target> Memory for TreeBorrowsMemory<T> {
+    type Provenance = TreeBorrowsProvenance;
+    type T = T;
+
+    fn new() -> Self {
+        Self { allocations: List::new(), _phantom: std::marker::PhantomData }
+    }
+}
+```
+
+Here we define some helper methods to implement the memory interface.
+
+```rust
+impl<T: Target> TreeBorrowsMemory<T> {
+    /// create a location state list for an allocation:
+    /// all locations inside the permission have the given permission.
+    fn init_location_states(permission: Permission, alloc_size: Size) -> List<LocationState> {
+        let mut location_states = List::new();
+        for _ in Int::ZERO..alloc_size.bytes() {
+            location_states.push(LocationState {
+                accessed: Accessed::No,
+                permission,
+            });
+        }
+
+        location_states
+    }
+
+    /// Create a new node for a pointer (reborrow)
+    fn reborrow(
+        &mut self, 
+        ptr: Pointer<TreeBorrowsProvenance>,
+        pointee_size: Size,
+        permission: Permission
+    ) -> Result<Pointer<TreeBorrowsProvenance>> {
+        let Some((alloc_id, parent_path, offset)) = self.check_ptr(ptr, pointee_size)? else {
+            return ret(ptr);
+        };
+
+        let child_path = self.allocations.mutate_at(alloc_id.0, |allocation| {
+            // Create the new child node
+            let child_node = Node {
+                children: List::new(),
+                location_states: Self::init_location_states(permission, allocation.size()),
+            };
+
+            // Add the new node to the tree
+            let child_path = allocation.extra.root.add_node(parent_path, child_node)?;
+
+            // Perform read on the new child, updating all nodes accordingly.
+            allocation.extra.root.access(Some(child_path), AccessKind::Read, offset, pointee_size)?;
+
+            ret::<Result<Path>>(child_path)
+        })?;
+
+        // Create the child pointer and return it 
+        ret(Pointer {
+            provenance: Some((alloc_id, child_path)),
+            ..ptr
+        })
+    }
+
+
+    /// Return the borrow tag, allocation ID and offset
+    fn check_ptr(&self, ptr: Pointer<TreeBorrowsProvenance>, len: Size) -> Result<Option<(AllocId, Path, Size)>> {
+        // For zero-sized accesses, there is nothing to check.
+        // (Provenance monotonicity says that if we allow zero-sized accesses
+        // for `None` provenance we have to allow it for all provenance.)
+        if len.is_zero() {
+            return ret(None);
+        }
+        // We do not even have to check for null, since no allocation will ever contain that address.
+        // Now try to access the allocation information.
+        let Some((alloc_id, path)) = ptr.provenance else {
+            // An invalid pointer.
+            throw_ub!("dereferencing pointer without provenance");
+        };
+        let allocation = self.allocations[alloc_id.0];
+
+        // Compute relative offset
+        let offset = allocation.offset_in_alloc(ptr.addr, len)?;
+
+        // All is good!
+        ret(Some((alloc_id, path, offset)))
+    }
+}
+```
+
+# Memory Operations
+Then we implement the memory model interface for the Tree Borrow.
+
+### Allocate and Deallocate
+
+We create a new tree for one allocation
+
+```rust
+impl<T: Target> Memory for TreeBorrowsMemory<T> {
+    fn allocate(&mut self, kind: AllocationKind, size: Size, align: Align) -> NdResult<Pointer<Self::Provenance>>  {
+        let addr = Allocation::pick_base_address::<T>(self.allocations, size, align)?;
+
+        // Calculate the proverance for the root node.
+        let alloc_id = AllocId(self.allocations.len());
+
+        // Create the root node for the tree.
+        // Initially, we set the permission as `Active`.
+        let root = Node {
+            children: List::new(),
+            location_states: Self::init_location_states(Permission::Active, size),
+        };
+
+        // Path to the root node
+        let path = List::new();
+
+        let allocation = Allocation {
+            addr,
+            align,
+            kind,
+            live: true,
+            data: list![AbstractByte::Uninit; size.bytes()],
+            extra: TreeBorrowsAllocationExtra { root },
+        };
+
+        self.allocations.push(allocation);
+
+        ret(Pointer { addr, provenance: Some((alloc_id, path)) })
+    }
+}
+
+impl<T: Target> Memory for TreeBorrowsMemory<T> {
+    fn deallocate(&mut self, ptr: Pointer<Self::Provenance>, kind: AllocationKind, size: Size, align: Align) -> Result {
+        let Some((alloc_id, path)) = ptr.provenance else {
+            throw_ub!("deallocating invalid pointer")
+        };
+        // This lookup will definitely work, since AllocId cannot be faked.
+        let mut allocation = self.allocations[alloc_id.0];
+
+        allocation.deallocation_check(ptr.addr, kind, size, align)?;
+
+        // Check that ptr has the permission to write the entire allocation.
+        allocation.extra.root.access(Some(path), AccessKind::Write, Size::ZERO, size)?;
+
+        // Mark it as dead. That's it.
+        self.allocations.mutate_at(alloc_id.0, |allocation| {
+            allocation.live = false;
+        });
+
+        ret(())
+    }
+}
+```
+
+### Load Operation
+
+```rust
+impl<T: Target> Memory for TreeBorrowsMemory<T> {
+    fn load(&mut self, ptr: Pointer<Self::Provenance>, len: Size, align: Align) -> Result<List<AbstractByte<Self::Provenance>>> {
+       let Some((alloc_id, path, offset)) = self.check_ptr(ptr, len)? else {
+            return ret(list![]);
+        };
+
+        let data = self.allocations.mutate_at(alloc_id.0, |allocation| {
+            // Check for aliasing violations.
+            allocation.extra.root.access(Some(path), AccessKind::Read, offset, len)?;
+
+            // Load the data.
+            allocation.load(ptr.addr, offset, len, align)
+        })?;
+
+        ret(data)
+    }
+}
+```
+
+### Store Operation
+
+```rust
+impl<T: Target> Memory for TreeBorrowsMemory<T> {
+    fn store(&mut self, ptr: Pointer<Self::Provenance>, bytes: List<AbstractByte<Self::Provenance>>, align: Align) -> Result {
+        let size = Size::from_bytes(bytes.len()).unwrap();
+        let Some((alloc_id, path, offset)) = self.check_ptr(ptr, size)? else {
+            return ret(());
+        };
+
+        self.allocations.mutate_at(alloc_id.0, |allocation| {
+            // Check for aliasing violations.
+            allocation.extra.root.access(Some(path), AccessKind::Write, offset, size)?;
+
+            // Store the data.
+            allocation.store(ptr.addr, offset, bytes, align)
+        })?;
+
+        ret(())
+    }
+}
+```
+
+### Retagging Operation
+
+```rust
+impl<T: Target> Memory for TreeBorrowsMemory<T> {
+    fn retag_ptr(&mut self, ptr: Pointer<Self::Provenance>, ptr_type: PtrType, _fn_entry: bool) -> Result<Pointer<Self::Provenance>> {
+        match ptr_type {
+            PtrType::Ref { mutbl, pointee } => {
+                let permission = match mutbl {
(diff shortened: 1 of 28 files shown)


spec/mem/tree_borrows/memory.md
```diff
@@ -0,0 +1,298 @@
...
+    fn deallocate(&mut self, ptr: Pointer<Self::Provenance>, kind: AllocationKind, size: Size, align: Align) -> Result {
+        let Some((path, alloc_id)) = ptr.provenance else {
+            throw_ub!("deallocating invalid pointer")
+        };
+        // This lookup will definitely work, since AllocId cannot be faked.
+        let mut allocation = self.allocations[alloc_id.0];
+
+        allocation.deallocation_check(ptr.addr, kind, size, align)?;
+
+        // Check that ptr has the permission to write the entire allocation
```
RalfJung, line 168:
```suggestion
        // Check that ptr has the permission to write the entire allocation.
```

spec/mem/tree_borrows/memory.md
```diff
@@ -0,0 +1,298 @@
...
+            return ret(list![]);
+        };
+
+        // Recursively update the tree and check the existence of UBs 
+        let mut allocation = self.allocations[alloc_id.0];
+
+        // Load the data
+        let data = allocation.load(ptr.addr, offset, len, align)?;
+
+        allocation.extra.root.access(path, AccessKind::Read, offset, len, true)?;
```
RalfJung, line 198:
```suggestion
        let mut allocation = self.allocations[alloc_id.0];

        // Check for aliasing violations.
        allocation.extra.root.access(path, AccessKind::Read, offset, len, true)?;
        // Load the data.
        let data = allocation.load(ptr.addr, offset, len, align)?;
```
Can this be done with `mutate_at`?

spec/mem/tree_borrows/memory.md
```diff
@@ -0,0 +1,298 @@
...
+        });
+
+        ret(())
+    }
+}
+```
+
+### Load Operation
+
+Corresponding to the `AccessKind::Read`.
```
RalfJung, line 183: I don't understand the point of this comment -- AccessKind::Read is an internal Tree Borrows thing, so not sure why it would be called out here?

I think you can have alloc, dealloc, load, store in one block and just say above that this is defining the core memory operations.

spec/mem/tree_borrows/state_machine.md
```diff
@@ -0,0 +1,115 @@
...
+    /// Represents an activated (written to) mutable reference
+    Active,
+    /// Represents a shared (immutable) reference
+    Frozen,
+    /// Represents a dead reference
+    Disabled,
+}
+```
+
+In addition, we also need to track whether a location has already been accessed.
```
RalfJung, line 19:
```suggestion
In addition, we also need to track whether a location has already been accessed with a pointer corresponding to this node.
```

spec/mem/tree_borrows/state_machine.md
```diff
@@ -0,0 +1,115 @@
...
+            (NodeRelation::Foreign, AccessKind::Write) => self.foreign_write(),
+            (NodeRelation::Foreign, AccessKind::Read) => self.foreign_read(),
+            (NodeRelation::Child, AccessKind::Read) => self.child_read(),
+            (NodeRelation::Child, AccessKind::Write) => self.child_write(),
+        }
+    }
+}
+
+impl Accessed {
+    fn transition(self, node_relation: NodeRelation) -> Accessed {
```
RalfJung, line 95:
```suggestion
    fn transition(self, node_relation: NodeRelation) -> Accessed {
        // A node is "accessed" once any of its children gets accessed.
```

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,140 @@
+# Tree Structure in Tree Borrows
+
+The core data structure of the Tree Borrows is a *Tree*. We use a tree to track reborrows in each allocation. The tree consists of several nodes, each corresponding to a borrow tag (i.e., a reborrow operation). Each node has a parent and a list of children. Additionally, on memory locations (bytes) within the allocation, we track each tag's state defined in [state_machine.md](state_machine.md).
```
RalfJung, line 3:
```suggestion
The core data structure of Tree Borrows is a *tree*, with a state machine in each node.
We use a tree to track reborrows in each allocation; each reborrow adds a new node to the tree.
The per-node state machine is defined in [state_machine.md](state_machine.md).

Structurally, we use the usual function representation of a tree: we store a list of children.
```

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,140 @@
...
+
+```rust
+struct Node {
+    children: List<Node>,
+    /// State for each location
+    location_states: List<LocationState>,
+}
+```
+
+During each memory access, we update states according to the state machine defined in [state_machine.md](state_machine.md). When a node is accessed, each node in the tree can be divided into two disjoint sets: *Child* and *Foreign*, based on its relative position to the accessed node. The Child set includes the node itself and all its descendants, while the Foreign set contains all other nodes.
```
RalfJung, line 13:
```suggestion
During each memory access, we update states according to the state machine.
When a node is accessed, each node in the tree can be divided into two disjoint sets: *child nodes* and *foreign nodes*, based on its relative position to the accessed node.
The *child* set includes the node itself and all its descendants, while the *foreign* set contains all other nodes.
```

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,140 @@
...
+    fn transition(
+        &mut self, 
+        node_relation: NodeRelation,
+        access_kind: AccessKind,
+        offset_in_alloc: Size,
+        size: Size
+    ) -> Result {
+        let offset_start = offset_in_alloc.bytes();
+        for offset in offset_start..offset_start + size.bytes() {
+            let curr_state = self.location_states[offset];
```
RalfJung, line 47:
```suggestion
            // FIXME: Use `try_mutate_at` once that works with `Result`.
            let curr_state = self.location_states[offset];
```

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    /// Recusively do state transition on the `self` node and all its descendants.
+    /// The `path` is the path from the `self` to the accessed node.
+    /// The `is_ancestor` indicates whether the `self` is an ancestor of the accessed node.
+    ///
+    /// This method will throw UBs, representing the undefined behavior captured by Tree Borrows.
+    fn access(
+        &mut self,
+        path: Path, // root -> child
```
RalfJung, line 62:
```suggestion
        path: Path, // self -> child
```
I think?

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,140 @@
...
+            let next_state = curr_state.transition(access_kind, node_relation)?;
+            self.location_states.set(offset, next_state);
+        }
+
+        ret(())
+    }
+
+    /// Recusively do state transition on the `self` node and all its descendants.
+    /// The `path` is the path from the `self` to the accessed node.
+    /// The `is_ancestor` indicates whether the `self` is an ancestor of the accessed node.
```
RalfJung, line 57:
```suggestion
    /// `path` is the path from the `self` to the accessed node.
    /// `is_ancestor` indicates whether `self` is an ancestor of the accessed node.
```

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,140 @@
...
+    ///
+    /// This method will throw UBs, representing the undefined behavior captured by Tree Borrows.
+    fn access(
+        &mut self,
+        path: Path, // root -> child
+        access_kind: AccessKind,
+        offset_in_alloc: Size,
+        size: Size,
+        is_ancestor: bool,
```
RalfJung, line 66: I am confused by this, it seems redundant with `path`? Specifically, if there is a path from `self` to the accessed node, then `self` must be an ancestor of the accessed node?

Would it work to have an `Option<Path>`, where `None` means "accessed node is not a descendant of `self`"?

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,140 @@
...
+    ) -> Result {
+        // Indicates whether the accessed node is a child or foreign
+        // *from the perspective of the `self`*.
```
RalfJung, line 69:
```suggestion
        // *from the perspective of `self`*.
```
There should be never a `the` before `self`.

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,140 @@
...
+        // If the `self` is an ancestor of the accessed node, the accessed node is a child of the `self`.
+        // Otherwise, the accessed node is a `foreign` of the `self`
+        let node_relation = if is_ancestor { NodeRelation::Child } else { NodeRelation::Foreign };
+
+        // perform state transition on the `self`
+        self.transition(node_relation, access_kind, offset_in_alloc, size)?;
+
+        let children_num = self.children.len();
+
+        // If the `self` is leaf, stop here.
```
RalfJung, line 80:
```suggestion
        // If `self` is a leaf, stop here.
```

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,140 @@
...
+        if children_num == Int::ZERO { return ret(()); }
+
+        // The path from the sub-root
```
RalfJung, line 83: Here things become somewhat more complicated than I expected.

What about the pseudo-code?
* for all children (so no `if children_num == Int::ZERO` needed, the loop will just never run then)
  * if the path starts with this child, do a child access with the path shortened by the first element
  * otherwise, do a child access with the path being `None` (i.e., child is not an ancestor of accessed node)

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,140 @@
...
+```rust
+impl Node {
+    /// Add a new child node to the tree whose root is self
+    /// The `path` is the path from the `self` to the parent of the `node`.
+    ///
+    /// Return the path from the `self` to the `node`
+    fn add_node(&mut self, parent_path: Path, child: Node) -> Result<Path> {
+        if parent_path.len() == Int::ZERO {
+            self.children.push(child);
+            return ret(list![self.children.len() - Int::ONE]);
```
RalfJung, line 115:
```suggestion
            let child_idx = self.children.len();
            self.children.push(child);
            return ret(list![child_idx]);
```

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,140 @@
...
+We also implement some helper methods for manipulating Trees.
+
+```rust
+impl Node {
+    /// Add a new child node to the tree whose root is self
+    /// The `path` is the path from the `self` to the parent of the `node`.
+    ///
+    /// Return the path from the `self` to the `node`
+    fn add_node(&mut self, parent_path: Path, child: Node) -> Result<Path> {
+        if parent_path.len() == Int::ZERO {
```
RalfJung, line 113:
```suggestion
        // If this is where we want to insert, we are done.
        if parent_path.len() == Int::ZERO {
```

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,140 @@
...
+    }
+}
+```
+
+We also implement some helper methods for manipulating Trees.
+
+```rust
+impl Node {
+    /// Add a new child node to the tree whose root is self
+    /// The `path` is the path from the `self` to the parent of the `node`.
```
RalfJung, line 109:
```suggestion
    /// `path` is the path from `self` to the parent of `node`.
```

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,140 @@
...
+    ///
+    /// Return the path from the `self` to the `node`
+    fn add_node(&mut self, parent_path: Path, child: Node) -> Result<Path> {
+        if parent_path.len() == Int::ZERO {
+            self.children.push(child);
+            return ret(list![self.children.len() - Int::ONE]);
+        }
+
+        // If the `self` is leaf, we fail to find the parent node.
```
RalfJung, line 118:
```suggestion
        // If `self` is a leaf and the path keeps going, then the path is invalid.
```

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,140 @@
...
+        if self.children.len() == Int::ZERO {
+            panic!("Tree Borrows: Not found the parent node");
```
RalfJung, line 120:
```suggestion
            panic!("Node::add_node: invalid parent path");
```

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        // The `path` indicates which node to access next; call it the sub-root.
+        let sub_root_id = parent_path.first().unwrap();
+
+        // The path from the sub-root to the child
+        let sub_path = parent_path.subslice_with_length(Int::ONE, parent_path.len() - Int::ONE);
```
RalfJung, line 127: We should add a `split_first` method to `List`, like [this one](https://doc.rust-lang.org/nightly/std/primitive.slice.html#method.split_first).

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,140 @@
...
+
+        self.children.mutate_at(sub_root_id, |sub_root| {
```
RalfJung, line 129:
```suggestion
        // Find the right child, and add the node there.
        self.children.mutate_at(sub_root_id, |sub_root| {
```

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,140 @@
...
+            let mut child_path = sub_root.add_node(sub_path, child)?;
+            // FIXME: use push_front
```
RalfJung, line 131:
```suggestion
            let mut child_path = sub_root.add_node(sub_path, child)?;
            // We got a path from `sub_root` to the new child; update it to start at `self`.
            // FIXME: use push_front
```

tooling/minimize/tests/pass/tree_borrows/sb_fails.rs
```diff
@@ -0,0 +1,65 @@
...
+    pub fn main() {
+        foo(&mut (1, 2));
+    }
+}
+
+fn main() {
+    fnentry_invalidation::main();
+    pass_invalid_mut::main();
+    return_invalid_mut::main();
+}
```
RalfJung, line 65: Missing newline at end of file (you can see it in the github diff)

tooling/minimize/tests/pass/tree_borrows/sb_fails.rs
```diff
@@ -0,0 +1,65 @@
+//@ compile-flags: --minimize-tree-borrows
+
+// The tests were taken from Miri Tree Borrows 
+// https://github.com/rust-lang/miri/blob/6680b2fa09496b60c40c6ce09449f46efbf253d5/tests/pass/tree_borrows/sb_fails.rs
+// FIXME: Some tests in the original test suite are not currently supported by MiniRust
```
RalfJung, line 5: Can you please add a list of which tests, so that we can add them later? This applies to everywhere that you added this FIXME.

tooling/minimize/tests/ub/tree_borrows/array_access_offset.rs
RalfJung, line 1: I don't understand the system behind most of these `ub` tests. Please add a comment in each one saying what is being tested here.

tooling/minimize/tests/ub/tree_borrows/array_overlapping_write_read.rs
```diff
@@ -0,0 +1,12 @@
+//@ compile-flags: --minimize-tree-borrows
+fn main() {
+    unsafe { 
+        let mut data = [42u8, 57]; 
+        let x = &mut data[0] as *mut u8; // (x, [R, R]) 
+        let y = &mut data[1] as *mut u8; // (x, [R, R]) (y, [R, R]) 
+        *x = 42; // (x, [A, R]) (y, [D, R]) 
+        *y = 57; // (x, [A, D]) (y, [D, A])
```
RalfJung, line 8: Why does a write-read test to two writes?

RalfJung (commented):
Thanks! I really like how this path-based version worked out. :)

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,145 @@
...
+            })?;
+        }
+
+        ret(())
+    }
+
+    /// Recusively do state transition on `self` node and all its descendants.
+    /// `path` is the path from `self` to the accessed node.
+    /// `path` is None when the accessed node is not a descendant of `self`.
+    /// The `is_ancestor` indicates whether `self` is an ancestor of the accessed node.
```
RalfJung, line 64: This line is outdated

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,145 @@
...
+        path: Option<Path>, // self -> child
+        access_kind: AccessKind,
+        offset_in_alloc: Size,
+        size: Size,
+    ) -> Result {
+        // Indicates whether the accessed node is a child or foreign
+        // *from the perspective of `self`*.
+        //
+        // If `self` is an ancestor of the accessed node, the accessed node is a child of `self`.
+        // Otherwise, the accessed node is a `foreign` of `self`
```
RalfJung, line 78:
```suggestion
        // Otherwise, the accessed node is a `foreign` of `self`.
```

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,145 @@
...
+        let node_relation = if path.is_some() { NodeRelation::Child } else { NodeRelation::Foreign };
+
+        // perform state transition on `self`
```
RalfJung, line 81:
```suggestion
        // Perform state transition on `self`.
```

spec/mem/tree_borrows/tree.md
```diff
@@ -0,0 +1,145 @@
...
+        self.transition(node_relation, access_kind, offset_in_alloc, size)?;
+
+        for child_id in Int::ZERO..self.children.len() {
+            // If the path starts with this child, do a child access with the path shortened by the first element.
+            // Otherwise, do a child access with the path being None (i.e., child is not an ancestor of accessed node)
+            let sub_path = if let Some(path) = path {
+                match path.first() {
+                    Some(first) if first == child_id => Some(path.subslice_with_length(Int::ONE, path.len() - Int::ONE)),
```
RalfJung, line 89: This should also `split_first` once that exists:
`if let Some((head, tail)) = path.and_then(|p| p.split_first()) { ... }`

sun-jacobi:
Fixed above issues :)

tooling/minimize/tests/ub/tree_borrows/mutable_ref_child_write_frozen.rs
```diff
@@ -0,0 +1,15 @@
+//@ compile-flags: --minimize-tree-borrows
+
+// Check that a foreign read makes an active mutable reference frozen.
+// After that, writing to this mutable reference becomes invalid.
```
RalfJung, line 4:
```suggestion
// After that, writing to this mutable reference is UB.
```
same for all the other tests

tooling/minimize/tests/ub/tree_borrows/mutable_ref_child_write_frozen.rs
```diff
@@ -0,0 +1,15 @@
...
+
+fn main() {
+    let x = &mut 31; // (x, Reserved)   
+    let xraw = x as *mut i32; // (x, Reserved) 
+    let y = unsafe { &mut *xraw }; // (x, Reserved) -> (y, Reserved)
+    *y = 42; // (x, Active) -> (y, Active)
+    assert!(unsafe { *xraw } == 42); // (x, Active) -> (y, Frozen)
+
+    assert!(*y == 42); // Child read to Frozen is ok.
```
RalfJung, line 13:
```suggestion
    assert!(*y == 42); // Child read from Frozen is ok.
```
Same everywhere else: it's "read from", not "read to"

RalfJung (commented):
Looks good! Just some fine-tuning of the comments.

sun-jacobi:
@RalfJung Thank you for the detailed comments :)

RalfJung:
Thank you for the nice work. :)

Can you please squash the commits (while staying on the same base wrt master, if possible)? Then I will merge. :D

PR #221, opened 2024-07-29
[Tree Borrows] Full Tree Borrows Support

sun-jacobi:
Add the missing part of Tree Borrows, corresponding to 
1. [Part 3: Introducing Protectors](https://perso.crans.org/vanille/treebor/protectors.html) 
2. [Part 4: Dealing with Cells (interior mutability)](https://perso.crans.org/vanille/treebor/interiormut.html)


JoJoDeveloping:
Be reminded that Reserved has the special case for ReservedIM that only appears when you're not protected.

sun-jacobi:
This patch is currently stacked on #223

sun-jacobi:
By https://github.com/rust-lang/miri/pull/3732, the current protector end semantics are slightly different from the original version in the [Part 3: Introducing Protectors](https://perso.crans.org/vanille/treebor/protectors.html).

sun-jacobi:
@JoJoDeveloping  I would appreciate it if you could give me some advice on this PR :) 

FYI, the main implementation is in the`spec/mem/tree_borrows`

JoJoDeveloping:
Here are some things I noticed:
* there currently is only one kind of protectors. Miri distinguishes Strong and Weak Protectors, the former additionally preventing _any_ deallocation
* Currently, protection data is duplicated:
  + each node stores whether it is protected
  + the entire memory maintains not just a _set_ of active calls, but instead a map that for each active calls also maintains the entries protected by it.
  + it turns out that this map only ever stores the empty set, so something is wrong here: it should either be a `Set` (instead of a `Map<_, Set<_>>`), or the data stored there should actually be modified. Then the `end_call` hook also must be made coherent.
  
  this introduces an additional state coherence invariant that these two never desync. I'm afraid that having too many of these makes it in the end impossible to reason about this.
* `Permission::default` should probably panic when it's called with a `& T where T: !Freeze`.
* The handling of protected `ReservedIM` is wrong: These should never be created. Instead, when we do a protected reborrow for a `&mut T where T: !Freeze`, it should just become a normal `Reserved` instead of a `ReservedIM`. 
  + Please also add a test for this. In general, the tests introduced at the following two PRs are intricate and might be ported over, if applicable (they might use multi-threading so unsure):
    * https://github.com/rust-lang/miri/pull/3742
    * https://github.com/rust-lang/miri/pull/3732
* The code at https://github.com/minirust/minirust/blob/b5fe3fc42cd09cec4d6a0f29d6e69d60fbb4b1cc/spec/mem/tree_borrows/tree.md?plain=1#L60 is duplicated several times, better factor it into its own method.

sun-jacobi:
@JoJoDeveloping  Thank you for the helpful comment :)

spec/mem/tree_borrows/tree.md
```diff
@@ -46,12 +62,13 @@ impl Node {
         node_relation: NodeRelation,
         access_kind: AccessKind,
         offset_in_alloc: Size,
-        size: Size
+        size: Size,
     ) -> Result {
+        let protected = self.protected != Protected::No;
```
JoJoDeveloping, line 67: This is wrong. This used to be 
```
let protected = if let Protected::Yes(call_id) = self.protected { active_calls.contains_key(call_id)} else { false }; 
```
two days ago when I looked at it, which was (more) correct. Please change it back to what it was before everywhere. Protectors are only relevant when their calls are active.

It is still not fully correct, see the comment above.

JoJoDeveloping:
I noticed something today I forgot yesterday: I think your handling of protectors is still not fully correct. An active protector has two effects:
* It slightly changes the transition matrix. Notably, ´Reserved` becomes conflicted when there is a protector persent
* It makes many foreign accesses UB. However, this foreign access UB _only_ happens when the tag to which it is happening has been _accessed_ before (i.e. is `Accessed::Yes`). Basically, foreign accesses to `Accessed::No` locations never cause protector UB.

This causes some slight mismatches between the intended and actual semantics. Where precisely the bug is then depends on what the `actively_protected` parameter in `Permission::transition` is supposed to mean:
* if it means "protected and accessed," then `foreign_read` might not make `Reserved`'s conflicted when it's a protected but not-yet-accessed node. These should, however, become conflicted even if not yet accessed.
* if it means "protected (no matter whether accessed or not)", then `foreign_write` causes too much UB since it should not be UB for protected but not-yet-accessed nodes. There should not be UB for not yet accessed nodes.

This semantics is quite intricate, so it's rather easy to overlook this. There should presumably be a test (or several) testing that:
* non-accessed protected nodes do not cause UB on foreign writes
* non-accessed protected `Reserved` nodes still become conflicted and block child writes
* when a protector ends, it no longer has any effect (such a test seems to be missing?)

spec/mem/tree_borrows/tree.md
sun-jacobi: The current `release_protector` will make the protected node
`self.protected = Protected::No`

Does it make sense?

JoJoDeveloping: Ah, yes, that works.

spec/mem/tree_borrows/tree.md
```diff
@@ -4,13 +4,47 @@ The core data structure of Tree Borrows is a *tree*, with a state machine in eac
...
+        match (fn_entry, is_box) {
+            (true, true) => Protected::Weak,
+            (true, false) => Protected::Strong,
+            _ => Protected::No,
+        }
+    }
+}
+```
+
+Then we can define the node. Structurally, we use the usual function representation of a tree: we store a list of children.
```
JoJoDeveloping, line 38:
```suggestion
Then we can define the node. Structurally, we use the usual functional representation of a tree: we store a list of children.
```

spec/mem/tree_borrows/tree.md
```diff
@@ -130,5 +164,78 @@ impl Node {
...
+            self.children.mutate_at(child_id, |child| { child.release_protector(sub_path, &location_states, active_calls) })?;
+        }
+
+        ret(())
+    }
+
+    /// Recusively check whether there is a protected node in `self` and all its descendants.
+    fn contain_strong_protector(&self) -> bool {
+        self.protected != Protected::No || self.children.any(|child| child.contain_strong_protector())
+    }
```
JoJoDeveloping, line 239: What does this method do? It should check for `self.protected == Protected::Strong`, since it's used during deallocation to check for strong protectors, shouldn't it?

Also the comment is wrong.

spec/mem/tree_borrows/memory.md
```diff
@@ -195,6 +225,11 @@ impl<T: Target> Memory for TreeBorrowsMemory<T> {
         // Check that ptr has the permission to write the entire allocation.
         allocation.extra.root.access(Some(path), AccessKind::Write, Size::ZERO, size)?;
```
JoJoDeveloping, line 226:
```suggestion
        allocation.extra.root.access(Some(path), AccessKind::Write, Offset::ZERO, size)?;
```

`Offset` is a type alias for `Size`, using it here increases readability.

JoJoDeveloping (commented):
I've looked over it again and only found one minor thing that is still wrong, I think. So, pending this, you'd have successfully implemented all of Tree Borrows :tada: 

What remains is to be seen whether Ralf likes it :smile: In anticipation, you could take a sharp look at your code to spot some repeated patterns (like e.g. how you compute the sub-path when traversing the tree), and try factoring them into their own methods.

Besides, I'm always confused whether this PR or #223 is the actual PR, since one of them subsumes the other (right?). Could you close (or mark as draft) that other PR?

spec/mem/tree_borrows/tree.md
sun-jacobi: oops, thanks for pointing out this :)

We need a test like `deallocate_weak_protector.rs`, but unfortunately we cannot translate `Box` to minirust.
Maybe a `FIXME` for adding this test later.

sun-jacobi:
> Besides, I'm always confused whether this PR or https://github.com/minirust/minirust/pull/223 is the actual PR, since one of them subsumes the other (right?). Could you close (or mark as draft) that other PR? 

Yeah, I also noticed that it is kind of meaningless, closed it :)

spec/mem/tree_borrows/memory.md
```diff
@@ -50,6 +50,8 @@ type TreeBorrowsAllocation = Allocation<TreeBorrowsProvenance, TreeBorrowsAlloca
 ```rust
 pub struct TreeBorrowsMemory<T: Target> {
     allocations: List<TreeBorrowsAllocation>,
+    /// Map from active call IDs to their protecting nodes.
+    active_calls: Map<CallId, List<TreeBorrowsProvenance>>,
```
RalfJung, line 54: There seems to be a state invariant here: a `TreeBorrowsProvenance` is in the list for some `CallId` if and only if it is currently (weakly or strongly) protected.

That should be documented. Also, I wonder if we can avoid that redundancy...

spec/mem/tree_borrows/memory.md
```diff
@@ -109,16 +107,32 @@ impl<T: Target> TreeBorrowsMemory<T> {
...
         })
     }
 
+    /// Remove the protector.
+    /// `provenance` is the provenance of the protector.
+    /// Perform a special implicit access on all locations that have been accessed.
+    fn release_protector(&mut self, provenance: TreeBorrowsProvenance) -> Result {
+        let (alloc_id, path) = provenance;
+        self.allocations.mutate_at(alloc_id.0, |allocation| {
+            // Special case for the `Box`, since the `Box` could be deallocated during the execution of the function
```
RalfJung, line 126:
```suggestion
            // If this is weakly protected, it could have been already deallocated.
            // In that case, we don't do anything here -- fake reads from freed memory don't
            // make a lot of sense.
```

spec/mem/interface.md
```diff
@@ -59,6 +59,9 @@ pub enum AllocationKind {
     Function,
 }
 
+/// The ID of each function call (i.e stack frame)
+type CallId = Int;
```
RalfJung, line 63: Let's not have two types, `CallId` and `FrameId`. Instead, maybe add a `type FrameId;` to the `trait Memory`?

RalfJung (commented):
Just the first comments from the meeting.

spec/mem/tree_borrows/memory.md
```diff
@@ -109,16 +107,33 @@ impl<T: Target> TreeBorrowsMemory<T> {
...
+    /// Remove the protector.
+    /// `provenance` is the provenance of the protector.
+    /// Perform a special implicit access on all locations that have been accessed.
+    fn release_protector(&mut self, provenance: TreeBorrowsProvenance) -> Result {
+        let (alloc_id, path) = provenance;
+        self.allocations.mutate_at(alloc_id.0, |allocation| {
+            let protected_node = allocation.extra.root.get_node(path)?;
+            // If the protector is weak (i.e. from a `Box` argument) and the location was deallocated
+            // there is no need to peform the implicit here.
+            if protected_node.protected == Protected::Weak && !allocation.live { return ret(()); }
```
JoJoDeveloping, line 129: shoud we have this panic if the allocation is dead but the protector is strong?

spec/mem/basic.md
```diff
@@ -17,6 +17,8 @@ pub struct AllocId(Int);
 
 impl<T: Target> Memory for BasicMemory<T> {
     type Provenance = AllocId;
+    type FrameExtra = ();
```
RalfJung, line 20:
```suggestion

    type FrameExtra = ();
```

spec/mem/interface.md
```diff
@@ -72,6 +72,11 @@ pub trait Memory {
     /// The type of pointer provenance.
     type Provenance;
 
+    /// Extra information for each stack frame.
+    type FrameExtra;
+
+    fn new_extra() -> Self::FrameExtra;
```
RalfJung, line 78: Please rename to `new_call`, and move this down to just above `end_call`.

spec/mem/interface.md
```diff
@@ -114,13 +119,22 @@ pub trait Memory {
...
+        _fn_entry: bool,
+        _frame_extra: &mut Self::FrameExtra
+    ) -> Result<Pointer<Self::Provenance>> {
         if let Some(layout) = ptr_type.safe_pointee() {
             self.dereferenceable(ptr, layout.size)?;
         }
         ret(ptr)
     }
 
+    /// Memory model hook inserted at the end of each function call.
```
RalfJung, line 135:
```suggestion
    /// Memory model hook invoked at the end of each function call.
```

spec/mem/tree_borrows/memory.md
```diff
@@ -54,50 +54,50 @@ pub struct TreeBorrowsMemory<T: Target> {
     _phantom: std::marker::PhantomData<T>,
 }
 
+pub struct TreeBorrowsFrameExtra {
```
RalfJung, line 57:
```suggestion
/// Our per-frame state is the list of nodes that are protected by this call.
pub struct TreeBorrowsFrameExtra {
```

spec/mem/tree_borrows/memory.md
```diff
@@ -254,22 +284,41 @@ impl<T: Target> Memory for TreeBorrowsMemory<T> {
...
-    fn retag_ptr(&mut self, ptr: Pointer<Self::Provenance>, ptr_type: PtrType, _fn_entry: bool) -> Result<Pointer<Self::Provenance>> {
+    fn retag_ptr(
+        &mut self,
+        ptr: Pointer<Self::Provenance>,
+        ptr_type: PtrType,
+        fn_entry: bool,
+        frame_extra: &mut Self::FrameExtra,
+    ) -> Result<Pointer<Self::Provenance>> {
         match ptr_type {
+            PtrType::Ref { mutbl, pointee } if !pointee.freeze && mutbl == Mutability::Immutable => ret(ptr),
```
RalfJung, line 295:
```suggestion
            PtrType::Ref { mutbl, pointee } if !pointee.freeze && mutbl == Mutability::Immutable => {
                // Shared reference to interior mutable type: retagging is a NOP.
                ret(ptr)
            }
```

spec/mem/tree_borrows/state_machine.md
```diff
@@ -44,49 +44,63 @@ impl Permission {
...
-                Permission::Frozen => Permission::Frozen,
-                Permission::Disabled => Permission::Disabled,
-            }
-        )
+    fn foreign_read(self, actively_protected: bool) -> Result<Permission> {
+        match self {
+            Permission::Reserved { .. } if actively_protected => ret(Permission::Reserved { conflicted: true }),
+            Permission::Reserved { conflicted } => ret(Permission::Reserved { conflicted }),
+            Permission::Active if actively_protected =>
+                throw_ub!("Tree Borrows: reading from the foreign of an actively protected pointer with Active permission"),
```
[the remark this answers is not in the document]
RalfJung: (this is still open)

spec/mem/tree_borrows/state_machine.md
```diff
@@ -44,49 +49,70 @@ impl Permission {
...
             }
         )
     }
 
-    fn child_write(self) -> Result<Permission> {
+    fn child_write(self, protected: Protected) -> Result<Permission> {
         match self {
-            Permission::Reserved => ret(Permission::Active),
-            Permission::Active => ret(Permission::Active),
+            Permission::Reserved { conflicted: true } if protected.yes() =>
```
RalfJung, line 60: Why do some transitions check `protected.yes()` and others `protected.active(accessed)`?

spec/mem/tree_borrows/state_machine.md
```diff
@@ -44,49 +49,70 @@ impl Permission {
...
+            Permission::Active if protected.active(accessed) =>
+                throw_ub!("Tree Borrows: reading from the foreign of an actively protected pointer with Active permission"),
+            Permission::Reserved { .. } if protected.yes() => ret(Permission::Reserved { conflicted: true }),
+            Permission::Active => ret(Permission::Frozen),
+            _ => ret(self),
+        }
+    }
+
+    fn foreign_write(self, protected: Protected, accessed: Accessed) -> Result<Permission> {
+        if !protected.active(accessed) {
```
RalfJung, line 79: Why does this function look so different from the others?

spec/mem/tree_borrows/state_machine.md
```diff
@@ -44,49 +49,70 @@ impl Permission {
...
-            (NodeRelation::Foreign, AccessKind::Read) => self.foreign_read(),
+            (NodeRelation::Foreign, AccessKind::Write) => self.foreign_write(protected, accessed),
+            (NodeRelation::Foreign, AccessKind::Read) => self.foreign_read(protected, accessed),
+            (NodeRelation::Child, AccessKind::Write) => self.child_write(protected),
             (NodeRelation::Child, AccessKind::Read) => self.child_read(),
-            (NodeRelation::Child, AccessKind::Write) => self.child_write(),
+        }
+    }
+
+    fn default(mutbl: Mutability, pointee: PointeeInfo, protected: Protected) -> Result<Permission> {
```
RalfJung, line 110: This doesn't need `result`.

Also please move this to the top of the impl block.

spec/mem/tree_borrows/state_machine.md
```diff
@@ -44,49 +49,70 @@ impl Permission {
...
+        match mutbl {
+            Mutability::Mutable if !pointee.freeze && protected.no() => ret(Permission::ReservedIM),
+            Mutability::Mutable => ret(Permission::Reserved { conflicted: false }),
```
RalfJung, line 113: This needs a *detailed* comment explaining why we use `Reserved` for protected interior mutable references.

spec/mem/tree_borrows/state_machine.md
```diff
@@ -107,10 +133,25 @@ impl LocationState {
...
     ) -> Result {
-        self.permission = self.permission.transition(access_kind, node_relation)?;
+        self.permission = self.permission.transition(access_kind, node_relation, protected, self.accessed)?;
         self.accessed = self.accessed.transition(node_relation);
         ret(())
     }
+
+    /// Create a location state list:
+    /// all locations share the given permission.
+    fn list(permission: Permission, len: Size) -> List<LocationState> {
```
RalfJung, line 145: Maybe call this `new_list` to make it more clear that this is a constructor?

Also please move it to the top of the impl block.

spec/mem/tree_borrows/state_machine.md
```diff
@@ -44,49 +49,70 @@ impl Permission {
...
+                throw_ub!("Tree Borrows: writing to the child of a protected pointer with Conflicted Reserved permission"),
             Permission::Frozen => throw_ub!("Tree Borrows: writing to the child of a pointer with Frozen permission"),
             Permission::Disabled => throw_ub!("Tree Borrows: writing to the child of a pointer with Disabled permission"),
+            _ => ret(Permission::Active),
         }
     }
 
-    fn foreign_read(self) -> Result<Permission> {
-        ret(
+    fn foreign_read(self, protected: Protected, accessed: Accessed) -> Result<Permission> {
```
RalfJung, line 68: In Miri, all transitions are given is a `protected: bool`. I think we should do the same here, let's have the smallest possible amount of degrees of freedom.

spec/mem/tree_borrows/tree.md
```diff
@@ -130,5 +161,83 @@ impl Node {
         })
     }
 
+    /// Get a child node of `self`
+    /// `path` is the path from `self` to the target child node.
+    fn get_node(&mut self, path: Path) -> Result<Node> {
```
RalfJung, line 166: This duplicates code in `add_node`. Can you call `get_node` from there to avoid the duplication?

spec/mem/tree_borrows/tree.md
```diff
@@ -130,5 +161,83 @@ impl Node {
...
+    /// Recusively do state transition on all foreigns of the protected node.
+    /// `path` is the path from `self` to the proctected node.
+    /// `path` is None when the protected node is not a descendant of `self`.
+    /// `location_states` are the location states of the protected node.
+    fn release_protector(
+        &mut self,
+        path: Option<Path>, // self -> protected node
+        location_states: &List<LocationState>,
+    ) -> Result {
+        let node_relation = if path.is_some() { NodeRelation::Child } else { NodeRelation::Foreign };
```
RalfJung, line 200:
```suggestion
        // Indicates whether the protected node is a child or foreign
        // *from the perspective of `self`*.
        let node_relation = if path.is_some() { NodeRelation::Child } else { NodeRelation::Foreign };
```

spec/mem/tree_borrows/tree.md
```diff
@@ -130,5 +161,83 @@ impl Node {
...
+
+        // If `self` is the protected node, we are done.
```
RalfJung, line 202:
```suggestion
        // If `self` is the protected node, we are done: the special access
        // does not apply to children of the protected node.
```

spec/mem/tree_borrows/tree.md
```diff
@@ -130,5 +161,83 @@ impl Node {
...
+        }
+
+        for offset in Int::ZERO..location_states.len() {
+            let LocationState { accessed, permission } = location_states[offset];
+
+            // If the location has never been accessed, there is no need to perform an access here.
+            if accessed != Accessed::Yes { continue; }
+
+            // If there was a write to the location (i.e. if the permission is Active),
+            // we perform a write access here. Otherwise, we perform a read access here.
```
RalfJung, line 215: @JoJoDeveloping What is there was a write and then a foreign read? Then the permission will be frozen, but there was a write.  So the comment saying that this checks whether there was a write to the location doesn't seem right.

spec/mem/tree_borrows/tree.md
```diff
@@ -130,5 +161,83 @@ impl Node {
...
+    ) -> Result {
+        let node_relation = if path.is_some() { NodeRelation::Child } else { NodeRelation::Foreign };
+
+        // If `self` is the protected node, we are done.
+        if path.is_some_and(|p| p.is_empty()) {
+            self.protected = Protected::No;
+            return ret(());
+        }
+
+        for offset in Int::ZERO..location_states.len() {
```
RalfJung, line 208:
```suggestion
        // Perform state transition on `self`.
        for offset in Int::ZERO..location_states.len() {
```

spec/mem/tree_borrows/tree.md
```diff
@@ -130,5 +161,83 @@ impl Node {
...
+                Permission::Active => AccessKind::Write,
+                _ => AccessKind::Read,
+            };
+
+            self.location_states.mutate_at(Int::from(offset), |location_state|{
+                location_state.transition(access_kind, node_relation, self.protected)
+            })?;
+        }
+
+        for child_id in Int::ZERO..self.children.len() {
```
RalfJung, line 226:
```suggestion
        // Recursively visit children.
        for child_id in Int::ZERO..self.children.len() {
```

spec/mem/tree_borrows/state_machine.md
```diff
@@ -107,10 +133,25 @@ impl LocationState {
         &mut self,
         access_kind: AccessKind,
         node_relation: NodeRelation,
+        protected: Protected,
     ) -> Result {
-        self.permission = self.permission.transition(access_kind, node_relation)?;
+        self.permission = self.permission.transition(access_kind, node_relation, protected, self.accessed)?;
         self.accessed = self.accessed.transition(node_relation);
```
RalfJung, line 139: Protected nodes cannot get disabled. So what I would expect is that there is a check here: if `permission` is now `Disabled` and `protected.yes()`, then we cause UB. Isn't that the entire point of protectors? How can things work when that is not done...?

RalfJung (commented):
I have gone over all the things in `spec`. Still more to review in the tests, but I won't get to that today. :)

spec/mem/tree_borrows/tree.md
```diff
@@ -4,13 +4,46 @@ The core data structure of Tree Borrows is a *tree*, with a state machine in eac
...
+
+Then we can define the node. Structurally, we use the usual functional representation of a tree: we store a list of children.
 
 ```rust
 struct Node {
     children: List<Node>,
     /// State for each location
     location_states: List<LocationState>,
+    /// Indicates whether the node is protected by a function call,
+    /// i.e., whether the original reference passed as an argument of a function call.
```
RalfJung, line 45:
```suggestion
    /// i.e., whether the original reference passed as an argument of a function call.
    /// This will be some kind of protection (weak or strong) if and only if this node is in
    /// some frame's `extra.protectors` list.
```

spec/mem/tree_borrows/tree.md
```diff
@@ -130,5 +161,83 @@ impl Node {
...
+            };
+
+            self.children.mutate_at(child_id, |child| { child.release_protector(sub_path, &location_states) })?;
+        }
+
+        ret(())
+    }
+
+    /// Recusively check whether there is a strongly protected node in `self` and all its descendants.
+    fn contain_strong_protector(&self) -> bool {
```
RalfJung, line 239:
```suggestion
    /// This is used to reject deallocation as long as there's a strong protector anywhere.
    fn contains_strong_protector(&self) -> bool {
```

spec/mem/tree_borrows/memory.md
```diff
@@ -193,12 +218,17 @@ impl<T: Target> Memory for TreeBorrowsMemory<T> {
         allocation.deallocation_check(ptr.addr, kind, size, align)?;
 
         // Check that ptr has the permission to write the entire allocation.
-        allocation.extra.root.access(Some(path), AccessKind::Write, Size::ZERO, size)?;
+        allocation.extra.root.access(Some(path), AccessKind::Write, Offset::ZERO, size)?;
+
+        // Check that allocation is not strongly protected.
+        if allocation.extra.root.contain_strong_protector() {
```
RalfJung, line 224: Alternatively to this, we could have an `AccessKind::Dealloc` and have `access` itself also check for strong protectors. That would avoid a second traversal over the full tree. I am honestly not sure what is better... @sun-jacobi  @JoJoDeveloping any opinion?

spec/mem/tree_borrows/state_machine.md
sun-jacobi: A foreign read should make a non-accessed Reserved becomes `Reserved { conflicted: true }`.

sun-jacobi: Explicitly splitting these two states is somewhat clear to me, what do you think?

spec/mem/tree_borrows/state_machine.md
sun-jacobi: Same reason as the above reply.

spec/mem/tree_borrows/tree.md
sun-jacobi: That foreign read should be UB, I think
```rust
Permission::Active if protected.active(accessed) =>
                throw_ub!("Tree Borrows: reading from the foreign of an actively protected pointer with Active permission")
```

spec/mem/tree_borrows/state_machine.md
sun-jacobi: In this implementation, we cause UB when the permission **becomes** `Disabled`. This is slightly different from Miri, sorry for confusing

spec/mem/tree_borrows/state_machine.md
JoJoDeveloping: > Why do some transitions check `protected.yes()` and others `protected.active(accessed)`?

For reasons outlined [here](https://github.com/minirust/minirust/pull/221#issuecomment-2276884520).

JoJoDeveloping: Miri achieves this by doing this here: https://github.com/rust-lang/miri/blob/master/src/borrow_tracker/tree_borrows/tree.rs#L111

spec/mem/tree_borrows/tree.md
JoJoDeveloping: Yes, this protector end semantics only happens with protectors which means that the foreign <anything> would have been UB. I.e. that can not happen.

spec/mem/tree_borrows/memory.md
sun-jacobi: This makes sense to me :)

spec/mem/tree_borrows/state_machine.md
JoJoDeveloping: Yes, it's implemented differently to miri but achieves the same effect. I also think it is somewhat more clean.

spec/mem/tree_borrows/state_machine.md
RalfJung: Yeah I think I'd prefer if we did what Miri does, and have a single "transitions to disabled on initialized protected locations are UB" check.

spec/mem/tree_borrows/tree.md
RalfJung: Ah, good point.

Please clarify the comment to say that.

spec/mem/tree_borrows/state_machine.md
RalfJung: Why do you think this is more clean? I find the Miri version more clean -- a single general principle, uniformly enforced, is better than having to copy the same logic in several places.

spec/mem/tree_borrows/memory.md
RalfJung: Okay, let's keep this as-is then. :)

spec/mem/tree_borrows/state_machine.md
sun-jacobi: I also really like the Miri version :)

But when I was implementing this, I thought that this version could explicitly show all the state transitions, directly corresponding to the state machine.

RalfJung: To me, the state machine is what Miri has, and the rule that "when protected transitions to disabled, that's UB" is implemented outside the state machine.

spec/mem/tree_borrows/tree.md
sun-jacobi: What do you mean by "there"

spec/mem/tree_borrows/state_machine.md
sun-jacobi: > the rule that "when protected transitions to disabled, that's UB" is implemented outside the state machine.

This makes sense to me. Fixed it :)

spec/mem/tree_borrows/state_machine.md
```diff
@@ -65,49 +65,43 @@ impl Permission {
...
+    fn foreign_write(self, protected: bool) -> Result<Permission> {
         match self {
-            Permission::ReservedIM if !protected.active(accessed) => ret(Permission::ReservedIM),
-            _ if !protected.active(accessed) => ret(Permission::Disabled),
-            Permission::Reserved { .. } => throw_ub!("Tree Borrows: writing to the foreign of an actively protected pointer with Reserved permission"),
-            Permission::Active => throw_ub!("Tree Borrows: writing to the foreign of an actively protected pointer with Active permission"),
-            Permission::Frozen => throw_ub!("Tree Borrows: writing to the foreign of an actively protected pointer with Frozen permission"),
-            Permission::Disabled => panic!("Permission::foreign_write: Impossible Combination (Accessed + Protected + Disabled)"),
-            Permission::ReservedIM => panic!("Permission::foreign_write: Impossible Combination (Accessed + Protected + ReservedIM)"),
+            Permission::ReservedIM if !protected => ret(Permission::ReservedIM),
```
JoJoDeveloping, line 90: The `!protected` is unnecessary. It's still there because it being there makes the theorem that nothing protected survives foreign writes much simpler to prove (it holds by definition)

spec/mem/tree_borrows/tree.md
RalfJung: Can you change `add_node` to call `get_node` and thus avoid the duplication? Or is there some way to factor out the common code into a shared helper function?

spec/mem/tree_borrows/tree.md
```diff
@@ -216,8 +216,11 @@ impl Node {
             // If the location has never been accessed, there is no need to perform an access here.
             if accessed != Accessed::Yes { continue; }
 
-            // If there was a write to the location (i.e. if the permission is Active),
+            // If the permission is Active,
             // we perform a write access here. Otherwise, we perform a read access here.
+            // Note that since this implicit access only occurs with actively protected nodes,
+            // a foreign read/write of an Active location should be UB.
+            // This condition is equivalent to checking whether there was a (child) write to this location.
```
RalfJung, line 218:
```suggestion
            // This condition is hence equivalent to checking whether there was a (child) write to this location.
```

spec/mem/tree_borrows/state_machine.md
RalfJung: Please add a comment explaining this.

spec/mem/tree_borrows/state_machine.md
```diff
@@ -144,9 +138,19 @@ impl LocationState {
...
+            match self.permission {
+                Permission::Disabled => panic!("Impossible state combination: Accessed + Protected + Disabled"),
+                Permission::ReservedIM => panic!("Impossible state combination: Accessed + Protected + ReservedIM"),
+                Permission::Active => throw_ub!("An actively protected pointer with Active permission becomes Disabled"),
+                Permission::Frozen => throw_ub!("An actively protected pointer with Frozen permission becomes Disabled"),
+                Permission::Reserved { .. } => throw_ub!("An actively protected pointer with Reserved permission becomes Disabled"),
+            }
+        }
+        self.permission = permission;
         self.accessed = self.accessed.transition(node_relation);
```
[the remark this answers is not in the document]
RalfJung: For `accessed` it is in fact crucial to do this before the UB check, I think? If this `accessed` is "no" before this access but becomes "yes" now and the state becomes (or already is) "Disabled", that should be UB!

I don't know if that state can actually be reached though.

spec/mem/tree_borrows/state_machine.md
```diff
@@ -140,18 +140,20 @@ impl LocationState {
...
+        self.permission = old_perm.transition(access_kind, node_relation, protected)?;
+        self.accessed = self.accessed.transition(node_relation);
+
+        if self.permission == Permission::Disabled && protected && self.accessed == Accessed::Yes {
+            match old_perm {
                 Permission::Disabled => panic!("Impossible state combination: Accessed + Protected + Disabled"),
                 Permission::ReservedIM => panic!("Impossible state combination: Accessed + Protected + ReservedIM"),
                 Permission::Active => throw_ub!("An actively protected pointer with Active permission becomes Disabled"),
                 Permission::Frozen => throw_ub!("An actively protected pointer with Frozen permission becomes Disabled"),
                 Permission::Reserved { .. } => throw_ub!("An actively protected pointer with Reserved permission becomes Disabled"),
```
RalfJung, line 153:
```suggestion
                Permission::Active => throw_ub!("A protected pointer with Active permission becomes Disabled"),
                Permission::Frozen => throw_ub!("A protected pointer with Frozen permission becomes Disabled"),
                Permission::Reserved { .. } => throw_ub!("A protected pointer with Reserved permission becomes Disabled"),
```

spec/mem/tree_borrows/state_machine.md
```diff
@@ -140,18 +140,20 @@ impl LocationState {
...
         protected: bool,
     ) -> Result {
-        let permission = self.permission.transition(access_kind, node_relation, protected)?;
-        if permission == Permission::Disabled && protected && self.accessed == Accessed::Yes {
-            match self.permission {
+        let old_perm = self.permission;
+        self.permission = old_perm.transition(access_kind, node_relation, protected)?;
+        self.accessed = self.accessed.transition(node_relation);
+
+        if self.permission == Permission::Disabled && protected && self.accessed == Accessed::Yes {
```
RalfJung, line 147:
```suggestion
        // Protected nodes may never transition to "Disabled, Accessed". That is UB.
        if protected && self.permission == Permission::Disabled && self.accessed == Accessed::Yes {
```

sun-jacobi:
Fixed above.

spec/mem/tree_borrows/memory.md
```diff
@@ -144,15 +144,14 @@ impl<T: Target> TreeBorrowsMemory<T> {
...
-        if len.is_zero() {
-            return ret(None);
-        }
         // We do not even have to check for null, since no allocation will ever contain that address.
         // Now try to access the allocation information.
         let Some((alloc_id, path)) = ptr.provenance else {
+            // We allow zero-sized accesses for `None` provenance.
+            // (Provenance monotonicity says that if we allow this,
+            // we have to allow it for all provenance.)
+            if len.is_zero() { return ret(None); }
```
RalfJung, line 153: I understand now why you changed this, but it's not quite the right behavior. Probably retagging should not use check_ptr.

Please add two tests:
- it should be UB to violate the protector even with a zero-sized pointee, if it has provenance
- it should *not* be UB to have a zero-sized pointer that is outside the bounds of the allocation

spec/mem/tree_borrows/state_machine.md
```diff
@@ -49,6 +49,10 @@ impl Permission {
     fn default(mutbl: Mutability, pointee: PointeeInfo, protected: Protected) -> Permission {
         match mutbl {
             Mutability::Mutable if !pointee.freeze && protected.no() => Permission::ReservedIM,
+            // We only use `ReservedIM` for *unprotected* mutable references with interior mutability.
+            // If the reference is protected, we ignore the interior mutability.
+            // An example for why "Protected + Interior Mutability" is undesirable
+            // can be found in tooling/minimize/tests/ub/tree_borrows/protector/reservedim_spurious_write.rs.
```
RalfJung, line 55: This comment should be one line up, above the `if protected.no()`

tooling/minimize/tests/ub/tree_borrows/protector/foreign_read_active.rs
```diff
@@ -1,6 +1,6 @@
 //@ compile-flags: --minimize-tree-borrows
 
-// Check that foreign reading from an actively protected mutable reference with active permission is UB
+// Check that foreign reading from A protected mutable reference with active permission is UB
```
RalfJung:
```suggestion
// Check that foreign reading from a protected mutable reference with active permission is UB.
```

spec/mem/tree_borrows/state_machine.md
```diff
@@ -144,13 +148,14 @@ impl LocationState {
...
         if self.permission == Permission::Disabled && protected && self.accessed == Accessed::Yes {
             match old_perm {
                 Permission::Disabled => panic!("Impossible state combination: Accessed + Protected + Disabled"),
                 Permission::ReservedIM => panic!("Impossible state combination: Accessed + Protected + ReservedIM"),
-                Permission::Active => throw_ub!("An actively protected pointer with Active permission becomes Disabled"),
-                Permission::Frozen => throw_ub!("An actively protected pointer with Frozen permission becomes Disabled"),
-                Permission::Reserved { .. } => throw_ub!("An actively protected pointer with Reserved permission becomes Disabled"),
+                Permission::Active => throw_ub!("A protected pointer with Active permission becomes Disabled"),
+                Permission::Frozen => throw_ub!("A protected pointer with Frozen permission becomes Disabled"),
+                Permission::Reserved { .. } => throw_ub!("A protected pointer with Reserved permission becomes Disabled"),
```
RalfJung, line 158: Turns out we usually start these messages in lower case
```suggestion
                Permission::Active => throw_ub!("a protected pointer with Active permission becomes Disabled"),
                Permission::Frozen => throw_ub!("a protected pointer with Frozen permission becomes Disabled"),
                Permission::Reserved { .. } => throw_ub!("a protected pointer with Reserved permission becomes Disabled"),
```

spec/mem/tree_borrows/state_machine.md
```diff
@@ -103,13 +124,41 @@ impl Accessed {
...
+        protected: bool,
     ) -> Result {
-        self.permission = self.permission.transition(access_kind, node_relation)?;
+        let old_perm = self.permission;
+        self.permission = old_perm.transition(access_kind, node_relation, protected)?;
         self.accessed = self.accessed.transition(node_relation);
+
+        // Protected nodes may never transition to "Disabled, Accessed". That is UB.
+        if self.permission == Permission::Disabled && protected && self.accessed == Accessed::Yes {
+            match old_perm {
```
RalfJung, line 153:
```suggestion
            // This is UB, make sure to show a somewhat specific error.
            match old_perm {
```

spec/mem/tree_borrows/tree.md
```diff
@@ -4,13 +4,43 @@ The core data structure of Tree Borrows is a *tree*, with a state machine in eac
 We use a tree to track reborrows in each allocation; each reborrow adds a new node to the tree.
 The per-node state machine is defined in [state_machine.md](state_machine.md).
 
-Structurally, we use the usual function representation of a tree: we store a list of children.
+When a reborrow occurs at function entry (i.e. the original reference is passed as an argument), we add a *protector* to the reborrow.
```
RalfJung, line 7:
```suggestion
When a reborrow occurs at function entry (i.e. the original reference is passed as an argument), we add a *protector* to the reborrow.
When a node is protected, it is UB for this node to become `Disabled`.
```

spec/mem/tree_borrows/tree.md
```diff
@@ -4,13 +4,43 @@ The core data structure of Tree Borrows is a *tree*, with a state machine in eac
...
+There are two types of protectors: *strong* and *weak*. A *strong* protector corresponds to a normal reference, while a *weak* protector corresponds to a `Box`.
+The key difference is that only allocations without a strong protector can be deallocated during the execution of a function call.
```
RalfJung, line 9:
```suggestion
The key difference is that weak protectors still permit the memory to be deallocated, while it is UB to deallocate memory as long as there is any strongly protected node.
```

spec/mem/tree_borrows/memory.md
```diff
@@ -193,12 +218,17 @@ impl<T: Target> Memory for TreeBorrowsMemory<T> {
         allocation.deallocation_check(ptr.addr, kind, size, align)?;
 
         // Check that ptr has the permission to write the entire allocation.
-        allocation.extra.root.access(Some(path), AccessKind::Write, Size::ZERO, size)?;
+        allocation.extra.root.access(Some(path), AccessKind::Write, Offset::ZERO, size)?;
+
+        // Check that allocation is not strongly protected.
+        if allocation.extra.root.contains_strong_protector() {
```
RalfJung, line 224: So this will make it UB to deallocate memory even if there is a zero-sized strongly protected reference, right? Does that match Miri / Coq?
Cc @JoJoDeveloping

spec/mem/tree_borrows/tree.md
RalfJung: As you just said, this doesn't directly work as `add_node` needs to mutate.

Maybe we can have an `access_node` function that is basically a lens and can serve as getter and setter?

tooling/minimize/tests/ub/tree_borrows/protector/child_write_conflicted_reserved.rs
```diff
@@ -0,0 +1,17 @@
+//@ compile-flags: --minimize-tree-borrows
+
+// Check that a foreign read makes future child write to this mutable reference UB.
```
RalfJung, line 3:
```suggestion
// Check that a foreign read makes future child writes to this mutable reference UB.
```

tooling/minimize/tests/pass/tree_borrows/spurious_read.rs
```diff
@@ -0,0 +1,85 @@
...
+// This thread's job is to
+// - retag `y` protected
+// - (wait a bit that the other thread performs its spurious read)
+// - remove `y`'s protector
+// - attempt a write through `y`.
+extern "C" fn thread(thread_data_ptr: *const ()) {
+    let ThreadData { counter_ptr, data_ptr }  = unsafe { *(thread_data_ptr as *const ThreadData) };
+
+    fn foo(y: &mut u8, counter_ptr: *mut u32) -> *mut u8 {
+        unsafe { atomic_fetch_add(counter_ptr, 1); }
```
RalfJung, line 24:
```suggestion
        unsafe { atomic_fetch_add(counter_ptr, 1) };
```

tooling/minimize/tests/pass/tree_borrows/spurious_read.rs
```diff
@@ -0,0 +1,85 @@
...
+    // - do a read through `x`
+    // - remove `x`'s protector
+    // Most of the complexity here is synchronization.
+    fn foo(x: &mut u8, counter_ptr: *mut u32) -> *mut u8 {
+        unsafe { atomic_fetch_add(counter_ptr, 1); }
+        while unsafe { atomic_load(counter_ptr) == 1 } {}
+        // This is ensuring taht we have UB *without* the spurious read,
+        // so we don't read here.
+        let x = x as *mut u8;
+        unsafe { assert!(*x == 0); }
```
RalfJung, line 78:
```suggestion
        unsafe { assert!(*x == 0) };
```

tooling/minimize/tests/ub/tree_borrows/protector/reservedim_spurious_write.rs
```diff
@@ -0,0 +1,84 @@
+//@ compile-flags: --minimize-tree-borrows
+//@revisions: without with
+
+// This test was taken from Miri Tree Borrows
+// https://github.com/rust-lang/miri/blob/master/tests/fail/tree_borrows/reservedim_spurious_write.rs
+
+// Illustrating a problematic interaction between Reserved, interior mutability, and protectors.
```
RalfJung, line 7:
```suggestion
// Illustrating a problematic interaction between Reserved, interior mutability, and protectors.
// This explains why protected mutable references to `!Freeze` start in `Reserved`, not `ReservedIM`.
```

tooling/minimize/tests/ub/tree_borrows/protector/reservedim_spurious_write.rs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// Create a lazy Reserved with interior mutability.
+// Wait for the other thread's spurious write then observe the side effects
+// of that write.
+extern "C" fn thread(thread_data_ptr: *const ()) {
+    let ThreadData { counter_ptr, data_ptr }  = unsafe { *(thread_data_ptr as *const ThreadData) };
+
+    fn inner(y: &mut Cell<()>, counter_ptr: *mut u32) -> *mut u8 {
+        unsafe { atomic_fetch_add(counter_ptr, 1); } // 1 -> 2
```
RalfJung, line 31:
```suggestion
        assert!(unsafe { atomic_fetch_add(counter_ptr, 1) } == 1); // 1 -> 2
```
This makes sure that the comment is actually correct.

spec/mem/tree_borrows/memory.md
```diff
@@ -89,10 +89,13 @@ impl<T: Target> TreeBorrowsMemory<T> {
         protected: Protected,
         frame_extra: &mut TreeBorrowsFrameExtra,
     ) -> Result<Pointer<TreeBorrowsProvenance>> {
-        let Some(((alloc_id, parent_path), offset)) = self.check_ptr(ptr, pointee_size)? else {
-            return ret(ptr);
+        let Some((alloc_id, parent_path)) = ptr.provenance else {
+            // An invalid pointer.
+            throw_ub!("Tree Borrows: reborrowing pointer without provenance");
```
RalfJung, line 94: I don't think this should be UB... how can it be that the tests still pass?^^

This should just do nothing and return the pointer unchanged. Please also add a test for this case, something that creates a `*mut ()` via `ptr::without_provenance_mut` and then turns that into a reference and passes that to a function.

tooling/minimize/tests/ub/tree_borrows/protector/deallocate_strong_protector.rs
```diff
@@ -0,0 +1,18 @@
+//@ compile-flags: --minimize-tree-borrows
+
+// Check that deallocating an allocation contains strongly protected node is UB
```
JoJoDeveloping, line 3:
```suggestion
// Check that deallocating an allocation containing a strongly protected node is UB
```

tooling/minimize/tests/ub/tree_borrows/protector/foreign_write_frozen.rs
```diff
@@ -0,0 +1,19 @@
...
+
+// Check that a foreign write of a protected mutable reference with frozen permission is UB.
+
+fn foo(x: *mut i32, _y: &i32) {
+    // (x, Reserved) -> (y, P[Reserved])
+    unsafe {
+        assert!(*x == 31); // (x, Active) -> (y, P[Frozen])
+        *x = 42; // UB! Foreign Write to Protected Frozen.
+    } 
+}
```
JoJoDeveloping, line 11: This method is weird. I would have written it as follows:

```suggestion
fn foo(x: *mut i32, _frozen: &i32) {
    // (x, Reserved) -> (y, Reserved) -> (_frozen, P[Frozen])
    unsafe {
        *x = 42; // UB! Foreign Write to Protected Frozen.
    } 
}
```

In particular, the `assert!` statement is not needed. Even if it was there, the comment behind it is wrong, since the assert does not change anything.

JoJoDeveloping (commented):
I was asked to take a look at the tests. Apart from the one weirdness below, they look allright.

spec/mem/tree_borrows/memory.md
```diff
@@ -90,8 +90,7 @@ impl<T: Target> TreeBorrowsMemory<T> {
         frame_extra: &mut TreeBorrowsFrameExtra,
     ) -> Result<Pointer<TreeBorrowsProvenance>> {
         let Some((alloc_id, parent_path)) = ptr.provenance else {
-            // An invalid pointer.
-            throw_ub!("Tree Borrows: reborrowing pointer without provenance");
+            return ret(ptr);
```
RalfJung, line 93:
```suggestion
            // If the pointer has no provenance, just keep it unchanged. It can't be used to access any memory anyway.
            return ret(ptr);
```

RalfJung: Hm wait... this is only okay if `pointee_size` is zero. If it is non-zero, this should be UB!

tooling/minimize/tests/pass/zero_size_access.rs
```diff
@@ -0,0 +1,11 @@
+//@revisions: basic tree
+//@[tree]compile-flags: --minimize-tree-borrows
+
+// Check that we allow out-of-range zero-sized accesses.
+
+fn main() {
+    let xraw = &mut 42u8 as *mut u8;
+    let xzst = unsafe { xraw.add(1) as *mut () };
```
RalfJung, line 8:
```suggestion
    let xzst = unsafe { xraw.add(2) as *mut () };
```
To make this really out-of-bounds.

tooling/minimize/tests/pass/zero_size_access.rs
```diff
@@ -0,0 +1,11 @@
+//@revisions: basic tree
+//@[tree]compile-flags: --minimize-tree-borrows
+
+// Check that we allow out-of-range zero-sized accesses.
```
RalfJung, line 4:
```suggestion
// Check that we allow out-of-bounds zero-sized accesses.
```

tooling/minimize/tests/ub/tree_borrows/protector/zero_size.rs
```diff
@@ -0,0 +1,19 @@
+//@ compile-flags: --minimize-tree-borrows
+
+// Check that retagging works even with a zero-sized pointee.
+
+fn foo(x: *mut i32, y: &mut()) {
+    // (x, Reserved) -> (y, P[Reserved])
+    let y = y as *mut () as *mut i32;
+    unsafe { *y = 42 }; // y becomes Accessed.
```
RalfJung, line 8:
```suggestion
    unsafe { *y = 42 }; // y becomes Accessed, and Active.
```
Is this right?

sun-jacobi: Yes :)

tooling/minimize/tests/pass/zero_size_access.rs
sun-jacobi: Isn't `xraw.add(2)` already UB ? 

https://doc.rust-lang.org/std/primitive.pointer.html#safety-8

RalfJung: Ah yes it should be `wrapping_add`.

RalfJung: Yeah :D

spec/mem/tree_borrows/memory.md
JoJoDeveloping: It matches Coq, idk about miri (but @sun-jacobi can test that himself).

sun-jacobi: It also matches Miri :)

spec/mem/tree_borrows/memory.md
sun-jacobi: oops, I added a UB test case for this...

https://github.com/minirust/minirust/pull/221/files#diff-14483848f1f65ed137acedd3856e275ef52b1b2516bdd8e84424941890f6c7ba

tooling/minimize/tests/pass/zero_size_access.rs
RalfJung: Then for this PR please add a comment like
```
// FIXME(TreeBorrows): This test isn't quite right, the pointer is not actually ouf-of-bounds.
// Use `wrapping_add(2)` once that is possible.
```

spec/mem/tree_borrows/memory.md
```diff
@@ -90,7 +90,8 @@ impl<T: Target> TreeBorrowsMemory<T> {
         frame_extra: &mut TreeBorrowsFrameExtra,
     ) -> Result<Pointer<TreeBorrowsProvenance>> {
         let Some((alloc_id, parent_path)) = ptr.provenance else {
-            return ret(ptr);
+            if pointee_size.is_zero() { return ret(ptr);}
+            throw_ub!("Tree Borrows: reborrowing a pointer without provenance");
```
RalfJung, line 94:
```suggestion
            throw_ub!("Tree Borrows: non-zero-sized reborrow of a pointer without provenance");
```

spec/mem/tree_borrows/memory.md
```diff
@@ -90,7 +90,8 @@ impl<T: Target> TreeBorrowsMemory<T> {
         frame_extra: &mut TreeBorrowsFrameExtra,
     ) -> Result<Pointer<TreeBorrowsProvenance>> {
         let Some((alloc_id, parent_path)) = ptr.provenance else {
-            return ret(ptr);
+            if pointee_size.is_zero() { return ret(ptr);}
```
RalfJung, line 93:
```suggestion
            // Pointers without provenance cannot access any memory, so giving them a new
            // tag makes no sense. If the pointee also has size zero, this is fine, otherwise UB.
            if pointee_size.is_zero() { return ret(ptr);}
```

tooling/minimize/tests/ub/tree_borrows/protector/no_provenance_argument.rs
RalfJung, line 1: There should we two versions of this test: with `u8` and with `()`. One is UB, one isn't. Do we have both tests?

tooling/minimize/tests/pass/zero_size_access.rs
RalfJung: Ah you have deleted the test. Okay that also works.

tooling/minimize/tests/ub/tree_borrows/protector/no_provenance_argument.rs
sun-jacobi: Yes, there is also a test for `()`. See https://github.com/minirust/minirust/blob/c8043920a88ee4503a5a4d7e0fd298afc7f84cae/tooling/minimize/tests/pass/tree_borrows/no_provenance_argument.rs

sun-jacobi: > Please put the tests next to each other. Protectors aren't really relevant for this, are they?

No, this is only related to retagging. But `no_provenance_zero_sized` should be in `ub/`, while `no_provenance_zero_sized` should be in `pass/`.

RalfJung: Ah yes, makes sense.

spec/mem/tree_borrows/memory.md
RalfJung: In Miri this does not seem to be considered UB, at least under Stacked Borrows:
```rust
use std::alloc::{alloc, dealloc, Layout};

fn test(_x: &mut (), ptr: *mut u8, l: Layout) {
    unsafe { dealloc(ptr, l) };
}

fn main() {
    let l = Layout::from_size_align(1, 1).unwrap();
    let ptr = unsafe { alloc(l) };
    unsafe {test(&mut *ptr.cast::<()>(), ptr, l);}
}
```

RalfJung: But it *is* UB under Tree Borrows...

RalfJung: Okay so let's match Miri TB for now, though I think we may want to change this in the future.

spec/mem/tree_borrows/memory.md
```diff
@@ -193,12 +224,17 @@ impl<T: Target> Memory for TreeBorrowsMemory<T> {
         allocation.deallocation_check(ptr.addr, kind, size, align)?;
 
         // Check that ptr has the permission to write the entire allocation.
-        allocation.extra.root.access(Some(path), AccessKind::Write, Size::ZERO, size)?;
+        allocation.extra.root.access(Some(path), AccessKind::Write, Offset::ZERO, size)?;
+
+        // Check that allocation is not strongly protected.
```
RalfJung, line 229:
```suggestion
        // Check that allocation is not strongly protected.
        // TODO: This makes it UB to deallocate memory even if the strong protector covers 0 bytes!
        // That's different from SB, and we might want to change it in the future.
```

sun-jacobi:
Squashed :)

PR #224, opened 2024-08-13
[spec] Introducing `PointeeInfo`

sun-jacobi:
Split a part of the #221 into a single PR, since this might affect the #218

### DIFF
--- a/spec/mem/pointer.md
+++ b/spec/mem/pointer.md
@@ -43,32 +43,31 @@ impl<Provenance> Pointer<Provenance> {
 We sometimes need information what it is that a pointer points to, this is captured in a "pointer type".
 
 ```rust
-/// A "layout" describes what we know about data behind a pointer.
-pub struct Layout {
+/// Describes what we know about data behind a pointer.
+pub struct PointeeInfo {
     pub size: Size,
     pub align: Align,
     pub inhabited: bool,
+    pub freeze: bool,
 }
 
 pub enum PtrType {
     Ref {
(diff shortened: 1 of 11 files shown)


spec/mem/pointer.md
```diff
@@ -43,32 +43,32 @@ impl<Provenance> Pointer<Provenance> {
 We sometimes need information what it is that a pointer points to, this is captured in a "pointer type".
 
 ```rust
-/// A "layout" describes what we know about data behind a pointer.
-pub struct Layout {
+/// Describes what we know about data behind a pointer.
+/// (This also means we have a finite representation even when the Rust type is recursive.)
```
RalfJung, line 47: This comment should go into `types.md`, not here.

RalfJung (commented):
Looks good, thanks! Just one small comment nit.

spec/mem/pointer.md
RalfJung: My suggestion would be to add it to the `Ptr(PtrType)` variant:
```rust
/// `Ptr` represents all pointer types: references, raw pointers, boxes, and function pointers.
/// A pointer type does *not* need the full pointee type, since (de)serializing a pointer does not
/// require knowledge about the pointee. We only track basic pointee information like size and
/// alignment that is required to check reference validity. This also means types have a finite
/// representation even when the Rust type is recursive.
Ptr(PtrType),
```

sun-jacobi: Thanks :)

PR #225, opened 2024-08-17
Introduce wide and thin pointer distinction

Indeximal:
This PR creates a new Pointer type which can hold metadata.
This required a large refactor in the memory interface and other places, since the pointer without metadata is now called a `ThinPointer`.

The tests pass like this and before I continue to rebase the SizeStrategy changes onto this, I would like this to be approved, as to avoid duplicate work.

### DIFF
--- a/spec/lang/representation.md
+++ b/spec/lang/representation.md
@@ -4,6 +4,7 @@ The main purpose of [types](types.md) is to define how [values](values.md) are (
 This is the *[representation relation]*, which is defined in the following.
 `decode` converts a list of bytes into a value; this operation can fail if the byte list is not a valid encoding for the given type.
 `encode` inverts `decode`; it will always work when the value is [well-formed][well-formed-value] for the given type (which the specification must ensure, i.e. violating this property is a spec bug).
+Unsized types cannot be represented as values and thus a call to `encode` or `decode` for such types can never occur in a well-formed program.
 
 [representation relation]: https://github.com/rust-lang/unsafe-code-guidelines/blob/master/reference/src/glossary.md#representation-relation
 [well-formed-value]: well-formed.md#well-formed-values
@@ -92,8 +93,16 @@ This is required to achieve a "monotonicity" with respect to provenance (as disc
 
 ### Pointers
 
+Pointers are significantly more complex to represent than just the integer address.
+For one, they need to encode the provenance.
+When decoding, we have to deal with the possibility of the pointer bytes not all having the same provenance;
+this is define to yield a pointer without provenance.
+On the other hand, some pointers are wide pointers which also need to encode their metadata.
+The helpers `decode_ptr` and `encode_ptr` deal with the thin pointers,
+and the `decode_ptr_meta` helper inferrs the metadata kind from the pointee layout.
+
 ```rust
-fn decode_ptr<M: Memory>(bytes: List<AbstractByte<M::Provenance>>) -> Option<Pointer<M::Provenance>> {
+fn decode_ptr<M: Memory>(bytes: List<AbstractByte<M::Provenance>>) -> Option<ThinPointer<M::Provenance>> {
     if bytes.len() != M::T::PTR_SIZE.bytes() { throw!(); }
     // Convert into list of bytes; fail if any byte is uninitialized.
     let bytes_data = bytes.try_map(|b| b.data())?;
@@ -105,25 +114,28 @@ fn decode_ptr<M: Memory>(bytes: List<AbstractByte<M::Provenance>>) -> Option<Poi
             provenance = None;
         }
     }
-    ret(Pointer { addr, provenance })
+    ret(ThinPointer { addr, provenance })
 }
 
-fn encode_ptr<M: Memory>(ptr: Pointer<M::Provenance>) -> List<AbstractByte<M::Provenance>> {
+fn encode_ptr<M: Memory>(ptr: ThinPointer<M::Provenance>) -> List<AbstractByte<M::Provenance>> {
     let bytes_data = M::T::ENDIANNESS.encode(Unsigned, M::T::PTR_SIZE, ptr.addr).unwrap();
     bytes_data.map(|b| AbstractByte::Init(b, ptr.provenance))
+    
 }
 
 impl Type {
     fn decode<M: Memory>(Type::Ptr(ptr_type): Self, bytes: List<AbstractByte<M::Provenance>>) -> Option<Value<M>> {
+        // TODO(UnsizedTypes): Decode pointer metadata
         let ptr = decode_ptr::<M>(bytes)?;
         if !ptr_type.addr_valid(ptr.addr) {
-            return None;
+            throw!();
         }
-        ret(Value::Ptr(ptr))
+        ret(Value::Ptr(ptr.widen(None)))
     }
     fn encode<M: Memory>(Type::Ptr(_): Self, val: Value<M>) -> List<AbstractByte<M::Provenance>> {
         let Value::Ptr(ptr) = val else { panic!() };
-        encode_ptr::<M>(ptr)
+        assert!(ptr.metadata.is_none(), "Metadata doesn't exist yet!");
+        encode_ptr::<M>(ptr.thin_pointer)
     }
 }
 ```
@@ -367,9 +379,9 @@ impl<Provenance> DefinedRelation for AbstractByte<Provenance> {
 }
 ```
 
-Similarly, on `Pointer` we say that adding provenance makes it more defined:
+Similarly, on `Pointer` we say that adding provenance in the thin pointer or metadata makes it more defined:
 ```rust
-impl<Provenance> DefinedRelation for Pointer<Provenance> {
+impl<Provenance> DefinedRelation for ThinPointer<Provenance> {
     fn le_defined(self, other: Self) -> bool {
         self.addr == other.addr &&
             match (self.provenance, other.provenance) {
@@ -377,6 +389,14 @@ impl<Provenance> DefinedRelation for Pointer<Provenance> {
                 (Some(prov1), Some(prov2)) => prov1 == prov2,
                 _ => false,
             }
+    }
+}
+
+impl<Provenance> DefinedRelation for Pointer<Provenance> {
+    fn le_defined(self, other: Self) -> bool {
+        self.thin_pointer.le_defined(other.thin_pointer) &&
+            // TODO(UnsizedTypes): There might be provenance in the meta to check in the future.
+            self.metadata == other.metadata
     }
 }
 ```
@@ -467,7 +487,7 @@ We also use this to lift retagging from pointers to compound values.
 
 ```rust
 impl<M: Memory> ConcurrentMemory<M> {
-    fn typed_store(&mut self, ptr: Pointer<M::Provenance>, val: Value<M>, ty: Type, align: Align, atomicity: Atomicity) -> Result {
+    fn typed_store(&mut self, ptr: ThinPointer<M::Provenance>, val: Value<M>, ty: Type, align: Align, atomicity: Atomicity) -> Result {
         assert!(val.check_wf(ty).is_ok(), "trying to store {val:?} which is ill-formed for {:#?}", ty);
         let bytes = ty.encode::<M>(val);
         self.store(ptr, bytes, align, atomicity)?;
@@ -475,7 +495,7 @@ impl<M: Memory> ConcurrentMemory<M> {
         ret(())
     }
 
-    fn typed_load(&mut self, ptr: Pointer<M::Provenance>, ty: Type, align: Align, atomicity: Atomicity) -> Result<Value<M>> {
+    fn typed_load(&mut self, ptr: ThinPointer<M::Provenance>, ty: Type, align: Align, atomicity: Atomicity) -> Result<Value<M>> {
         let bytes = self.load(ptr, ty.size::<M::T>(), align, atomicity)?;
         ret(match ty.decode::<M>(bytes) {
             Some(val) => {
--- a/spec/lang/step/expressions.md
+++ b/spec/lang/step/expressions.md
@@ -35,19 +35,19 @@ impl<M: Memory> Machine<M> {
             Constant::Bool(b) => Value::Bool(b),
             Constant::GlobalPointer(relocation) => {
                 let ptr = self.global_ptrs[relocation.name].wrapping_offset::<M::T>(relocation.offset.bytes());
-                Value::Ptr(ptr)
+                Value::Ptr(ptr.widen(None))
             },
             Constant::FnPointer(fn_name) => {
-                Value::Ptr(Pointer {
+                Value::Ptr(ThinPointer {
                     addr: self.fn_addrs[fn_name],
                     provenance: None,
-                })
+                }.widen(None))
             },
             Constant::PointerWithoutProvenance(addr) => {
-                Value::Ptr(Pointer {
+                Value::Ptr(ThinPointer {
                     addr,
                     provenance: None,
-                })
+                }.widen(None))
             }
         })
     }
@@ -112,7 +112,7 @@ impl<M: Memory> Machine<M> {
         // We don't require the variant to be valid,
         // we are only interested in the bytes that the discriminator actually touches.
         let accessor = |idx: Offset, size: Size| {
-            let ptr = self.ptr_offset_inbounds(place.ptr, idx.bytes())?;
+            let ptr = self.ptr_offset_inbounds(place.ptr.thin_pointer, idx.bytes())?;
             // We have ensured that the place is aligned, so no alignment requirement here.
             self.mem.load(ptr, size, Align::ONE, Atomicity::None)
         };
@@ -136,7 +136,7 @@ impl<M: Memory> ConcurrentMemory<M> {
             throw_ub!("loading from a place based on a misaligned pointer");
         }
         // Alignment was already checked.
-        ret(self.typed_load(place.ptr, ty, Align::ONE, Atomicity::None)?)
+        ret(self.typed_load(place.ptr.thin_pointer, ty, Align::ONE, Atomicity::None)?)
     }
 }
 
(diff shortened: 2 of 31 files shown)


spec/lang/representation.md
```diff
@@ -92,8 +93,17 @@ This is required to achieve a "monotonicity" with respect to provenance (as disc
 
 ### Pointers
 
+Pointers are significantly more complex to represent than just the integer address.
+For one, they need to encode the provenance and decode it only when
+all bytes match.
```
RalfJung, line 98:
```suggestion
For one, they need to encode the provenance.
When decoding, we have to deal with the possibility of the pointer bytes not all having the same provenance;
this is define to yield a pointer without provenance.
```

spec/lang/representation.md
```diff
@@ -92,8 +93,17 @@ This is required to achieve a "monotonicity" with respect to provenance (as disc
...
+On the other hand, some pointers are wide pointers which also need
+to encode their metadata.
+The helpers `decode_ptr` and `encode_ptr` deal with the thin pointers,
+and the `decode_ptr_meta` helper inferrs the metadata kind from the
+pointee layout.
```
RalfJung, line 103: Elsewhere in this file we are using one-(sub)sentence-per-line, please stick to that style.

spec/lang/representation.md
```diff
@@ -105,25 +115,45 @@ fn decode_ptr<M: Memory>(bytes: List<AbstractByte<M::Provenance>>) -> Option<Poi
...
+        _ => { throw!(); }
+    }
+}
+
+fn encode_ptr_meta<M: Memory>(meta: Option<PointerMeta>) -> List<AbstractByte<M::Provenance>> {
+    match meta {
+        None => list!(),
+        Some(_) => unreachable!("TODO: others"),
+    }
 }
```
RalfJung, line 139: I don't understand the point of adding these functions, they are quite confusing.

I think pointer encode/decode should stay largely unchanged in this PR, and simply set the metadata field to `None` and assert that it is `None` in `encode`.

spec/lang/representation.md
```diff
@@ -371,8 +401,10 @@ Similarly, on `Pointer` we say that adding provenance makes it more defined:
 ```rust
 impl<Provenance> DefinedRelation for Pointer<Provenance> {
     fn le_defined(self, other: Self) -> bool {
-        self.addr == other.addr &&
-            match (self.provenance, other.provenance) {
+        self.thin_pointer.addr == other.thin_pointer.addr &&
```
RalfJung, line 404: Please don't skip over abstractions. Instead, add an `impl DefinedRelation for ThinPointer` and use that from here.

spec/lang/step/expressions.md
```diff
@@ -35,19 +35,19 @@ impl<M: Memory> Machine<M> {
...
-                Value::Ptr(ptr)
+                Value::Ptr(ptr.widen(None))
             },
             Constant::FnPointer(fn_name) => {
-                Value::Ptr(Pointer {
+                Value::Ptr((ThinPointer {
                     addr: self.fn_addrs[fn_name],
                     provenance: None,
-                })
+                }).widen(None))
```
RalfJung, line 44: There seem to be unnecessary parentheses here? (And in a bunch of other places as well.)

spec/lang/step/expressions.md
```diff
@@ -269,8 +269,8 @@ impl<M: Memory> Machine<M> {
         };
         assert!(offset <= ty.size::<M::T>());
 
-        let ptr = self.ptr_offset_inbounds(root.ptr, offset.bytes())?;
-        ret((Place { ptr, ..root }, field_ty))
+        let ptr = self.ptr_offset_inbounds(root.ptr.thin_pointer, offset.bytes())?;
+        ret((Place { ptr: ptr.widen(root.ptr.metadata), ..root }, field_ty))
```
RalfJung, line 273: This is not in general correct -- only projections to the last field preserve metadata.

So maybe make this `None` for now and add a TODO.

spec/lang/step/intrinsics.md
```diff
@@ -300,7 +306,7 @@ impl<M: Memory> Machine<M> {
             throw_ub!("invalid number of arguments for `Spawn` intrinsic");
         }
 
-        let Value::Ptr(ptr) = arguments[0].0 else {
+        let Value::Ptr(Pointer { thin_pointer: ptr, .. }) = arguments[0].0 else {
             throw_ub!("invalid first argument to `Spawn` intrinsic: not a pointer");
```
RalfJung, line 310: This should also require a thin pointer.

spec/lang/step/intrinsics.md
```diff
@@ -40,7 +40,7 @@ impl<M: Memory> Machine<M> {
         if arguments.len() != 1 {
             throw_ub!("invalid number of arguments for `PointerExposeProvenance` intrinsic");
         }
-        let Value::Ptr(ptr) = arguments[0].0 else {
+        let Value::Ptr(Pointer { thin_pointer: ptr, ..}) = arguments[0].0 else {
```
RalfJung, line 43: MIR currently requires this to be a thin pointer so I think we should do the same.

spec/lang/step/operators.md
```diff
@@ -234,7 +234,7 @@ impl<M: Memory> Machine<M> {
         match val {
             Value::Int(i) => i,
             Value::Bool(b) => if b { Int::from(1) } else { Int::from(0) },
-            Value::Ptr(p) => p.addr,
+            Value::Ptr(p) => p.thin_pointer.addr,
```
RalfJung, line 237: This just ignores the metadata entirely for comparisons... whereas in rustc comparison behaves as-if this was a tuple. So we'll have to do something with the metadata. For now we can at least assert that it is None to ensure we notice if this ever goes wrong.

spec/lang/step/operators.md
```diff
@@ -257,19 +257,20 @@ impl<M: Memory> Machine<M> {
...
     /// Perform a wrapping offset on the given pointer. (Can never fail.)
-    fn ptr_offset_wrapping(&self, ptr: Pointer<M::Provenance>, offset: Int) -> Pointer<M::Provenance> {
+    fn ptr_offset_wrapping(&self, ptr: ThinPointer<M::Provenance>, offset: Int) -> ThinPointer<M::Provenance> {
         ptr.wrapping_offset::<M::T>(offset)
     }
 
     /// Perform in-bounds arithmetic on the given pointer. This must not wrap,
     /// and the offset must stay in bounds of a single allocation.
-    fn ptr_offset_inbounds(&self, ptr: Pointer<M::Provenance>, offset: Int) -> Result<Pointer<M::Provenance>> {
+    // Question: Do we not care about the size here anymore? Or am I missing something?
```
RalfJung, line 266: Do we don't care about the size here.

spec/lang/step/operators.md
```diff
@@ -278,15 +279,17 @@ impl<M: Memory> Machine<M> {
         (left, l_ty): (Value<M>, Type),
         (right, _r_ty): (Value<M>, Type)
     ) -> Result<(Value<M>, Type)> {
-        let Value::Ptr(left) = left else { panic!("non-pointer left input to `PtrOffset`") };
+        let Value::Ptr(Pointer { thin_pointer: left, .. }) = left else {
+            panic!("non-pointer left input to `PtrOffsetFrom`")
```
RalfJung, line 283: You changed the error message here...

Also please make this panic if there is metadata.

spec/lang/step/operators.md
```diff
@@ -295,8 +298,12 @@ impl<M: Memory> Machine<M> {
...
         (right, _r_ty): (Value<M>, Type)
     ) -> Result<(Value<M>, Type)> {
-        let Value::Ptr(left) = left else { panic!("non-pointer left input to `PtrOffsetFrom`") };
-        let Value::Ptr(right) = right else { panic!("non-integer right input to `PtrOffsetFrom`") };
+        let Value::Ptr(Pointer { thin_pointer: left, .. }) = left else {
+            panic!("non-pointer left input to `PtrOffsetFrom`")
+        };
+        let Value::Ptr(Pointer { thin_pointer: right, .. }) = right else {
+            panic!("non-pointer right input to `PtrOffsetFrom`")
+        };
```
RalfJung, line 306: Please make this panic if there is metadata.

spec/lang/types.md
```diff
@@ -142,7 +142,7 @@ impl Type {
         match self {
             Int(int_type) => int_type.size,
             Bool => Size::from_bytes_const(1),
-            Ptr(_) => T::PTR_SIZE,
+            Ptr(_) => T::PTR_SIZE, // FIXME: sometimes 2 words
```
RalfJung, line 145:
```suggestion
            Ptr(_) => T::PTR_SIZE, // FIXME: take into account wide pointer metadata
```

spec/lang/types.md
```diff
@@ -163,13 +163,19 @@ impl Type {
...
-            Ptr(PtrType::Ref { pointee, .. } | PtrType::Box { pointee }) => pointee.inhabited,
+            Ptr(PtrType::Ref { pointee, .. } | PtrType::Box { pointee, .. }) => pointee.inhabited,
             Tuple { fields, .. } => fields.all(|(_offset, ty)| ty.inhabited()),
             Array { elem, count } => count == 0 || elem.inhabited(),
             Union { .. } => true,
             Enum { variants, .. } => variants.values().any(|variant| variant.ty.inhabited()),
         }
     }
+
+    /// Returns the metadata kind when this type is used as a pointee
```
RalfJung, line 174:
```suggestion
    /// Returns the metadata kind when this type is used as a pointee.
```

spec/lang/well-formed.md
```diff
@@ -31,17 +31,19 @@ impl IntType {
...
+        ensure_wf(T::valid_size(self.size), "PointeeInfo: size not valid")
     }
 }
 
 impl PtrType {
     fn check_wf<T: Target>(self) -> Result<()> {
         match self {
-            PtrType::Ref { pointee, mutbl: _ } | PtrType::Box { pointee } => {
+            PtrType::Ref { pointee, meta_kind, .. } | PtrType::Box { pointee, meta_kind } => {
+                // TODO: ensure pointee size and align fit with metadata kind.
```
RalfJung, line 42:
```suggestion
                // TODO(UnsizedTypes): ensure pointee size and align fit with metadata kind.
```
Please mark all TODOs like that. Then we can easily grep later in the semester to see if we forgot any of them.

spec/lang/well-formed.md
```diff
@@ -349,13 +351,28 @@ impl ValueExpr {
                         }
                     }
                     PtrOffset { inbounds: _ } => {
-                        ensure_wf(matches!(left, Type::Ptr(_)), "BinOp::PtrOffset: invalid left type")?;
+                        let Type::Ptr(left_ptr_ty) = left else {
+                            throw_ill_formed!("BinOp::PtrOffset: invalid left type");
```
RalfJung, line 355:
```suggestion
                            throw_ill_formed!("BinOp::PtrOffset: invalid left type: not a pointer");
```
(same below)

spec/lang/well-formed.md
```diff
@@ -349,13 +351,28 @@ impl ValueExpr {
...
+                        };
+                        if left_ptr_ty.meta_kind() != PointerMetaKind::None {
+                            throw_ill_formed!("BinOp::PtrOffset: invalid left type: unsized");
```
RalfJung, line 358:
```suggestion
                            throw_ill_formed!("BinOp::PtrOffset: invalid left type: unsized pointee");
```
(same below)

spec/mem/pointer.md
```diff
@@ -57,11 +86,16 @@ pub enum PtrType {
         mutbl: Mutability,
         /// Describes what we know about the pointee.
         pointee: PointeeInfo,
+        /// Indicates what kind of metadata this pointer carries.
+        meta_kind: PointerMetaKind,
```
RalfJung, line 90: Wouldn't it make sense to put this inside `PointeeInfo`?

spec/lang/representation.md
Indeximal: Fair, this was an artifact of the future

spec/lang/step/expressions.md
Indeximal: Oh yeah I thought they were necessary and usually rely on rust fmt to remove them if not. Guess I will actually learn rust syntax with this project xD

spec/lang/step/intrinsics.md
Indeximal: Yes indeed, rustc gives a compilation error for `*const [u16] as usize`, noting "cast through a thin pointer first"

spec/lang/step/operators.md
RalfJung: Yes this is quite poorly documented. And indeed, it is lexicographically sorted.

For now, I think you should make well-formed reject and wide pointer passed to this operand, and then add a panic if `p` has metadata.

spec/mem/pointer.md
Indeximal: Maybe. But it felt a bit nicer to have all meta_kind here and not just in the raw variant. And I didn't want to mess with the PointeeInfo if not necessary, since the other PR is using it aswell.

There is also a debate to be had, whether using `Option<T>` or `T` with a `None` variant is nicer. I noticed a the weird decision of mine to go for option 1 with `PointerMeta`, which makes checking and constructing `None`s easier and option 2 with `PointerMetaKind`, which avoids having to have the signature of `Option<T>` whenever it is used.

RalfJung:
> There is also a debate to be had, whether using Option<T> or T with a None variant is nicer. I noticed a slightly inconsistent decision of mine to go for option 1 with PointerMeta, which makes checking and constructing Nones easier and option 2 with PointerMetaKind, which avoids having to have the signature of Option<T> whenever it is used.

Yes I realized that. A priori the `Option<_>` variant seems nicer since then one can reuse a bunch of standard `Option` methods. But it doesn't seem to be a problem for `PointerMetaKind`...

spec/mem/pointer.md
RalfJung: IMO this is conceptually clearly part of `PointeeInfo` -- it tells us something about the pointee type -- so unless there's a strong argument to the contrary, please add it there. In MIR, a reference has mutability and pointee type; in MiniRust, we replace pointee type by `PointeeInfo`. But then adding a third thing here doesn't have a good justification IMO.

spec/lang/representation.md
```diff
@@ -146,14 +131,16 @@ impl Type {
             throw!();
         }
         let remaining_bytes = bytes.subslice_with_length(M::T::PTR_SIZE.bytes(), bytes.len() - M::T::PTR_SIZE.bytes());
-        let meta = decode_ptr_meta::<M>(ptr_type.meta_kind(), remaining_bytes)?;
-        ret(Value::Ptr(ptr.widen(meta)))
+        if !remaining_bytes.is_empty() {
+            throw!();
+        }
+        ret(Value::Ptr(ptr.widen(None)))
```
RalfJung, line 137: Can you revert this to how it was before the PR? The entire subslicing thing is unnecessary. (And I think when metadata becomes a thing, we can find a way to write this that uses less subslicing -- ideally by sharing some logic with the encode/code for tuples.)

spec/lang/step/intrinsics.md
```diff
@@ -306,7 +306,7 @@ impl<M: Memory> Machine<M> {
             throw_ub!("invalid number of arguments for `Spawn` intrinsic");
         }
 
-        let Value::Ptr(Pointer { thin_pointer: ptr, .. }) = arguments[0].0 else {
+        let Value::Ptr(Pointer { thin_pointer: ptr, metadata: None }) = arguments[0].0 else {
             throw_ub!("invalid first argument to `Spawn` intrinsic: not a pointer");
```
RalfJung, line 310:
```suggestion
            throw_ub!("invalid first argument to `Spawn` intrinsic: not a thin pointer");
```
probably occurs a few more times in this file

spec/lang/types.md
```diff
@@ -142,7 +142,8 @@ impl Type {
         match self {
             Int(int_type) => int_type.size,
             Bool => Size::from_bytes_const(1),
-            Ptr(_) => T::PTR_SIZE, // FIXME: sometimes 2 words
+            Ptr(p) if p.meta_kind() == PointerMetaKind::None => T::PTR_SIZE,
+            Ptr(_) => libspecr::Int::from_u64(2) * T::PTR_SIZE,
```
RalfJung, line 146:
```suggestion
            Ptr(_) => Int::from(2) * T::PTR_SIZE,
```

Indeximal: Doesn‘t work, `Int` is importet as `use Type::*`. Could do `Ptrsize + Ptrsize` tho

spec/lang/representation.md
```diff
@@ -397,13 +384,19 @@ impl<Provenance> DefinedRelation for AbstractByte<Provenance> {
...
+    }
+}
+
 impl<Provenance> DefinedRelation for Pointer<Provenance> {
     fn le_defined(self, other: Self) -> bool {
         self.thin_pointer.addr == other.thin_pointer.addr &&
-        // TODO: There might be provenance in the meta to check in the future.
-        self.metadata == other.metadata && 
+            self.metadata.le_defined(other.metadata) && 
             match (self.thin_pointer.provenance, other.thin_pointer.provenance) {
```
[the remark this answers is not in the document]
Indeximal: :facepalm:

Indeximal:
> > There is also a debate to be had, whether using Option or T with a None variant is nicer. I noticed a slightly inconsistent decision of mine to go for option 1 with PointerMeta, which makes checking and constructing Nones easier and option 2 with PointerMetaKind, which avoids having to have the signature of Option whenever it is used.
> 
> Yes I realized that. A priori the `Option<_>` variant seems nicer since then one can reuse a bunch of standard `Option` methods. But it doesn't seem to be a problem for `PointerMetaKind`...

Okay so, I think I like how it is right now, because the the `PointerMeta` can really not exist: when the pointee is sized; making `Option<PointerMeta>` a fitting representation. However, the kind of the metadata always exists and is known, it is just that one kind is that there is no need for metadata; therefore a `None` variant makes sense.

RalfJung:
And unfortunately there are some conflicts again, but hopefully they are minor.

PR #229, opened 2024-08-22
Refactor size constant as a SizeStrategy

Indeximal:
This refactor introduces the `SizeStrategy` that will allow slices in the next step.

### DIFF
--- a/spec/lang/step/expressions.md
+++ b/spec/lang/step/expressions.md
@@ -136,6 +136,7 @@ impl<M: Memory> ConcurrentMemory<M> {
             throw_ub!("loading from a place based on a misaligned pointer");
         }
         // Alignment was already checked.
+        // `ty` is ensured to be sized by WF of callers: Loads, Validates and InPlace arguments.
         ret(self.typed_load(place.ptr.thin_pointer, ty, Align::ONE, Atomicity::None)?)
     }
 }
@@ -143,6 +144,7 @@ impl<M: Memory> ConcurrentMemory<M> {
 impl<M: Memory> Machine<M> {
     fn eval_value(&mut self, ValueExpr::Load { source }: ValueExpr) -> Result<(Value<M>, Type)> {
         let (place, ty) = self.eval_place(source)?;
+        // WF ensures all load expressions are sized.
         let v = self.mem.place_load(place, ty)?;
 
         ret((v, ty))
@@ -250,7 +252,7 @@ impl<M: Memory> Machine<M> {
         // (We don't do a full retag here, this is not considered creating a new pointer.)
         if let Some(layout) = ptr_type.safe_pointee() {
             assert!(layout.align.is_aligned(ptr.thin_pointer.addr)); // this was already checked when the value got created
-            self.mem.dereferenceable(ptr.thin_pointer, layout.size)?;
+            self.mem.dereferenceable(ptr.thin_pointer, layout.size.compute(ptr.metadata))?;
         }
         // Check whether this pointer is sufficiently aligned.
         // Don't error immediately though! Unaligned places can still be turned into raw pointers.
@@ -273,7 +275,7 @@ impl<M: Memory> Machine<M> {
             Type::Union { fields, .. } => fields[field],
             _ => panic!("field projection on non-projectable type"),
         };
-        assert!(offset <= ty.size::<M::T>());
+        assert!(offset <= ty.size::<M::T>().compute(root.ptr.metadata));
 
         let ptr = self.ptr_offset_inbounds(root.ptr.thin_pointer, offset.bytes())?;
         // TODO(UnsizedTypes): Field projections to the last field should retain the metadata.
@@ -288,14 +290,15 @@ impl<M: Memory> Machine<M> {
         let (offset, field_ty) = match ty {
             Type::Array { elem, count } => {
                 if index >= 0 && index < count {
-                    (index * elem.size::<M::T>(), elem)
+                    let elem_size = elem.size::<M::T>().expect_sized("WF ensures array element is sized");
+                    (index * elem_size, elem)
                 } else {
                     throw_ub!("out-of-bounds array access");
                 }
             }
             _ => panic!("index projection on non-indexable type"),
         };
-        assert!(offset <= ty.size::<M::T>());
+        assert!(offset <= ty.size::<M::T>().compute(root.ptr.metadata));
 
         let ptr = self.ptr_offset_inbounds(root.ptr.thin_pointer, offset.bytes())?;
         ret((Place { ptr: ptr.widen(None), ..root }, field_ty))
--- a/spec/lang/step/intrinsics.md
+++ b/spec/lang/step/intrinsics.md
@@ -372,8 +372,8 @@ impl<M: Memory> Machine<M> {
         let PtrType::Ref { pointee, .. } = ptr_ty else {
             throw_ub!("invalid argument to `RawEq` intrinsic: not a reference");
         };
-        // TODO(UnsizedTypes): Add UB if the argument is unsized.
-        let PointeeInfo { size, align, .. } = pointee;
+        // TODO(UnsizedTypes): Throw UB if the pointee is unsized.
+        let PointeeInfo { size: SizeStrategy::Sized(size), align, .. } = pointee;
         let bytes = self.mem.load(ptr.thin_pointer, size, align, Atomicity::None)?;
 
         let Some(data) =  bytes.try_map(|byte| byte.data()) else {
@@ -444,7 +444,8 @@ impl<M: Memory> Machine<M> {
         };
 
         let (val, ty) = arguments[1];
-        let size = ty.size::<M::T>();
+        // TODO(UnsizedTypes): Throw UB if the type is unsized
+        let SizeStrategy::Sized(size) = ty.size::<M::T>();
         let Some(align) = Align::from_bytes(size.bytes()) else {
             throw_ub!("invalid second argument to `AtomicStore` intrinsic: size not power of two");
         };
@@ -474,7 +475,7 @@ impl<M: Memory> Machine<M> {
             throw_ub!("invalid first argument to `AtomicLoad` intrinsic: not a thin pointer");
         };
 
-        let size = ret_ty.size::<M::T>();
+        let size = ret_ty.size::<M::T>().expect_sized("WF ensures intrinsic return types are sized");
         let Some(align) = Align::from_bytes(size.bytes()) else {
             throw_ub!("invalid return type for `AtomicLoad` intrinsic: size not power of two");
         };
@@ -482,6 +483,7 @@ impl<M: Memory> Machine<M> {
             throw_ub!("invalid return type for `AtomicLoad` intrinsic: size too big");
         }
 
+        // `ret_ty` is ensured to be sized above.
         let val = self.mem.typed_load(ptr, ret_ty, align, Atomicity::Atomic)?;
         ret(val)
     }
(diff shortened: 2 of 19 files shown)


spec/mem/pointer.md
```diff
@@ -65,22 +65,39 @@ impl<Provenance> ThinPointer<Provenance> {
...
 
 ## Pointee
 
 We sometimes need information what it is that a pointer points to, this is captured in a "pointer type".
+However, for unsized types the layout might depend on the pointer metadata, which gives rise to the "size strategy".
 
 ```rust
 /// Describes what we know about data behind a pointer.
 pub struct PointeeInfo {
     pub meta_kind: PointerMetaKind,
```
RalfJung, line 87: This field should not be needed any more, it is defined by size strategy, right?

spec/mem/pointer.md
```diff
@@ -97,6 +114,31 @@ pub enum PtrType {
...
+    /// Returns the size when the type must be statically sized
+    pub fn unwrap_size(self) -> Size {
+        match self {
+            SizeStrategy::Sized(size) => size,
+            // TODO(UnsizedTypes): Panic on other variants
+        }
+    }
+
+    /// Returns the dynamic size, but the caller must provide compatible metadata.
+    pub fn resolve(self, meta: Option<PointerMeta>) -> Size {
```
RalfJung, line 135: Maybe `execute` or `compute`?

spec/lang/step/statements.md
```diff
@@ -127,7 +127,7 @@ impl<M: Memory> Machine<M> {
             throw_ub!("de-initializing a place based on a misaligned pointer");
         }
         // Alignment was already checked.
-        self.mem.deinit(p.ptr.thin_pointer, ty.size::<M::T>(), Align::ONE)?;
+        self.mem.deinit(p.ptr.thin_pointer, ty.size::<M::T>().resolve(p.ptr.metadata), Align::ONE)?;
```
RalfJung, line 130: Let's restrict this to sized types for now.

spec/lang/step/expressions.md
```diff
@@ -288,14 +290,14 @@ impl<M: Memory> Machine<M> {
         let (offset, field_ty) = match ty {
             Type::Array { elem, count } => {
                 if index >= 0 && index < count {
-                    (index * elem.size::<M::T>(), elem)
+                    (index * elem.size::<M::T>().expect_sized("WF ensures array element is sized"), elem)
```
RalfJung, line 293: Please let-bind the elem_size to make this a less nested expression.

spec/lang/well-formed.md
```diff
@@ -28,12 +28,33 @@ impl IntType {
...
+            SizeStrategy::Sized(size) => { ensure_wf(T::valid_size(size), "SizeStrategy: size not valid")?; }
+        };
+
+        ret(())
+    }
+
+    fn check_aligned_to(self, align: Align) -> Result<()> {
+        match self {
+            SizeStrategy::Sized(size) => {
+                ensure_wf(size.bytes() % align.bytes() == 0, "check_aligned_to: size not aligned")?;
```
RalfJung, line 43:
```suggestion
                ensure_wf(size.bytes() % align.bytes() == 0, "check_aligned_to: size not a multiple of alignment")?;
```

spec/lang/well-formed.md
```diff
@@ -54,12 +75,6 @@ impl Type {
     fn check_wf<T: Target>(self) -> Result<()> {
         use Type::*;
 
-        // Ensure that the size is valid and a multiple of the alignment.
-        let size = self.size::<T>();
-        ensure_wf(T::valid_size(size), "Type: size not valid")?;
-        let align = self.align::<T>();
-        ensure_wf(size.bytes() % align.bytes() == 0, "Type: size is not multiple of alignment")?;
```
RalfJung, line 61: Why did move this to the end of the function?

spec/lang/well-formed.md
```diff
@@ -76,24 +91,28 @@ impl Type {
                 for (offset, ty) in fields {
                     // Recursively check the field type.
                     ty.check_wf::<T>()?;
+                    // Ensure it is a sized type.
+                    ensure_wf(ty.size::<T>().is_sized(), "Type::Tuple: unsized field type")?;
                     // And ensure it fits after the one we previously checked.
                     ensure_wf(offset >= last_end, "Type::Tuple: overlapping fields")?;
-                    last_end = offset + ty.size::<T>();
+                    last_end = offset + ty.size::<T>().expect_sized("Ensured to be sized above");
```
RalfJung, line 98:
```suggestion
                    last_end = offset + ty.size::<T>().expect_sized("ensured to be sized above");
```
We follow rustc diagnostics style here (also applies a few more times in this PR)

spec/lang/well-formed.md
```diff
@@ -410,6 +440,8 @@ impl PlaceExpr {
             }
             Field { root, field } => {
                 let root = root.check_wf::<T>(locals, prog)?;
+                // TODO(UnsizedTypes): this needs to be relaxed for unsized structs
+                ensure_wf(root.size::<T>().is_sized(), "PlaceExpr::Field: root is unsized")?;
```
RalfJung, line 444: We're already checking that it is a Tuple or Union. Those are always sized. So this extra check here does not seem necessary?

spec/lang/well-formed.md
```diff
@@ -199,6 +224,7 @@ impl Constant {
             (Constant::GlobalPointer(relocation), Type::Ptr(_)) => {
                 relocation.check_wf(prog.globals)?;
             }
+            // Question: do we not care about the pointer type?
             (Constant::FnPointer(fn_name), Type::Ptr(_)) => {
```
RalfJung, line 228: We don't care very strongly, no. But we can require this to be a function pointer, sure -- why not.

spec/mem/pointer.md
```diff
@@ -98,6 +114,38 @@ pub enum PtrType {
...
+    /// Computes the dynamic size, but the caller must provide compatible metadata.
+    pub fn compute(self, meta: Option<PointerMeta>) -> Size {
+        match (self, meta) {
+            (SizeStrategy::Sized(size), None) => size,
+            _ => panic!("Pointer meta data does not match type"),
+        }
+    }
+
+    /// Returns the metadata kind which matches is needed to compute this strategy
+    pub fn matching_meta_kind(self) -> PointerMetaKind {
```
RalfJung, line 143: Just `meta_kind` is sufficient IMO.

tooling/miniutil/src/build/global.rs
```diff
@@ -1,8 +1,8 @@
 use crate::build::*;
 
 impl ProgramBuilder {
-    pub fn declare_global_zero_initialized<T: TypeConv>(&mut self) -> PlaceExpr {
-        let bytes = List::from_elem(Some(0), T::get_size().bytes());
+    pub fn declare_global_zero_initialized<T: TypeConv + Sized>(&mut self) -> PlaceExpr {
```
RalfJung, line 4: The `+ Sized` is redundant, generic types are by default required to be sized.

tooling/miniutil/src/build/mod.rs
```diff
@@ -223,6 +223,10 @@ pub fn size(bytes: impl Into<Int>) -> Size {
     Size::from_bytes(bytes).unwrap()
 }
 
+pub fn sized_size(bytes: impl Into<Int>) -> SizeStrategy {
```
RalfJung, line 226: That's an odd function name -- and also an unused function, it seems?

spec/lang/well-formed.md
Indeximal: The size calculation assumes the type to be well formed, which isn't true at the beginning of the function. E.g. the array does expect_sized of the element times the count. Now this only works after the sizedness has been asserted.

spec/lang/well-formed.md
Indeximal: Would the acronym "WF" also be lowercased in the beginning of the sentence?

spec/lang/well-formed.md
Indeximal: This is indeed unnecessary, particularly since no expect_sized is used in the field expression evaluation.

tooling/miniutil/src/build/global.rs
Indeximal: Yeah sure, I actually knew that, but somehow got confused anyway, thinking `?Sized` would be a supertrait of `TypeConv`, which a) isn't necessary, since the implicit Sized by default of course doesn't apply to trait and b) isn't actually legal rust. Reverting of course.

tooling/miniutil/src/build/mod.rs
Indeximal: Ah yes indeed. Used to be used in some tests, before they got rewritten to use TypeConv by another PR.

spec/lang/well-formed.md
RalfJung: "WF" is fine

spec/mem/tree_borrows/memory.md
```diff
@@ -236,7 +236,8 @@ impl<T: Target> Memory for TreeBorrowsMemory<T> {
         fn_entry: bool,
     ) -> Result<Pointer<Self::Provenance>> {
         ret(if let Some((permission, size, protected)) = Self::ptr_permissions(ptr_type, fn_entry) {
-            self.reborrow(ptr.thin_pointer, size, permission, protected, frame_extra)?.widen(ptr.metadata)
+            let pointee_size = size.compute(ptr.metadata);
+            self.reborrow(ptr.thin_pointer, pointee_size, permission, protected, frame_extra)?.widen(ptr.metadata)
```
Indeximal, line 240: This was the only conflict I fixed, making `ptr_permissions` return a strategy.

Indeximal:
Addressed your comments and rebased to fix the conflict

spec/lang/well-formed.md
RalfJung: Okay, makes sense. Please add a comment explaining that.

spec/lang/step/intrinsics.md
```diff
@@ -475,8 +475,7 @@ impl<M: Memory> Machine<M> {
             throw_ub!("invalid first argument to `AtomicLoad` intrinsic: not a thin pointer");
         };
 
-        // TODO(UnsizedTypes): Throw UB if the type is unsized
-        let SizeStrategy::Sized(size) = ret_ty.size::<M::T>();
+                let size = ret_ty.size::<M::T>().expect_sized("WF ensures intrinsic return types are sized");
```
RalfJung, line 478: Something seems wrong with the indentation here.

spec/mem/pointer.md
```diff
@@ -135,11 +135,12 @@ impl SizeStrategy {
...
         match (self, meta) {
             (SizeStrategy::Sized(size), None) => size,
-            _ => panic!("Pointer meta data does not match type"),
+            _ => panic!("pointer meta data does not match type"),
         }
     }
 
-    /// Returns the metadata kind which matches is needed to compute this strategy
+    /// Returns the metadata kind which matches is needed to compute this strategy.
+    /// I.e `self.meta_kind().matches(meta)` implies `self.compute(meta)` is well defined.
```
RalfJung, line 143:
```suggestion
    /// Returns the metadata kind which is needed to compute this strategy,
    /// i.e., `self.meta_kind().matches(meta)` implies `self.compute(meta)` is well-defined.
```

Indeximal:
Addressed issues and squashed

RalfJung:
Great, thanks. :)

PR #231, opened 2024-08-27
Add support for `arith_offset`

### LINKED ISSUE #227: Add support for `arith_offset`
We can't translate the following program:

```rust
fn main() {
    let ptr = &mut 42u8 as *mut u8;
    let _ = ptr.wrapping_add(1);
}
```

**Error message**:
```
thread 'rustc' panicked at minimize/src/bb.rs:288:25:
unsupported Rust intrinsic `arith_offset`
```

sun-jacobi:
Closes #227

### DIFF
--- a/tooling/minimize/src/bb.rs
+++ b/tooling/minimize/src/bb.rs
@@ -10,6 +10,13 @@ enum StatementResult {
     Intrinsic { intrinsic: IntrinsicOp, destination: PlaceExpr, arguments: List<ValueExpr> },
 }
 
+/// A MIR terminator becomes a MiniRust terminator, possibly preceded by a list
+/// of statements to execute at the end of the previous basic block.
+struct TerminatorResult {
+    stmts: List<Statement>,
+    terminator: Terminator,
+}
+
 impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
     /// Translate the given basic block and insert it into `self` with the given name.
     /// May insert more than one block because some MIR statements turn into MiniRust terminators.
@@ -40,7 +47,10 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
                 }
             }
         }
-        let terminator = self.translate_terminator(bb.terminator());
+        let TerminatorResult { stmts, terminator } = self.translate_terminator(bb.terminator());
+        for stmt in stmts.iter() {
+            cur_block_statements.push(stmt);
+        }
         let cur_block = BasicBlock { statements: cur_block_statements, terminator };
         let old = self.blocks.insert(cur_block_name, cur_block);
         assert!(old.is_none()); // make sure we do not overwrite a bb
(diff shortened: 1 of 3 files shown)


spec/mem/basic.md
```diff
@@ -118,7 +118,7 @@ impl<Provenance, Extra> Allocation<Provenance, Extra> {
         }
 
         let offset_in_alloc = addr - self.addr;
-        if offset_in_alloc < 0 || offset_in_alloc + pointee_size.bytes() > self.size().bytes() {
+        if !pointee_size.is_zero() && (offset_in_alloc < 0 || offset_in_alloc + pointee_size.bytes() > self.size().bytes()) {
```
RalfJung, line 121: Why this change? This function is never even called for zero-sized pointees, I think?

tooling/minimize/src/bb.rs
```diff
@@ -10,6 +10,11 @@ enum StatementResult {
     Intrinsic { intrinsic: IntrinsicOp, destination: PlaceExpr, arguments: List<ValueExpr> },
 }
 
+struct TerminatorResult {
```
RalfJung, line 13:
```suggestion
/// A MIR terminator becomes a MiniRust terminator, possibly preceded by a list
/// of statements to execute at the end of the previous basic block.
struct TerminatorResult {
```

tooling/minimize/src/bb.rs
```diff
@@ -229,27 +239,46 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
+                        terminator: Terminator::Intrinsic {
+                            intrinsic: IntrinsicOp::RawEq,
+                            arguments: args
+                                .iter()
+                                .map(|x| self.translate_operand(&x.node, x.span))
+                                .collect(),
+                            ret: self.translate_place(&destination, span),
+                            next_block: target.as_ref().map(|t| self.bb_name_map[t]),
+                        },
+                        stmts: List::new(),
```
RalfJung, line 265: Please put `stmts` before `terminator`, since logically they also execute first.

tooling/minimize/src/bb.rs
```diff
@@ -229,27 +239,46 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
                     },
+                "arith_offset" => {
+                    let terminator =
+                        Terminator::Goto(target.as_ref().map(|t| self.bb_name_map[t]).unwrap());
```
RalfJung, line 269:
```suggestion
                        Terminator::Goto(self.bb_name_map[target.unwrap()]);
```

tooling/minimize/src/bb.rs
```diff
@@ -229,27 +239,46 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
+                    let destination = self.translate_place(&destination, span);
+
+                    let source = ValueExpr::BinOp {
```
RalfJung, line 272: "source" is a confusing name here, maybe call it "result" or so?

tooling/minimize/tests/pass/arith_offset.rs
```diff
@@ -0,0 +1,50 @@
...
+//@[tree]compile-flags: --minimize-tree-borrows
+
+fn wrapping_offset() {
+    let data = [1u8, 2, 3, 4, 5];
+    let first = &data[0] as *const u8;
+
+    unsafe {
+        assert!(*first == 1);
+        assert!(*first.wrapping_offset(2) == 3);
+        assert!(*first.wrapping_offset(4) == 5);
```
RalfJung, line 11: Please also add a case that actually goes out-of-bounds. Something like `*first.wrapping_offset(100).wrapping_offset(-100) == 1`.

spec/mem/basic.md
RalfJung: Ah, I see, this is also used directly by TB. I think that should be done differently... I have pushed 6e77e25 which avoids calling `offset_in_alloc` when the size is zero, then no change should be needed here.

RalfJung: Sorry, the push failed... it's now pushed as commit 5e38d1c.

sun-jacobi:
Resolved above.

RalfJung:
Thanks :)

PR #232, opened 2024-08-27
Add a missing test to sb_fails.rs

sun-jacobi:


tooling/minimize/tests/pass/tree_borrows/sb_fails.rs
```diff
@@ -60,6 +57,20 @@ mod return_invalid_mut {
...
+    // Copied directly from fail/stacked_borrows/static_memory_modification.rs
+    // Version that fails TB: fail/tree_borrows/static_memory_modification.rs
+    static X: usize = 5;
+
+    #[allow(mutable_transmutes)]
+    pub fn main() {
+        let _x = unsafe {
+            std::mem::transmute::<&usize, &mut usize>(&X) // In SB this mutable reborrow fails.
+            // But in TB we are allowed to transmute as long as we don't write.
+        };
```
RalfJung, line 70: Can you add an `assert!(*x == 5);`? Seems better to also check that this actually does a read.

I will do the same on the Miri side.

RalfJung: In fact, make that `*&*x == 5`. Just to add an extra reborrow.

sun-jacobi: Done

PR #233, opened 2024-08-29
Add slice support

Indeximal:
This PR will add slice support to minirust, which includes a new SizeStrategy, MetaKind and the Type itself.

This currently includes a broken minimize test and therefore is still a draft.

### DIFF
--- a/spec/lang/representation.md
+++ b/spec/lang/representation.md
@@ -96,10 +96,10 @@ This is required to achieve a "monotonicity" with respect to provenance (as disc
 Pointers are significantly more complex to represent than just the integer address.
 For one, they need to encode the provenance.
 When decoding, we have to deal with the possibility of the pointer bytes not all having the same provenance;
-this is define to yield a pointer without provenance.
+this is defined to yield a pointer without provenance.
 On the other hand, some pointers are wide pointers which also need to encode their metadata.
-The helpers `decode_ptr` and `encode_ptr` deal with the thin pointers,
-and the `decode_ptr_meta` helper inferrs the metadata kind from the pointee layout.
+The helpers `decode_ptr` and `encode_ptr` deal with thin pointers.
+For wide pointers, `PtrType::as_wide_pair` defines the pointer as a pair of a thin pointer and some metadata type.
 
 ```rust
 fn decode_ptr<M: Memory>(bytes: List<AbstractByte<M::Provenance>>) -> Option<ThinPointer<M::Provenance>> {
@@ -123,19 +123,83 @@ fn encode_ptr<M: Memory>(ptr: ThinPointer<M::Provenance>) -> List<AbstractByte<M
     
 }
 
+impl PointerMetaKind {
+    /// Returns the representable type of the metadata if there is any.
+    pub fn ty<T: Target>(self) -> Option<Type> {
+        match self {
+            PointerMetaKind::None => None,
+            PointerMetaKind::ElementCount => Some(Type::Int(IntType { signed: Unsigned, size: T::PTR_SIZE })),
+        }
+    }
+}
+
+impl PtrType {
+    /// Returns a pair type representing this wide pointer or `None` if it is thin.
+    pub fn as_wide_pair<T: Target>(self) -> Option<Type> {
+        let meta_ty = self.meta_kind().ty::<T>()?;
+        let thin_pointer_field = (Offset::ZERO, Type::Ptr(PtrType::Raw { meta_kind: PointerMetaKind::None }));
+        let metadata_field = (T::PTR_SIZE, meta_ty);
+        ret(Type::Tuple {
+            fields: list![thin_pointer_field, metadata_field],
+            size: Int::from(2) * T::PTR_SIZE,
+            // This anyway does not matter as we only use this type to encode/decode values.
+            align: T::PTR_ALIGN,
+        })
+    }
+}
+
+impl PointerMeta {
+    fn from_value<M: Memory>(value: Value<M>) -> Option<PointerMeta> {
+        match value {
+            Value::Int(count) => Some(PointerMeta::ElementCount(count)),
+            _ => None,
+        }
+    }
+
+    fn into_value<M: Memory>(self) -> Value<M> {
+        match self {
+            PointerMeta::ElementCount(count) => Value::Int(count),
+        }
+    }
+}
+
 impl Type {
     fn decode<M: Memory>(Type::Ptr(ptr_type): Self, bytes: List<AbstractByte<M::Provenance>>) -> Option<Value<M>> {
-        // TODO(UnsizedTypes): Decode pointer metadata
-        let ptr = decode_ptr::<M>(bytes)?;
-        if !ptr_type.addr_valid(ptr.addr) {
-            throw!();
+        if let Some(pair_ty) = ptr_type.as_wide_pair::<M::T>() {
+            // This will recursively call this decode function again, but with a thin pointer type.
+            let Value::Tuple(parts) = pair_ty.decode::<M>(bytes)? else {
+                panic!("as_wide_pair always returns a tuple type");
+            };
+            let Value::Ptr(ptr) = parts[0] else {
+                panic!("as_wide_pair always returns tuple with the first field being a thin pointer");
+            };
+            let meta = PointerMeta::from_value(parts[1]);
+            assert!(meta.is_some(), "as_wide_pair always returns a suitable metadata type");
+            ret(Value::Ptr(ptr.thin_pointer.widen(meta)))
+        } else {
+            // Handle thin pointers.
+            let ptr = decode_ptr::<M>(bytes)?;
+            if !ptr_type.addr_valid(ptr.addr) {
+                throw!();
+            }
+            ret(Value::Ptr(ptr.widen(None)))
         }
-        ret(Value::Ptr(ptr.widen(None)))
     }
-    fn encode<M: Memory>(Type::Ptr(_): Self, val: Value<M>) -> List<AbstractByte<M::Provenance>> {
-        let Value::Ptr(ptr) = val else { panic!() };
-        assert!(ptr.metadata.is_none(), "Metadata doesn't exist yet!");
-        encode_ptr::<M>(ptr.thin_pointer)
+    fn encode<M: Memory>(Type::Ptr(ptr_type): Self, val: Value<M>) -> List<AbstractByte<M::Provenance>> {
+        let Value::Ptr(ptr) = val else { panic!("val is WF for a pointer") };
+
+        if let Some(pair_ty) = ptr_type.as_wide_pair::<M::T>() {
+            let thin_ptr_value = Value::Ptr(ptr.thin_pointer.widen(None));
+            let meta_data_value = ptr.metadata.expect("val is WF for ptr_type and it has metadata").into_value::<M>();
+            let tuple = Value::Tuple(list![thin_ptr_value, meta_data_value]);
+
+            // This will recursively call this encode function again, but with a thin pointer type.
+            pair_ty.encode::<M>(tuple)
+        } else {
+            // Handle thin pointers.
+            assert!(ptr.metadata.is_none(), "ptr_type and value have mismatching metadata");
+            encode_ptr::<M>(ptr.thin_pointer)
+        }
     }
 }
 ```
@@ -330,6 +394,21 @@ impl Type {
             ret(())
         }, tagger).unwrap();
         bytes
+    }
+}
+```
+
+### Unsized types
+
+Unsized types do not have values and thus there is no representation relation.
+
+```rust
+impl Type {
+    fn decode<M: Memory>(Type::Slice { .. }: Self, bytes: List<AbstractByte<M::Provenance>>) -> Option<Value<M>> {
+        panic!("tried to decode a slice")
+    }
+    fn encode<M: Memory>(Type::Slice { .. }: Self, val: Value<M>) -> List<AbstractByte<M::Provenance>> {
+        panic!("tried to endoce a slice")
     }
 }
 ```
--- a/spec/lang/step/operators.md
+++ b/spec/lang/step/operators.md
@@ -238,7 +238,7 @@ impl<M: Memory> Machine<M> {
             Value::Bool(b) => if b { Int::from(1) } else { Int::from(0) },
             Value::Ptr(p) if p.metadata.is_none() => p.thin_pointer.addr,
             // TODO(UnsizedTypes): Prepare wide pointer for rel op: `addr * (MAX_META_VALUE+1) + meta_value`
-            Value::Ptr(_) => panic!("metadata does not exist yet"),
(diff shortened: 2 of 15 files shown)


Indeximal:
I added miniutil test and made minimize compile again, but the hacky minimize test I wrote ran into several problems, including not haveing the `RValue::Len` and `predict` being unimplemented which is called by PointerWithExposedProvenance probably because of the transmute.

I should probably remove the test for a later time, when I implement unsizing coercion.

Some other feedback would be useful tho at this point :)

spec/mem/pointer.md
```diff
@@ -124,17 +130,18 @@ impl SizeStrategy {
     }
 
     /// Returns the size when the type must be statically sized.
-    pub fn expect_sized(self, _msg: &str) -> Size {
+    pub fn expect_sized(self, msg: &str) -> Size {
         match self {
             SizeStrategy::Sized(size) => size,
-            // TODO(UnsizedTypes): Panic on other variants
+            _ => panic!("expect_sized called on unsized strategy: {msg}"),
```
RalfJung, line 136:
```suggestion
            _ => panic!("expect_sized called on unsized type: {msg}"),
```

spec/mem/pointer.md
```diff
@@ -124,17 +130,18 @@ impl SizeStrategy {
...
         }
     }
 
     /// Computes the dynamic size, but the caller must provide compatible metadata.
     pub fn compute(self, meta: Option<PointerMeta>) -> Size {
         match (self, meta) {
             (SizeStrategy::Sized(size), None) => size,
+            (SizeStrategy::Slice(elem_size), Some(PointerMeta::ElementCount(count))) => count * elem_size,
```
RalfJung, line 144: Somewhere we need a FIXME to check that the resulting size is not too big.

spec/lang/representation.md
```diff
@@ -334,6 +389,19 @@ impl Type {
 }
 ```
 
+### Unrepresentable types
```
RalfJung, line 392:
```suggestion
### Unsized types

Unsized types do not have values and thus there is no representation relation.
```

spec/lang/representation.md
```diff
@@ -99,7 +99,9 @@ When decoding, we have to deal with the possibility of the pointer bytes not all
 this is define to yield a pointer without provenance.
 On the other hand, some pointers are wide pointers which also need to encode their metadata.
 The helpers `decode_ptr` and `encode_ptr` deal with the thin pointers,
-and the `decode_ptr_meta` helper inferrs the metadata kind from the pointee layout.
+and the `decode` call them when the pointer is thin.
```
RalfJung, line 102: I can't quite parse this... should it be "calls them"? And no "the"?
But it also seems unnecessary to say this, the `decode` sources are right below. Saying they deal with thin pointers conveys all the relevant high-level parts, IMO.

spec/lang/representation.md
```diff
@@ -123,19 +125,72 @@ fn encode_ptr<M: Memory>(ptr: ThinPointer<M::Provenance>) -> List<AbstractByte<M
     
 }
 
+impl PtrType {
+    // Returns a 2-tuple type representing this wide pointer or `None` if it is thin.
```
RalfJung, line 129:
```suggestion
    // Returns a pair type representing this wide pointer or `None` if it is thin.
```
Maybe call it `as_pair` then?

spec/lang/representation.md
```diff
@@ -123,19 +125,72 @@ fn encode_ptr<M: Memory>(ptr: ThinPointer<M::Provenance>) -> List<AbstractByte<M
...
+    fn as_tuple_ty<M: Memory>(self) -> Option<Type> {
+        let meta_ty = match self.meta_kind() {
+            PointerMetaKind::None => throw!(),
```
RalfJung, line 132: We *could* put a unit type here and then treat thin and wide pointers more uniformly?

spec/lang/representation.md
```diff
@@ -123,19 +125,72 @@ fn encode_ptr<M: Memory>(ptr: ThinPointer<M::Provenance>) -> List<AbstractByte<M
...
+            PointerMetaKind::ElementCount => Type::Int(IntType { signed: Unsigned, size: M::T::PTR_SIZE }),
+        };
+        let thin_pointer_field = (Offset::ZERO, Type::Ptr(PtrType::Raw { meta_kind: PointerMetaKind::None }));
+        let metadata_field = (Offset::from_bytes(M::T::PTR_SIZE.bytes()).expect("PTR_SIZE is positive"), meta_ty);
```
RalfJung, line 136:
```suggestion
        let metadata_field = (M::T::PTR_SIZE, meta_ty);
```

spec/lang/representation.md
```diff
@@ -123,19 +125,72 @@ fn encode_ptr<M: Memory>(ptr: ThinPointer<M::Provenance>) -> List<AbstractByte<M
...
+        ret(Type::Tuple {
+            fields: list![thin_pointer_field, metadata_field],
+            size: Int::from(2) * M::T::PTR_SIZE,
+            align: Align::from_bytes(M::T::PTR_SIZE.bytes()).expect("PTR_SIZE is aligned"),
```
RalfJung, line 140:
```suggestion
            // This anyway does not matter as we only use this type to encode/decode values.
            align: M::T::PTR_ALIGN,
```

spec/lang/representation.md
```diff
@@ -123,19 +125,72 @@ fn encode_ptr<M: Memory>(ptr: ThinPointer<M::Provenance>) -> List<AbstractByte<M
...
+        })
+    }
+}
+
+impl PointerMeta {
+    fn from<M: Memory>(value: Value<M>) -> Option<PointerMeta> {
```
RalfJung, line 146:
```suggestion
    fn from_value<M: Memory>(value: Value<M>) -> Option<PointerMeta> {
```
For symmetry with `into_value`.

spec/lang/representation.md
```diff
@@ -123,19 +125,72 @@ fn encode_ptr<M: Memory>(ptr: ThinPointer<M::Provenance>) -> List<AbstractByte<M
...
+            ret(Value::Ptr(ptr.widen(None)))
+        } else {
+            let tuple_ty = ptr_type.as_tuple_ty::<M>().expect("ptr_type has metadata");
+            let Value::Tuple(parts) = tuple_ty.decode::<M>(bytes)? else {
+                panic!("as_tuple_ty always returns a tuple type");
+            };
+            let Value::Ptr(ptr) = parts[0] else {
+                panic!("as_tuple_ty always returns tuple with the first field being a thin pointer");
+            };
+            let meta = PointerMeta::from(parts[1]);
```
RalfJung, line 176: Shouldn't this panic if we get a `None`, since as_tuple_ty should always produce a suitable tuple type?

spec/lang/representation.md
```diff
@@ -123,19 +125,72 @@ fn encode_ptr<M: Memory>(ptr: ThinPointer<M::Provenance>) -> List<AbstractByte<M
     
 }
 
+impl PtrType {
+    // Returns a 2-tuple type representing this wide pointer or `None` if it is thin.
+    fn as_tuple_ty<M: Memory>(self) -> Option<Type> {
+        let meta_ty = match self.meta_kind() {
+            PointerMetaKind::None => throw!(),
+            PointerMetaKind::ElementCount => Type::Int(IntType { signed: Unsigned, size: M::T::PTR_SIZE }),
```
RalfJung, line 133: Seems reasonably to factor this into a `ty` method on `PointerMetaKind`.

tooling/minitest/src/tests/slice.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    for_index(-1);
+    for_index(2);
+}
+
+fn make_slice_ptr_tuple<T: TypeConv>() -> Type {
+    tuple_ty(&[(size(0), <*mut T>::get_type()), (size(8), <u64>::get_type())], size(16), align(1))
+}
+
+fn ref_as_slice<T: TypeConv>(f: &mut FunctionBuilder, arr: PlaceExpr, known_len: u64) -> PlaceExpr {
```
RalfJung, line 187: Please put helper methods at the top of the file. And at least `ref_as_slice` is non-trivial enough that it should have a doc comment.

tooling/minimize/tests/pass/slice.rs
```diff
@@ -0,0 +1,27 @@
...
+    print(a[1]);
+    print(a[2]);
+    print(a[3]);
+}
+
+fn main() {
+    let x: &[i32; 5] = &[50, -40, 30, -20, 10];
+
+    // We do not have normal unsizing yet, so we cheat.
+    let ptr = SlicePointerTuple { begin: x[0] as *const i32, len: 5 };
```
RalfJung, line 20: Here you are casting `50` to a pointer, is that your intent? I think you mean `&x[0] as *const i32`, which is better written as `x.as_ptr()`.

That's where the `predict` call came from.

RalfJung (commented):
Looks good, thanks!

I'd say let's remove the minimize parts from this PR, so we can focus on the core MiniRust support.

Before we can add the minimize test we'll need an operand that can extract the metadata from a pointer.

spec/lang/representation.md
Indeximal: Maybe, but keep in mind that there needs to be a separate "base case" for thin pointers anyway, because the Pointer::decode will be called by the Tuple::decode, meaning it is recursive. Preventing this by an explicit panic seemed like the least error prone solution to me.

tooling/minimize/tests/pass/slice.rs
Indeximal: Was not my intent, but it did come up during debugging, but I wrongly assumed you cannot cast i32 to a pointer (only usize). But yeah, did was indeed the problem (altho I cannot use `as_ptr` since it does an implicit unsizing).

spec/lang/representation.md
Indeximal, line 176:
```suggestion
            let meta = PointerMeta::from_value(parts[1]);
```

spec/lang/representation.md
RalfJung: Ah, good point. That should be noted in a comment somewhere.

But also, maybe there is a way here to structure this that does not require any unwraps -- something more like `if let Some(...)` then `if ...`?

tooling/minimize/tests/pass/slice.rs
RalfJung: > I wrongly assumed you cannot cast i32 to a pointer (only usize)

That is a common footgun, indeed. There's a long-standing open issue somewhere to warn against non-usize-int-to-ptr casts, but so far it hasn't been implemented yet.

spec/lang/representation.md
```diff
@@ -99,7 +99,9 @@ When decoding, we have to deal with the possibility of the pointer bytes not all
 this is define to yield a pointer without provenance.
 On the other hand, some pointers are wide pointers which also need to encode their metadata.
 The helpers `decode_ptr` and `encode_ptr` deal with the thin pointers,
```
RalfJung, line 101:
```suggestion
The helpers `decode_ptr` and `encode_ptr` deal with the thin pointers.
```

Indeximal:
Oke I think I address all currently raised issues. If it looks good I will squash it

spec/lang/representation.md
```diff
@@ -96,10 +96,11 @@ This is required to achieve a "monotonicity" with respect to provenance (as disc
 Pointers are significantly more complex to represent than just the integer address.
 For one, they need to encode the provenance.
 When decoding, we have to deal with the possibility of the pointer bytes not all having the same provenance;
-this is define to yield a pointer without provenance.
+this is defined to yield a pointer without provenance.
 On the other hand, some pointers are wide pointers which also need to encode their metadata.
-The helpers `decode_ptr` and `encode_ptr` deal with the thin pointers,
-and the `decode_ptr_meta` helper inferrs the metadata kind from the pointee layout.
+The helpers `decode_ptr` and `encode_ptr` deal with the thin pointers.
```
RalfJung, line 101:
```suggestion
The helpers `decode_ptr` and `encode_ptr` deal with thin pointers.
```

spec/lang/representation.md
```diff
@@ -96,10 +96,11 @@ This is required to achieve a "monotonicity" with respect to provenance (as disc
...
+For wide pointers `PtrType::as_wide_pair` defines the pointer as a pair of a thin pointer,
```
RalfJung, line 102:
```suggestion
For wide pointers, `PtrType::as_wide_pair` defines the pointer as a pair of a thin pointer and some metadata type.
```

spec/lang/representation.md
```diff
@@ -96,10 +96,11 @@ This is required to achieve a "monotonicity" with respect to provenance (as disc
...
+and some metadata type.
```
RalfJung, line 103:
```suggestion
```

spec/lang/step/operators.md
```diff
@@ -238,7 +238,7 @@ impl<M: Memory> Machine<M> {
             Value::Bool(b) => if b { Int::from(1) } else { Int::from(0) },
             Value::Ptr(p) if p.metadata.is_none() => p.thin_pointer.addr,
             // TODO(UnsizedTypes): Prepare wide pointer for rel op: `addr * (MAX_META_VALUE+1) + meta_value`
-            Value::Ptr(_) => panic!("metadata does not exist yet"),
+            Value::Ptr(_) => unimplemented!("pointer metadata comparasion"),
```
RalfJung, line 241:
```suggestion
            Value::Ptr(_) => unimplemented!("comparing pointer with metadata"),
```

spec/lang/well-formed.md
```diff
@@ -51,7 +55,7 @@ impl SizeStrategy {
 impl PointeeInfo {
     fn check_wf<T: Target>(self) -> Result<()> {
         // We do *not* require that size is a multiple of align!
-        ensure_wf(matches!(self.size, SizeStrategy::Sized(_)), "PointeeInfo: Unimplemented size strategy found")?;
+        // TODO(UnsizedTypes): Ensure any future alignment strategy matches the size strategy.
```
RalfJung, line 58: FWIW I am thinking we might have a `SizeAndAlign` type to represent both strategies, since they are coupled. But that's a future discussion :)

spec/lang/well-formed.md
```diff
@@ -759,6 +767,7 @@ impl<M: Memory> Value<M> {
                 };
                 data.check_wf(variant.ty)?;
             }
+            (_, Type::Slice { ..}) => throw_ill_formed!("Value: slices cannot be represented as values"),
```
RalfJung, line 770:
```suggestion
            (_, Type::Slice { .. }) => throw_ill_formed!("Value: slices cannot be represented as values"),
```

RalfJung (commented):
Just a bunch of nits :)

RalfJung:
Awesome! The first actual unsized type in MiniRust :tada:

PR #234, opened 2024-08-30
Add support for `size_of_val` and `align_of_val`

sun-jacobi:


tooling/minimize/src/bb.rs
```diff
@@ -272,7 +272,26 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
 
                 return TerminatorResult { stmts: list!(stmt), terminator };
             }
+            rs::sym::size_of_val => {
+                let destination = self.translate_place(&destination, span);
+
+                let size = self.size_of(intrinsic.args.type_at(0));
```
RalfJung, line 278: For unsized types this won't be right. We can't make this a constant. So it makes little sense to translate this as a constant IMO -- we should add new operands for this in MiniRust proper.

Indeximal:
Yeah, this will be able to use the UnOp from here: https://github.com/minirust/minirust/pull/218/commits/75f77eee1615382a18596d2c3d01ded4706eba0f , altho that commit still needs to be updated & reviewed,

sun-jacobi:
@RalfJung 
Thank you for the comment. I think we can wait for @Indeximal 's commit to be merged :)

sun-jacobi:
@Indeximal 
Hi, I would like to ask if you have a plan to add the `AlignOfVal` :)

Indeximal:
> @Indeximal Hi, I would like to ask if you have a plan to add the `AlignOfVal`:)

@sun-jacobi Yes eventually it will be added aswell. However, first there will be a refactor which merges the internal representation of size and align, which I‘ll start working on Soon (tm).
For now align is still always constant, so if you need to move forward with #234 I could change the align translation to the new operand once I implement it, and you could let it be a constant expression for now.

sun-jacobi:
> > @Indeximal Hi, I would like to ask if you have a plan to add the `AlignOfVal`:)
> 
> @sun-jacobi Yes eventually it will be added aswell. However, first there will be a refactor which merges the internal representation of size and align, which I‘ll start working on Soon (tm). For now align is still always constant, so if you need to move forward with #234 I could change the align translation to the new operand once I implement it, and you could let it be a constant expression for now.

@Indeximal 
Thank you! Please feel free to proceed with your current plan. I’m busy with other things at the moment, so I’m happy to wait :)

sun-jacobi:
@RalfJung 
I see. Recently I may not have time working on this issue, sorry.

@Indeximal Would you like to work on `size_of_val` and `align_of_val`.

RalfJung:
Thanks for the quick reply. :)

PR #236, opened 2024-09-03
Add GetMetadata to allow a minimize slice test

Indeximal:
`RValue::Len` and the `UnOp::PtrMetadata` get translated to the new `GetMetadata` unop, which can infer the return type from the pointer operand type.

This allows slice support in minimize, even if it can't use unsizing or thin pointer casts.

### DIFF
--- a/spec/lang/step/operators.md
+++ b/spec/lang/step/operators.md
@@ -71,7 +71,7 @@ impl<M: Memory> Machine<M> {
 ```rust
 impl<M: Memory> Machine<M> {
     fn eval_un_op(&self, UnOp::GetMetadata: UnOp, (operand, op_ty): (Value<M>, Type)) -> Result<(Value<M>, Type)> {
-        let Value::Ptr(ptr) = operand else { panic!("non pointer get-metadata") };
+        let Value::Ptr(ptr) = operand else { panic!("non-pointer get-metadata") };
 
         if let Some(meta) = ptr.metadata {
             let meta_value = meta.into_value::<M>();
--- a/spec/lang/step/operators.md
+++ b/spec/lang/step/operators.md
@@ -66,6 +66,24 @@ impl<M: Memory> Machine<M> {
 }
 ```
 
+### Wide pointer operators
+
+```rust
+impl<M: Memory> Machine<M> {
+    fn eval_un_op(&self, UnOp::GetMetadata: UnOp, (operand, op_ty): (Value<M>, Type)) -> Result<(Value<M>, Type)> {
+        let Value::Ptr(ptr) = operand else { panic!("non pointer get-metadata") };
+
+        if let Some(meta) = ptr.metadata {
+            let meta_value = meta.into_value::<M>();
+            let meta_ty = meta.meta_kind().ty::<M::T>().expect("meta is not None");
+            ret((meta_value, meta_ty))
+        } else {
+            ret((unit_value::<M>(), unit_type()))
+        }
+    }
+}
+```
+
 ## Binary operators
 
 ```rust
--- a/spec/lang/syntax.md
+++ b/spec/lang/syntax.md
@@ -116,6 +116,8 @@ pub enum UnOp {
     Int(IntUnOp),
     /// A form of cast; the return type is given by the specific cast operation.
     Cast(CastOp),
+    /// Returns the metadata of a pointer as a value. For a thin pointer this is `()`.
+    GetMetadata,
 }
 
 pub enum IntBinOp {
--- a/tooling/minimize/src/rvalue.rs
+++ b/tooling/minimize/src/rvalue.rs
@@ -81,13 +81,14 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
                     (Neg, Type::Int(_)) => build::neg(operand),
                     (Not, Type::Int(_)) => build::bit_not(operand),
                     (Not, Type::Bool) => build::not(operand),
+                    (PtrMetadata, Type::Ptr(_)) => build::get_metadata(operand),
                     (op, _) =>
                         rs::span_bug!(span, "UnOp {op:?} called with unsupported type {ty_smir}."),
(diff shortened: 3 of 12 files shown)


Indeximal:
I tried to call `slice.iter().count()`, but that gave me an unsupported message for `intrinsics::ptr_offset_from_unsigned`. Further iterator testing is to be done, as you suggested.

RalfJung:
I have added support for `ptr_offset_from_unsigned` now, so it should get one step further. :D

Indeximal:
After some debugging I have found the next hurdle, that I will work on soon: PtrToPtr casts with differnt metadata (in particular `GetThinPointer`).

So e.g. this:
```
fn foo(s_ptr: *const [i32]) {
    let e_ptr = s_ptr as *const i32;
}
```
compiled to code with an UB transmute. Now minimize at least catches this.
I think this can be done in a next PR tho, together with `ConstructWidePointer`, so I consider this PR ready to review.

spec/lang/syntax.md
```diff
@@ -110,6 +110,8 @@ pub enum CastOp {
     /// The program is well-formed even if the output type has a different size than the
     /// input type, but the operation is UB in that case.
     Transmute(Type),
+    /// Returns the metadata of a wide pointer as a value.
+    GetMetadata,
```
RalfJung, line 114: This doesn't feel like a cast to me. It can't be done with `as` or anything similar in Rust.

Please make it an `UnOp`.

tooling/minitest/src/tests/wide_ptr.rs
```diff
@@ -0,0 +1,61 @@
+use crate::*;
+use tests::slice::ref_as_slice;
+
+#[test]
+fn get_metadata_slice() {
```
RalfJung, line 5: Please move this into the `slice` file, rather than importing the helper function from there. (That helper should be private.)

tooling/minimize/src/rvalue.rs
```diff
@@ -172,10 +173,17 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
+                match self.translate_ty_smir(ty, span) {
+                    Type::Array { elem: _, count } =>
+                        ValueExpr::Constant(Constant::Int(count), <usize>::get_type()),
+                    Type::Slice { .. } =>
+                        build::get_metadata(build::addr_of(
+                            self.translate_place_smir(place, span),
+                            build::raw_ptr_ty(PointerMetaKind::ElementCount),
+                        )),
+
+                    _ => panic!("RValue::Len only supported for arrays & slices"),
```
RalfJung, line 185:
```suggestion
                    _ => rs::span_bug!(span, "RValue::Len only supported for arrays & slices"),
```
while we are at it, might as well make this a nicer error

tooling/minimize/src/rvalue.rs
```diff
@@ -227,8 +227,24 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
                             operand: GcCow::new(operand),
                         }
                     }
+
+                    smir::CastKind::PtrToPtr => {
+                        let operand_ty = operand.ty(&self.locals_smir).expect("locals are given");
```
RalfJung, line 232: `unwrap()` is enough here, like we do everywhere else.

tooling/minimize/src/rvalue.rs
```diff
@@ -227,8 +227,24 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
+                        let Type::Ptr(old_ptr_ty) = self.translate_ty_smir(operand_ty, span) else {
+                            panic!("ptr to ptr cast on non-pointer");
```
RalfJung, line 234: Please use `rs::span_bug!` for new panics like this.

RalfJung:
Please also rebase, now that the basic slice PR landed. :)

Indeximal:
Fixed, rebased & squashed. Offer your final nitpicks, otherwise I think this is ready.

spec/lang/step/operators.md
```diff
@@ -66,6 +66,24 @@ impl<M: Memory> Machine<M> {
...
+### Wide pointer operators
+
+```rust
+impl<M: Memory> Machine<M> {
+    fn eval_un_op(&self, UnOp::GetMetadata: UnOp, (operand, op_ty): (Value<M>, Type)) -> Result<(Value<M>, Type)> {
+        let Value::Ptr(ptr) = operand else { panic!("non pointer get-metadata") };
+
+        if let Some(meta) = ptr.metadata {
+            let meta_value = meta.into_value::<M>();
+            let meta_ty = meta.meta_kind().ty::<M::T>().expect("meta is not None");
```
RalfJung, line 78: This is where it'd be nicer if `MetaKind` would be an `Option`... then we wouldn't need an unwrap/expect here...

tooling/minimize/src/rvalue.rs
```diff
@@ -170,12 +171,18 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
             smir::Rvalue::Len(place) => {
-                // as slices are unsupported as of now, we only need to care for arrays.
                 let ty = place.ty(&self.locals_smir).unwrap();
-                let Type::Array { elem: _, count } = self.translate_ty_smir(ty, span) else {
-                    panic!()
-                };
-                ValueExpr::Constant(Constant::Int(count), <usize>::get_type())
+                match self.translate_ty_smir(ty, span) {
+                    Type::Array { elem: _, count } =>
+                        ValueExpr::Constant(Constant::Int(count), <usize>::get_type()),
```
RalfJung, line 177:
```suggestion
                        // FIXME: still evaluate the place -- it might have UB after all.
                        ValueExpr::Constant(Constant::Int(count), <usize>::get_type()),
```

RalfJung (commented):
Looks great, thanks!

PR #237, opened 2024-09-06
Add operations to construct and destruct wide pointers

Indeximal:
In particular, `GetThinPointer` is the pairing to the first part of `ptr::to_raw_parts`.
`ConstructWidePointer` is analogous to `core::ptr::from_raw_parts, altho it also works on references.

This PR is stacked on #236, the first commit being from there.

A discussion question for now is maybe, whether we want to allow the construction of a thin pointer by giving a unit value as metadata. Rust does it this way and it would be easy to also allow this. However, it seems unnecessary and kind of contradicts the current `PointerMetaKind::ty()` function which returns `None` and not the unit type, which was a decision to help prevent accidental recursion of the en/decode functions.

### DIFF
--- a/spec/lang/step/operators.md
+++ b/spec/lang/step/operators.md
@@ -70,8 +70,16 @@ impl<M: Memory> Machine<M> {
 
 ```rust
 impl<M: Memory> Machine<M> {
+    fn eval_un_op(&self, UnOp::GetThinPointer: UnOp, (operand, op_ty): (Value<M>, Type)) -> Result<(Value<M>, Type)> {
+        let Value::Ptr(ptr) = operand else { panic!("non-pointer GetThinPointer") };
+
+        let thin_ptr = Pointer { metadata: None, ..ptr };
+        let thin_ptr_ty = PtrType::Raw { meta_kind: PointerMetaKind::None };
+        ret((Value::Ptr(thin_ptr), Type::Ptr(thin_ptr_ty)))
+    }
+    
     fn eval_un_op(&self, UnOp::GetMetadata: UnOp, (operand, op_ty): (Value<M>, Type)) -> Result<(Value<M>, Type)> {
-        let Value::Ptr(ptr) = operand else { panic!("non-pointer get-metadata") };
+        let Value::Ptr(ptr) = operand else { panic!("non-pointer GetMetadata") };
 
         if let Some(meta) = ptr.metadata {
             let meta_value = meta.into_value::<M>();
@@ -360,6 +368,28 @@ impl<M: Memory> Machine<M> {
 
         let isize_int = IntType { signed: Signed, size: M::T::PTR_SIZE };
         ret((Value::Int(distance), Type::Int(isize_int)))
+    }
+}
+```
+
+### Wide pointer construction
+
+```rust
+impl<M: Memory> Machine<M> {
+    fn eval_bin_op(
+        &self,
+        BinOp::ConstructWidePointer(ptr_ty): BinOp,
+        (left, l_ty): (Value<M>, Type),
+        (right, _r_ty): (Value<M>, Type)
+    ) -> Result<(Value<M>, Type)> {
+        let Value::Ptr(Pointer { thin_pointer, metadata: None }) = left else {
+            panic!("non-thin-pointer left input to `ConstructWidePointer`")
+        };
+        let metadata = PointerMeta::from_value::<M>(right);
+        let wide_ptr = Value::Ptr(Pointer { thin_pointer, metadata });
+
+        wide_ptr.check_wf(Type::Ptr(ptr_ty)).expect("sanity check: constructed wide pointer is well-formed");
+        ret((wide_ptr, Type::Ptr(ptr_ty)))
     }
 }
 ```
--- a/spec/lang/well-formed.md
+++ b/spec/lang/well-formed.md
@@ -353,6 +353,10 @@ impl ValueExpr {
                             }
                         }
                     }
+                    GetThinPointer => {
+                        ensure_wf(matches!(operand, Type::Ptr(_)), "UnOp::GetThinPointer: invalid operand: not a pointer")?;
+                        Type::Ptr(PtrType::Raw { meta_kind: PointerMetaKind::None })
+                    }
                     GetMetadata => {
                         let Type::Ptr(ptr_ty) = operand else {
                             throw_ill_formed!("UnOp::GetMetadata: invalid operand: not a pointer");
@@ -426,6 +430,22 @@ impl ValueExpr {
                         }
                         let isize_int = IntType { signed: Signed, size: T::PTR_SIZE };
                         Type::Int(isize_int)
+                    }
+                    ConstructWidePointer(ptr_ty) => {
+                        let Type::Ptr(thin_ptr_ty) = left else {
+                            throw_ill_formed!("BinOp::ConstructWidePointer: invalid left type: not a pointer");
+                        };
+                        if thin_ptr_ty.meta_kind() != PointerMetaKind::None {
+                            throw_ill_formed!("BinOp::ConstructWidePointer: invalid left type: not a thin pointer");
+                        }
+
+                        // A thin pointer can also be the target type, with unit metadata.
+                        let meta_ty = ptr_ty.meta_kind().ty::<T>().unwrap_or_else(unit_type);
+                        if right != meta_ty {
+                            throw_ill_formed!("BinOp::ConstructWidePointer: invalid right type: not metadata of target");
+                        }
+
+                        Type::Ptr(ptr_ty)
                     }
                 }
             }
--- a/tooling/minimize/src/rvalue.rs
+++ b/tooling/minimize/src/rvalue.rs
@@ -180,6 +180,20 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
                         ));
                         ValueExpr::Variant { discriminant, data, enum_ty: ty }
                     }
+                    Type::Ptr(PtrType::Raw { .. }) => {
+                        if operands.len() != 2 {
+                            rs::span_bug!(
+                                span,
+                                "Aggregating a pointer with {} operands",
+                                operands.len()
+                            );
+                        }
+                        let ptr = self.translate_operand_smir(&operands[0], span);
+                        let meta = self.translate_operand_smir(&operands[1], span);
+
+                        // We rely on MIR being well formed and matching our type-meta-pairings for this to be WF.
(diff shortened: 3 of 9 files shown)


Indeximal:
I have now also implemented unsizing coercion and with that more or less finished the scope of this PR, making this ready for a first review

spec/lang/step/operators.md
```diff
@@ -66,6 +66,32 @@ impl<M: Memory> Machine<M> {
 }
 ```
 
+### Wide pointer operators
+
+```rust
+impl<M: Memory> Machine<M> {
+    fn eval_un_op(&self, UnOp::GetThinPointer: UnOp, (operand, op_ty): (Value<M>, Type)) -> Result<(Value<M>, Type)> {
+        let Value::Ptr(ptr) = operand else { panic!("non pointer get-thinpointer") };
```
RalfJung, line 74:
```suggestion
        let Value::Ptr(ptr) = operand else { panic!("non-pointer GetThinPointer") };
```

spec/lang/well-formed.md
```diff
@@ -431,6 +431,28 @@ impl ValueExpr {
...
                     }
+                    ConstructWidePointer(ptr_ty) => {
+                        let Type::Ptr(thin_ptr_ty) = left else {
+                            throw_ill_formed!("BinOp::ConstructWidePointer: invalid left type: not a pointer");
+                        };
+                        if thin_ptr_ty.meta_kind() != PointerMetaKind::None {
+                            throw_ill_formed!("BinOp::ConstructWidePointer: invalid left type: not a thin pointer");
+                        }
+
+                        // We could also allow constructing a thin pointer with the unit metadata here, rust does,
```
RalfJung, line 442: If Rust allows this and we don't that means Minimize has to do more work. Also this seems inconsistent with `GetMetadata` supporting thin pointers. So I think I'd prefer to support thin pointers here.

spec/lang/well-formed.md
```diff
@@ -431,6 +431,28 @@ impl ValueExpr {
...
+                        let Some(meta_ty) = ptr_ty.meta_kind().ty::<T>() else {
+                            throw_ill_formed!("BinOp::ConstructWidePointer: invalid target type: not a wide pointer");
+                        };
+                        if right != meta_ty {
+                            throw_ill_formed!("BinOp::ConstructWidePointer: invalid right type: not metadata of target");
+                        }
+
+                        // Question: Firstly, this could be a function pointer, but we don't seem to care about that in other places.
+                        // Secondly (outside my scope, just interested), the "user" now has control of the aliasing flags of the PointeeInfo,
+                        // is this not a problem?
```
RalfJung, line 453: They already have the same control with other operations, so that seems fine.

tooling/minitest/src/tests/wide_ptr.rs
```diff
@@ -34,6 +34,98 @@ fn get_thin_non_ptr_ill_formed() {
...
+        p.finish_function(f)
+    };
+
+    let p = p.finish_program(f);
+    assert_ill_formed::<BasicMem>(
+        p,
+        "BinOp::ConstructWidePointer: invalid left type: not a thin pointer",
+    );
+}
+#[test]
```
RalfJung, line 82: Please keep the usual empty line between tests

tooling/minimize/src/rvalue.rs
```diff
@@ -235,14 +235,18 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
+                            build::transmute(
+                                build::get_thin_pointer(operand),
+                                Type::Ptr(new_ptr_ty),
+                            )
                         } else {
-                            // TODO(UnsizedTypes): Implement this for wide to thin pointer casts
                             rs::span_bug!(
                                 span,
-                                "Unimplemented: casting pointers with different metadata"
+                                "PtrToPtr cast with wide target that differes from source"
```
RalfJung, line 249: Please add the old and new type to the error message

tooling/minimize/src/rvalue.rs
```diff
@@ -235,14 +235,18 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
                             rs::span_bug!(span, "ptr to ptr cast to non-pointer");
                         };
+                        let operand = self.translate_operand_smir(operand, span);
                         if old_ptr_ty.meta_kind() == new_ptr_ty.meta_kind() {
-                            let operand = self.translate_operand_smir(operand, span);
                             build::transmute(operand, Type::Ptr(new_ptr_ty))
+                        } else if new_ptr_ty.meta_kind() == PointerMetaKind::None {
+                            build::transmute(
+                                build::get_thin_pointer(operand),
+                                Type::Ptr(new_ptr_ty),
```
RalfJung, line 244: I think we only allow these casts on raw pointers, i.e. ``new_ptr_ty`` must be `Raw` and therefore the `transmute` here should be unnecessary. Can you make this an assertion?

tooling/minimize/tests/pass/slice.rs
```diff
@@ -28,4 +28,5 @@ fn main() {
 
     assert_some_elements(slice);
     assert!(slice.len() == 5);
+    assert!(slice.iter().count() == 5);
```
RalfJung, line 14: Does actually iterating with a for loop work, too? :)

tooling/minimize/src/rvalue.rs
```diff
@@ -250,6 +250,32 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
+                                    );
+                                }
+                                // FIXME(UnsizedTypes): do we support 32bit architectures here? If so, `usize` is probably wrong.
+                                build::construct_wide_pointer(
+                                    operand,
+                                    build::const_int_typed::<usize>(count),
+                                    Type::Ptr(new_ptr_ty),
+                                )
+                            }
+                            _ => rs::span_bug!(span, "Unsupported unsizing coercion"),
```
RalfJung, line 276: Please say in the error from which type to which type the unsupported coercion goes.

tooling/minimize/src/ty.rs
```diff
@@ -112,6 +112,18 @@ impl<'tcx> Ctxt<'tcx> {
...
 
+    /// Returns the minirust type of the pointee, given this `ty` is some form of pointer.
+    pub fn translate_pointee_ty_smir(&mut self, ty: smir::Ty, span: rs::Span) -> Type {
+        let ty = smir::internal(self.tcx, ty);
+        let pointee_ty = match ty.kind() {
+            rs::TyKind::Adt(adt_def, _) if adt_def.is_box() => ty.boxed_ty(),
+            rs::TyKind::Ref(_, ty, _mutbl) => *ty,
+            rs::TyKind::RawPtr(ty, _mutbl) => *ty,
+            x => rs::span_bug!(span, "TyKind not a pointer type: {x:?}"),
+        };
```
RalfJung, line 123: This is just `ty.builtin_deref(true).unwrap()`.

I am not sure if this needs a dedicated helper function.

RalfJung (commented):
Very nice :-)

tooling/minimize/tests/pass/slice.rs
```diff
@@ -0,0 +1,15 @@
+// Pass slice reference accross function boundaries
+pub fn assert_some_elements(a: &[i32]) {
+    assert!(a[1] == -40);
+    assert!(a[2] == 30);
+    assert!(a[3] == -20);
+}
+
+fn main() {
+    let x: [i32; 5] = [50, -40, 30, -20, 10];
+    let slice: &[i32] = &x;
```
RalfJung, line 10: Can you also add a variant that uses `slice::from_raw_parts`? Or do you want to make that a separate PR?
(It requires another new case in minimize to support `aggregate_raw_ptr`.)

tooling/minimize/src/ty.rs
Indeximal: Yeah sure, didn't know about this of course

tooling/minimize/tests/pass/slice.rs
Indeximal: Yessss

Indeximal:
When using subslicing to test the aggreate_raw_ptr I run into the problem with no MIR available...: "`core[5325]::slice::index::slice_start_index_len_fail` does not have a "optimized_mir".

When invoking `from_raw_parts` directly, I get an error that the rust intrinsic `ctpop` is missing.

tooling/minimize/src/rvalue.rs
```diff
@@ -180,6 +180,15 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
                         ));
                         ValueExpr::Variant { discriminant, data, enum_ty: ty }
                     }
+                    Type::Ptr(PtrType::Raw { .. }) => {
+                        let ptr = self.translate_operand_smir(&operands[0], span);
+                        let meta = self.translate_operand_smir(&operands[1], span);
+
+                        // Question: I mean I could check here that the meta maches the the target type,
+                        // but then I need the `Target` generic,
+                        // and don't think all places in minimize return well-formed expresessions in all cases.
```
RalfJung, line 189: You can generally rely on the MIR being well-formed.

tooling/minimize/src/rvalue.rs
```diff
@@ -180,6 +180,15 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
                         ));
                         ValueExpr::Variant { discriminant, data, enum_ty: ty }
                     }
+                    Type::Ptr(PtrType::Raw { .. }) => {
+                        let ptr = self.translate_operand_smir(&operands[0], span);
```
RalfJung, line 184: Please add an assertion that `operands.len() == 2`, just to document that we are never omitting an operand here.

tooling/minimize/src/rvalue.rs
```diff
@@ -247,23 +256,64 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
+                        let operand = self.translate_operand_smir(operand, span);
+                        match (old_pointee_ty, new_pointee_ty) {
+                            (Type::Array { count, elem: a_elem }, Type::Slice { elem: s_elem }) => {
+                                if a_elem != s_elem {
+                                    rs::span_bug!(
+                                        span,
+                                        "Unsizing to slice with different element type"
+                                    );
+                                }
+                                // FIXME(UnsizedTypes): do we support 32bit architectures here? If so, `usize` is probably wrong.
```
RalfJung, line 303: `usize` will use the "target" usize, so this is fine.

RalfJung:
Looks good, please squash. :)

PR #238, opened 2024-09-14
Add SizeOfVal operation

Indeximal:
This is a refactor of previous work, adding the `UnOp::SizeOfVal` to MiniRust. With that, at least half of #234 can be properly implemented.

Compared to the first iteration, this requires a safe pointer, in Raw pointers, the information is not sufficient to compute the size. And function pointers are currently not supported, altho, Rust does and returns 0 for the size of a function. So this is a discussion to be had.

Slices (which are the most interesting) are untested, because this PR isn't stacked on the lastet #237 and constructing slices is a pain before that.

### DIFF
--- a/spec/lang/step/operators.md
+++ b/spec/lang/step/operators.md
@@ -92,6 +92,20 @@ impl<M: Memory> Machine<M> {
 }
 ```
 
+### Size of value
+
+```rust
+impl<M: Memory> Machine<M> {
+    fn eval_un_op(&self, UnOp::SizeOfVal: UnOp, (operand, op_ty): (Value<M>, Type)) -> Result<(Value<M>, Type)> {
+        let Type::Ptr(PtrType::Ref { pointee, .. }) = op_ty else { panic!("non-reference input to SizeOfVal") };
+        let Value::Ptr(ptr) = operand else { panic!("non-pointer input to SizeOfVal") };
+
+        let size = pointee.size.compute(ptr.metadata);
+        ret((Value::Int(size.bytes()), Type::Int(IntType { signed: Unsigned, size: M::T::PTR_SIZE })))
+    }
+}
+```
+
 ## Binary operators
 
 ```rust
--- a/spec/lang/syntax.md
+++ b/spec/lang/syntax.md
@@ -121,6 +121,9 @@ pub enum UnOp {
     GetThinPointer,
     /// Returns the metadata of a pointer as a value. For a thin pointer this is `()`.
     GetMetadata,
+    /// Returns the dynamic size of the pointee behind the operand.
+    /// The operand must be a reference.
+    SizeOfVal,
 }
 
 pub enum IntBinOp {
--- a/spec/lang/well-formed.md
+++ b/spec/lang/well-formed.md
@@ -364,6 +364,14 @@ impl ValueExpr {
                         // If the pointer does not have metadata, this will still be well-formed but return the unit type.
                         ptr_ty.meta_kind().ty::<T>().unwrap_or_else(unit_type)
                     }
+                    SizeOfVal => {
+                        // TODO(UnsizedTypes): Support raw pointers as well.
+                        ensure_wf(
+                            matches!(operand, Type::Ptr(PtrType::Ref { .. })),
+                            "UnOp::SizeOfVal: invalid operand: not a reference"
+                        )?;
+                        Type::Int(IntType { signed: Unsigned, size: T::PTR_SIZE })
+                    }
                 }
             }
             BinOp { operator, left, right } => {
--- a/tooling/minitest/src/tests/size_of_val.rs
+++ b/tooling/minitest/src/tests/size_of_val.rs
@@ -1,0 +1,267 @@
+use crate::*;
+use miniutil::DefaultTarget;
+
+/// Helper which builds code to assert that size_of_val with given type gives the expected size.
+fn assume_size_of_ty(f: &mut FunctionBuilder, size: usize, ty: Type) {
+    // This is now kind of ugly, but there is no way to get a minirust reference type for a given minirust type anymore.
+    let pointee = PointeeInfo {
+        size: ty.size::<DefaultTarget>(),
+        align: ty.align::<DefaultTarget>(),
+        inhabited: true,
+        freeze: false,
+        unpin: false,
+    };
+    let i = f.declare_local_with_ty(ty);
+    f.storage_live(i);
+    f.assume(eq(size_of_val(addr_of(i, ref_ty(pointee))), const_int(size)));
+}
+
+/// Helper to call [`assume_size_of_ty`] when [`TypeConv`] is available.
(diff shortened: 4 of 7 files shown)


RalfJung:
> function pointers are currently not supported, altho, Rust does and returns 0 for the size of a function

size_of_val cannot be called on function pointers, so I don't quite know what you are talking about.

spec/lang/syntax.md
```diff
@@ -118,6 +118,8 @@ pub enum UnOp {
     Cast(CastOp),
     /// Returns the metadata of a pointer as a value. For a thin pointer this is `()`.
     GetMetadata,
+    /// Getting the dynamic size of the pointee behind the operand.
```
RalfJung, line 121:
```suggestion
    /// Returns the dynamic size of the pointee behind the operand.
    /// The operand must be a reference.
```

spec/lang/step/operators.md
```diff
@@ -84,6 +84,21 @@ impl<M: Memory> Machine<M> {
 }
 ```
 
+### Size of value
+
+```rust
+impl<M: Memory> Machine<M> {
+    fn eval_un_op(&self, UnOp::SizeOfVal: UnOp, (operand, op_ty): (Value<M>, Type)) -> Result<(Value<M>, Type)> {
+        let Type::Ptr(ptr_ty) = op_ty else { panic!("non-pointer input to SizeOfVal") };
+        let Some(pointee) = ptr_ty.safe_pointee() else { panic!("non-safe-pointer input to SizeOfVal") };
```
RalfJung, line 93: Seems unnecessary to support `Box` here, so you can just require a reference.

tooling/minitest/src/tests/size_of_val.rs
```diff
@@ -0,0 +1,271 @@
...
+
+        f.exit();
+        p.finish_function(f)
+    };
+
+    let p = p.finish_program(f);
+    assert_stop::<BasicMem>(p);
+}
+
+/// Tests size_of_val for pointers.
```
RalfJung, line 76: `usize` should also be tested, either here or with the integers.

tooling/minitest/src/tests/size_of_val.rs
```diff
@@ -0,0 +1,271 @@
...
+        f.assume(eq(size_of_val(addr_of(x, <*const u32>::get_type())), const_int(4_usize)));
+        f.exit();
+        p.finish_function(f)
+    };
+
+    let p = p.finish_program(f);
+    assert_ill_formed::<BasicMem>(p, "UnOp::SizeOfVal: invalid operand: not a safe pointer");
+}
+
+/// size_of_val for function pointers is not supported atm.
```
RalfJung, line 245:
```suggestion
/// size_of_val for function pointers makes little sense, and is hence rejected.
```

Indeximal:
> > function pointers are currently not supported, altho, Rust does and returns 0 for the size of a function
> 
> size_of_val cannot be called on function pointers, so I don't quite know what you are talking about.

Disagree: <https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a7e2d9817ff3daf88f81e695ea683d4c>

tooling/minitest/src/tests/size_of_val.rs
```diff
@@ -0,0 +1,271 @@
...
+        );
+        f.exit();
+        p.finish_function(f)
+    };
+
+    let p = p.finish_program(f);
+    assert_stop::<BasicMem>(p);
+}
+
+/// Tests size_of_val for a simple enum type.
```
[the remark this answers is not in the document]
Indeximal: Yeah idk, I wrote these tests over a month ago and just found it equally unnecessary to delete it now as to keep it. Should I remove it again?

RalfJung:
You are calling it on a *reference to a function pointer*. Actually you are calling it on a reference to a "function item type". And yes function item types are zero-sized.

But you can't call size_of_val on a function pointer (or function item type).

tooling/minitest/src/tests/size_of_val.rs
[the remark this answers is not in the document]
RalfJung: Please delete the enum test.

Indeximal:
> You are calling it on a _reference to a function pointer_. Actually you are calling it on a reference to a "function item type". And yes function item types are zero-sized.
> 
> But you can't call size_of_val on a function pointer (or function item type).

Hm okay, does this mean our `Type::Ptr(PtrType::FnPtr)` is our representation of a "function item type" or do we not model this feature? Then what I passed would be a `Type::Ptr(PtrType::Ref { .. })` with a pointee being a pointer, and all pointers are 8 or 16 bytes in MiniRust, as opposed to the 0 which rustc gives.

Indeximal:
I adjusted the PR to your feedback and stacked it on the now ready #237. This time I tried to do it with a merge commit instead of a rebase in a effort to make iterative reviewing easier. However, I am unsure whether this is better and whether this will just lead to problems when squashing eventually.

RalfJung:
Yeah this is one of the situations where a rebase and force push is needed. If you keep revisions (in reply to reviews) in new commits, rather than amending existing commits, it's still usually possible to see what changed since the last review. If that's annoying due to conflicts, then it's okay to squash -- that should be sufficiently rare, hopefully.

RalfJung:
> Hm okay, does this mean our Type::Ptr(PtrType::FnPtr) is our representation of a "function item type" or do we not model this feature? Then what I passed would be a Type::Ptr(PtrType::Ref { .. }) with a pointee being a pointer, and all pointers are 8 or 16 bytes in MiniRust, as opposed to the 0 which rustc gives.

Our `FnPtr` corresponds to Rust's `fn()` -- function pointers, not "function items". Note however that `size_of_val` cannot be called on a function pointer! It can only be called on `&T`. It *can* be called on `&fn()` (that's not what your example does, but it is possible). But `&fn()` is a MiniRust `TypePtr(PtrType::Ref)` where the pointee type is a function pointer -- it's a reference to a function pointer.

When you do `let x: fn() = foo;`, this insert a coercion from the function item type for `foo` to the function pointer type `fn()`. So in your example you want to do something like `size_of_val(&(foo as fn()))` to see the answer `8`, for a function pointer.

Indeximal:
Okay, thanks for your explaination

spec/lang/well-formed.md
```diff
@@ -364,6 +364,14 @@ impl ValueExpr {
                         // If the pointer does not have metadata, this will still be well-formed but return the unit type.
                         ptr_ty.meta_kind().ty::<T>().unwrap_or_else(unit_type)
                     }
+                    SizeOfVal => {
+                        // TODO(UnsizedTypes): Support for raw pointers aswell.
```
RalfJung, line 368:
```suggestion
                        // TODO(UnsizedTypes): Support raw pointers as well.
```

RalfJung (commented):
Just one last comment nit. :) Then please squash the commits.

PR #239, opened 2024-09-15
Add Trait Object Support

Indeximal:
This PR contains implementations of Trait Objects, including VTables and dynamic dispatch.

This PR doesn't include minimize support.

### DIFF
--- a/spec/lang/representation.md
+++ b/spec/lang/representation.md
@@ -160,8 +160,9 @@ impl<Provenance> PointerMeta<Provenance> {
     fn from_value<M: Memory<Provenance=Provenance>>(value: Value<M>) -> Option<PointerMeta<Provenance>> {
         match value {
             Value::Int(count) => Some(PointerMeta::ElementCount(count)),
-            Value::Ptr(ptr) => Some(PointerMeta::VTablePointer(ptr.thin_pointer)),
-            _ => None,
+            Value::Ptr(ptr) if ptr.metadata.is_none() => Some(PointerMeta::VTablePointer(ptr.thin_pointer)),
+            Value::Tuple(fields) if fields.is_empty() => None,
+            _ => panic!("PointerMeta::from_value called with invalid value"),
         }
     }
 
@@ -456,26 +457,6 @@ impl<M: Memory> Machine<M> {
     fn check_pointer(&self, ptr: Pointer<M::Provenance>, ptr_ty: PtrType) -> Result {
         ensure_else_ub(ptr.thin_pointer.addr.in_bounds(Unsigned, M::T::PTR_SIZE), "Value::Ptr: pointer out-of-bounds")?;
 
-        // Safe pointer, i.e. references, boxes
-        if let Some(pointee) = ptr_ty.safe_pointee() {
-            let size = pointee.layout.compute_size(ptr.metadata, self.vtable_lookup());
-            let align = pointee.layout.compute_align(ptr.metadata, self.vtable_lookup());
-            // The total size of slices must be at most `isize::MAX`.
-            ensure_else_ub(size.bytes().in_bounds(Signed, M::T::PTR_SIZE), "Value::Ptr: total size exeeds isize::MAX")?;
-
-            // Safe addresses need to be non-null, aligned, dereferenceable, and not point to an uninhabited type.
-            // (Think: uninhabited types have impossible alignment.)
-            ensure_else_ub(ptr.thin_pointer.addr != 0, "Value::Ptr: null safe pointer")?;
-            ensure_else_ub(align.is_aligned(ptr.thin_pointer.addr), "Value::Ptr: unaligned safe pointer")?;
-            ensure_else_ub(pointee.inhabited, "Value::Ptr: safe pointer to uninhabited type")?;
-            ensure_else_ub(
-                self.mem.dereferenceable(ptr.thin_pointer, size).is_ok(),
-                "Value::Ptr: non-dereferenceable safe pointer"
-            )?;
-
-            // However, we do not care about the data stored wherever this pointer points to.
-        }
-
         match (ptr.metadata, ptr_ty.meta_kind()) {
             (None, PointerMetaKind::None) => {}
             (Some(PointerMeta::ElementCount(num)), PointerMetaKind::ElementCount) =>
@@ -490,6 +471,26 @@ impl<M: Memory> Machine<M> {
             }
             _ => throw_ub!("Value::Ptr: invalid metadata"),
         };
+
+        // Safe pointer, i.e. references, boxes
+        if let Some(pointee) = ptr_ty.safe_pointee() {
+            let size = pointee.layout.compute_size(ptr.metadata, self.vtable_lookup());
+            let align = pointee.layout.compute_align(ptr.metadata, self.vtable_lookup());
+            // The total size must be at most `isize::MAX`.
+            ensure_else_ub(size.bytes().in_bounds(Signed, M::T::PTR_SIZE), "Value::Ptr: total size exeeds isize::MAX")?;
+
+            // Safe pointers need to be non-null, aligned, dereferenceable, and not point to an uninhabited type.
+            // (Think: uninhabited types have impossible alignment.)
+            ensure_else_ub(ptr.thin_pointer.addr != 0, "Value::Ptr: null safe pointer")?;
+            ensure_else_ub(align.is_aligned(ptr.thin_pointer.addr), "Value::Ptr: unaligned safe pointer")?;
+            ensure_else_ub(pointee.inhabited, "Value::Ptr: safe pointer to uninhabited type")?;
+            ensure_else_ub(
+                self.mem.dereferenceable(ptr.thin_pointer, size).is_ok(),
+                "Value::Ptr: non-dereferenceable safe pointer"
+            )?;
+
+            // However, we do not care about the data stored wherever this pointer points to.
+        }
 
         Ok(())
     }
--- a/spec/lang/representation.md
+++ b/spec/lang/representation.md
@@ -156,20 +156,27 @@ impl PtrType {
     }
 }
 
-impl<Provenance> PointerMeta<Provenance> {
-    fn from_value<M: Memory<Provenance=Provenance>>(value: Value<M>) -> Option<PointerMeta<Provenance>> {
-        match value {
-            Value::Int(count) => Some(PointerMeta::ElementCount(count)),
-            Value::Ptr(ptr) if ptr.metadata.is_none() => Some(PointerMeta::VTablePointer(ptr.thin_pointer)),
-            Value::Tuple(fields) if fields.is_empty() => None,
-            _ => panic!("PointerMeta::from_value called with invalid value"),
+impl PointerMetaKind {
+    /// Decodes a value to metadata.
+    /// The spec will only call this with values which are well formed for `self.ty()`,
+    /// but this may return ill-formed metadata (as defined by `Machine::check_pointer_metadata`), thus needs to be checked.
+    fn decode_value<M: Memory>(self, value: Value<M>) -> Option<PointerMeta<M::Provenance>> {
+        match (self, value) {
+            (PointerMetaKind::ElementCount, Value::Int(count)) => Some(PointerMeta::ElementCount(count)),
+            (PointerMetaKind::VTablePointer(_), Value::Ptr(ptr)) if ptr.metadata.is_none() => Some(PointerMeta::VTablePointer(ptr.thin_pointer)),
+            (PointerMetaKind::None, Value::Tuple(fields)) if fields.is_empty() => None,
+            _ => panic!("PointerMeta::decode_value called with invalid value"),
         }
     }
 
-    fn into_value<M: Memory<Provenance=Provenance>>(self) -> Value<M> {
-        match self {
-            PointerMeta::ElementCount(count) => Value::Int(count),
-            PointerMeta::VTablePointer(ptr) => Value::Ptr(ptr.widen(None)),
+    /// Encodes metadata as a value.
+    /// The spec ensures this is only called with well-formed metadata (as defined by `Machine::check_pointer_metadata`).
+    fn encode_as_value<M: Memory>(self, meta: Option<PointerMeta<M::Provenance>>) -> Value<M> {
+        match (self, meta) {
+            (PointerMetaKind::ElementCount, Some(PointerMeta::ElementCount(count))) => Value::Int(count),
+            (PointerMetaKind::VTablePointer(_), Some(PointerMeta::VTablePointer(ptr))) => Value::Ptr(ptr.widen(None)),
+            (PointerMetaKind::None, None) => unit_value(),
+            _ => panic!("PointerMeta::encode_as_value called with invalid value"),
         }
     }
 }
@@ -184,7 +191,8 @@ impl Type {
             let Value::Ptr(ptr) = parts[0] else {
                 panic!("as_wide_pair always returns tuple with the first field being a thin pointer");
             };
-            let meta = PointerMeta::from_value(parts[1]);
+            // This metadata might not be well-formed, but we are allowed to return ill-formed pointers here.
+            let meta = ptr_type.meta_kind().decode_value(parts[1]);
             assert!(meta.is_some(), "as_wide_pair always returns a suitable metadata type");
             ret(Value::Ptr(ptr.thin_pointer.widen(meta)))
         } else {
@@ -198,7 +206,7 @@ impl Type {
 
         if let Some(pair_ty) = ptr_type.as_wide_pair::<M::T>() {
             let thin_ptr_value = Value::Ptr(ptr.thin_pointer.widen(None));
-            let meta_data_value = ptr.metadata.expect("val is WF for ptr_type and it has metadata").into_value::<M>();
+            let meta_data_value = ptr_type.meta_kind().encode_as_value::<M>(ptr.metadata);
             let tuple = Value::Tuple(list![thin_ptr_value, meta_data_value]);
 
             // This will recursively call this encode function again, but with a thin pointer type.
@@ -453,11 +461,9 @@ fn ensure_else_ub(b: bool, msg: &str) -> Result<()> {
 }
 
 impl<M: Memory> Machine<M> {
-    /// Checks that a pointer is well-formed.
-    fn check_pointer(&self, ptr: Pointer<M::Provenance>, ptr_ty: PtrType) -> Result {
-        ensure_else_ub(ptr.thin_pointer.addr.in_bounds(Unsigned, M::T::PTR_SIZE), "Value::Ptr: pointer out-of-bounds")?;
-
-        match (ptr.metadata, ptr_ty.meta_kind()) {
+    /// Defines well-formedness for pointer metadata for a given kind.
+    fn check_pointer_metadata(&self, meta: Option<PointerMeta<M::Provenance>>, kind: PointerMetaKind) -> Result {
+        match (meta, kind) {
             (None, PointerMetaKind::None) => {}
             (Some(PointerMeta::ElementCount(num)), PointerMetaKind::ElementCount) =>
                 self.check_value(Value::Int(num), Type::Int(IntType::usize_ty::<M::T>()))?,
@@ -471,6 +477,15 @@ impl<M: Memory> Machine<M> {
             }
             _ => throw_ub!("Value::Ptr: invalid metadata"),
         };
+
+        Ok(())
+    }
+
+    /// Checks that a pointer is well-formed.
+    fn check_pointer(&self, ptr: Pointer<M::Provenance>, ptr_ty: PtrType) -> Result {
+        ensure_else_ub(ptr.thin_pointer.addr.in_bounds(Unsigned, M::T::PTR_SIZE), "Value::Ptr: pointer out-of-bounds")?;
+
+        self.check_pointer_metadata(ptr.metadata, ptr_ty.meta_kind())?;
 
         // Safe pointer, i.e. references, boxes
         if let Some(pointee) = ptr_ty.safe_pointee() {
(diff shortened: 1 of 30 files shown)


tooling/minimize/src/bb.rs
```diff
@@ -375,6 +375,8 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
             return self.translate_rs_intrinsic(instance, args, destination, target, span);
         }
 
+        // I think `instance.def` will be `InstanceKind::Virtual` for trait object calls ?
```
bjorn3, line 378: Correct

Indeximal:
I have rewritten the dynamic dispatch logic to use a `VTableLookup` expression and added a test as to how this would be used. The whole thing compiles, but there is one major todo, and that is converting between `VTableNames` and `PtrType::VTablePtr` addresses.
I am unsure whether we really want to move forward with piping vtable allocations through the decode/encode functions, as they are also used in memory methods which do also not have access to this information. Maybe storing a pointer value in the metadata is preferable anyway. Do you have any other ideas?

RalfJung:
> Maybe storing a pointer value in the metadata is preferable anyway. Do you have any other ideas?

That avoids having to pass a vtable dictionary to decode/encode, but in turn requires access to that same dictionary everywhere the size is computed. Is that so much better?

spec/lang/syntax.md
```diff
@@ -51,7 +47,17 @@ pub enum ValueExpr {
         #[specr::indirection]
         place: PlaceExpr,
     },
-
+    /// Lookup the function pointer for a trait object method.
+    /// Dynamic dispatch is represented as a `Call` to the result of this method,
+    /// with the `self` argument appropriately cast to a sized pointer type.
+    VTableLookup {
+        /// Must be a wide pointer to a trait object.
```
Indeximal, line 54: Note to self: Can be the metadata directly

spec/lang/syntax.md
```diff
@@ -477,4 +490,13 @@ pub struct Relocation {
     /// The offset within that allocation.
     pub offset: Offset,
 }
+
+/// A vtable for a trait-type pair.
+/// This is pointed to by the trait object metadata.
+pub struct VTable {
```
RalfJung, line 496: This needs to store some sort of identifier of the trait that this is a vtable for (for UB checking).

Indeximal:
I believe this is now ready for a review. It has gotten rather large with the churn of needing more information to compute sizes, adding the new type, the new metadata and the new operations, and all the miniutil changes that come with it.

spec/mem/pointer.md
```diff
@@ -25,6 +25,10 @@ We therefore use the term *thin pointer* for what has been described above, and
 /// of the address space.
 pub type Address = Int;
 
+/// A "trait name" is an identifier for the trait a vtable is for.
+/// This defines the size and function signatures of the methods.
```
RalfJung, line 29:
```suggestion
/// This defines the number of methods and their function signatures.
```
This isn't about the size of the type, right?

spec/lang/representation.md
```diff
@@ -155,17 +156,19 @@ impl PtrType {
     }
 }
 
-impl PointerMeta {
-    fn from_value<M: Memory>(value: Value<M>) -> Option<PointerMeta> {
+impl<Provenance> PointerMeta<Provenance> {
+    fn from_value<M: Memory<Provenance=Provenance>>(value: Value<M>) -> Option<PointerMeta<Provenance>> {
         match value {
             Value::Int(count) => Some(PointerMeta::ElementCount(count)),
+            Value::Ptr(ptr) => Some(PointerMeta::VTablePointer(ptr.thin_pointer)),
```
RalfJung, line 163: Can you ensure that `ptr` here does not have metadata? That should return `None`.

Or should it panic? Can this ever be called with non-metadata values?

spec/lang/representation.md
```diff
@@ -440,6 +452,48 @@ fn ensure_else_ub(b: bool, msg: &str) -> Result<()> {
...
 
 impl<M: Memory> Machine<M> {
+    /// Checks that a pointer is well-formed.
+    fn check_pointer(&self, ptr: Pointer<M::Provenance>, ptr_ty: PtrType) -> Result {
+        ensure_else_ub(ptr.thin_pointer.addr.in_bounds(Unsigned, M::T::PTR_SIZE), "Value::Ptr: pointer out-of-bounds")?;
+
+        // Safe pointer, i.e. references, boxes
+        if let Some(pointee) = ptr_ty.safe_pointee() {
+            let size = pointee.layout.compute_size(ptr.metadata, self.vtable_lookup());
+            let align = pointee.layout.compute_align(ptr.metadata, self.vtable_lookup());
```
RalfJung, line 462: Can we have `self.compute_size`/`self.compute_align` helper methods for this?

spec/lang/representation.md
```diff
@@ -440,6 +452,48 @@ fn ensure_else_ub(b: bool, msg: &str) -> Result<()> {
...
+            // The total size of slices must be at most `isize::MAX`.
```
RalfJung, line 463:
```suggestion
            // The total size must be at most `isize::MAX`.
```

spec/lang/representation.md
```diff
@@ -440,6 +452,48 @@ fn ensure_else_ub(b: bool, msg: &str) -> Result<()> {
...
+            ensure_else_ub(size.bytes().in_bounds(Signed, M::T::PTR_SIZE), "Value::Ptr: total size exeeds isize::MAX")?;
+
+            // Safe addresses need to be non-null, aligned, dereferenceable, and not point to an uninhabited type.
```
RalfJung, line 466:
```suggestion
            // Safe pointers need to be non-null, aligned, dereferenceable, and not point to an uninhabited type.
```

spec/lang/representation.md
```diff
@@ -440,6 +452,48 @@ fn ensure_else_ub(b: bool, msg: &str) -> Result<()> {
...
+            ensure_else_ub(pointee.inhabited, "Value::Ptr: safe pointer to uninhabited type")?;
+            ensure_else_ub(
+                self.mem.dereferenceable(ptr.thin_pointer, size).is_ok(),
+                "Value::Ptr: non-dereferenceable safe pointer"
+            )?;
+
+            // However, we do not care about the data stored wherever this pointer points to.
+        }
+
+        match (ptr.metadata, ptr_ty.meta_kind()) {
```
RalfJung, line 479: You should check this before calling `compute_size` since size computation may rely on this.

spec/lang/step/expressions.md
```diff
@@ -45,6 +45,15 @@ impl<M: Memory> Machine<M> {
                     provenance: None,
                 }.widen(None))
             },
+            Constant::VTablePointer(vtable_name) => {
+                let mut vtables = self.vtable_allocs.iter().filter(|(_, name)| *name == vtable_name);
+                // we could also pick a random one?
```
RalfJung, line 50:
```suggestion
                // FIXME: we should leave the choice of which vtable address to use to the frontend.
```

spec/lang/step/expressions.md
```diff
@@ -45,6 +45,15 @@ impl<M: Memory> Machine<M> {
                     provenance: None,
                 }.widen(None))
             },
+            Constant::VTablePointer(vtable_name) => {
+                let mut vtables = self.vtable_allocs.iter().filter(|(_, name)| *name == vtable_name);
```
RalfJung, line 49: `filter(...).next()` can be written as `find(...)`.

spec/lang/step/expressions.md
```diff
@@ -45,6 +45,15 @@ impl<M: Memory> Machine<M> {
...
+                // we could also pick a random one?
+                let Some((ptr, _)) = vtables.next() else {
+                    panic!("constant to unallocated vtable");
```
RalfJung, line 52: Does WF actually ensure that this panic cannot happen? I don't see it ensuring that.

spec/mem/pointer.md
```diff
@@ -124,6 +122,7 @@ pub enum PtrType {
         meta_kind: PointerMetaKind,
     },
     FnPtr,
+    VTablePtr,
```
RalfJung, line 125: Interesting, I had not thought of making this a type. But it makes sense. :)

spec/lang/step/operators.md
```diff
@@ -131,18 +134,46 @@ impl<M: Memory> Machine<M> {
...
+```rust
+impl<M: Memory> Machine<M> {
+    fn eval_un_op(&self, UnOp::VTableMethodLookup(method): UnOp, (operand, op_ty): (Value<M>, Type)) -> Result<(Value<M>, Type)> {
+        let (Value::Ptr(ptr), Type::Ptr(_ptr_ty)) = (operand, op_ty) else {
+            panic!("vtable lookup on non-pointer");
+        };
+        // It is checked in check_value that the vtable is always valid.
+        let vtable = self.vtable_lookup()(ptr.thin_pointer);
+        let Some(fn_name) = vtable.methods.get(method) else {
+            // This would be a type error, but since we don't store the trait, we do not type check this.
```
RalfJung, line 164:
```suggestion
            // This would be a type error, but since we don't store the trait in the VTabePtr type,
            // we do not type check this.
            // FIXME: Seems like it'd be nice to store the trait there?
```

spec/lang/syntax.md
```diff
@@ -419,10 +427,12 @@ pub enum IntrinsicOp {
 Finally, the general structure of programs and functions:
 
 ```rust
-/// Opaque types of names for functions and globals.
+/// Opaque types of names for functions, vtables, trait methods and globals.
```
RalfJung, line 430:
```suggestion
/// Opaque types of names for functions, vtables, trait methods, and globals.
```

spec/lang/well-formed.md
```diff
@@ -759,6 +776,16 @@ impl Program {
             }
         }
 
+        // Check vtables: All vtables for the same trait must have the same methods defined.
```
RalfJung, line 779:
```suggestion
        // Check vtables: All vtables for the same trait must have the same method names defined.
```

tooling/minimize/src/bb.rs
RalfJung: So what is the point of this comment? Should it be like this?
```suggestion
        // FIXME: turn `InstanceKind::Virtual` calls into trait method calls
```

tooling/miniutil/src/build/mod.rs
```diff
@@ -73,6 +84,32 @@ impl ProgramBuilder {
...
+            ty.layout::<DefaultTarget>().expect_size("only sized types can be trait objects"),
+            ty.layout::<DefaultTarget>().expect_align("only sized types can be trait objects"),
+        )
+    }
+
+    #[track_caller]
+    pub fn finish_vtable(&mut self, v: VTableBuilder) -> VTableName {
+        let name = v.name();
+        let vtable = v.finish_vtable();
+        // We could store and check the number of functions matches what was defined here.
```
RalfJung, line 103: Is this a TODO or FIXME, or what is the point of this comment?

tooling/miniutil/src/build/ty.rs
```diff
@@ -12,10 +12,28 @@ pub fn ref_ty(pointee: PointeeInfo) -> Type {
...
+    })
+}
+
 pub fn ref_mut_ty(pointee: PointeeInfo) -> Type {
     Type::Ptr(PtrType::Ref { mutbl: Mutability::Mutable, pointee })
 }
 
+pub fn ref_mut_ty_for(ty: Type) -> Type {
+    ref_mut_ty(PointeeInfo {
+        layout: ty.layout::<x86_64>(),
```
RalfJung, line 30:
```suggestion
        layout: ty.layout::<DefaultTarget>(),
```
Applies everywhere else as well.

tooling/miniutil/src/build/ty.rs
```diff
@@ -12,10 +12,28 @@ pub fn ref_ty(pointee: PointeeInfo) -> Type {
...
+        inhabited: true,
+        freeze: true,
+        unpin: true,
```
RalfJung, line 33: The fact that you hard-code these defaults makes this function quite confusing. It's not actually correct for all types.

tooling/miniutil/src/fmt/vtables.rs
```diff
@@ -0,0 +1,32 @@
...
+    vtables.sort_by_key(|(VTableName(name), _)| *name);
+
+    for (vname, vtable) in vtables {
+        out += &fmt_vtable(vname, vtable);
+    }
+    out
+}
+
+fn fmt_vtable(vname: VTableName, vtable: VTable) -> String {
+    let mut out = format!("vt{id} {{\n", id = vname.0.get_internal());
```
RalfJung, line 18:
```suggestion
    let mut out = format!("vtable{id} {{\n", id = vname.0.get_internal());
```
No need to use unreadable mnemonics. ;)

tooling/minitest/src/tests/trait_object.rs
```diff
@@ -0,0 +1,78 @@
...
+/// let x: usize = 42;
+/// let y: &dyn A = &x;
+/// assert_eq!(x == y.foo());
+/// ```
+#[test]
+fn dynamic_dispatch() {
+    let mut p = ProgramBuilder::new();
+
+    let mut trait_a = p.declare_trait();
+    let method_a_foo = trait_a.fresh_method_name();
```
RalfJung, line 24: Wouldn't `declare_method` be more consistent?

tooling/miniutil/src/build/mod.rs
```diff
@@ -198,6 +235,64 @@ impl FunctionBuilder {
...
+
+    pub fn fresh_method_name(&mut self) -> TraitMethodName {
+        let idx = self.next_method;
+        self.next_method += 1;
+        TraitMethodName(Name::from_internal(idx))
+    }
+
+    #[track_caller]
+    pub fn finish_trait(self) -> TraitName {
+        self.name
```
RalfJung, line 292:
```suggestion
        // We don't actually have to store the registered trait methods anywhere,
        // so all we do here is return the name.
        self.name
```

spec/lang/well-formed.md
```diff
@@ -759,6 +776,16 @@ impl Program {
...
+        let mut traits: Map<TraitName, Set<TraitMethodName>> = Map::new();
+        for (_name, vtable) in self.vtables {
+            let methods = vtable.methods.keys().collect();
+            match traits.insert(vtable.trait_name, methods) {
+                Some(prev_methods) => ensure_wf(methods == prev_methods, "Program: vtables with same trait but different methods")?,
```
RalfJung, line 784: Hm, should the program contain a `traits: Map<TraitName, Set<TraitMethodName>>` field so that this becomes a bit less ad-hoc here?

spec/lang/step/expressions.md
Indeximal: Well, it ensures `prog.vtables.contains_key(vtable_name)`, and here I assume `Machine` actually allocates every vtable in the program.

spec/lang/step/operators.md
Indeximal: This is a disadvantage of having `VTableMethodLookup` operate on the metadata (i.e. the vtable pointer) directly. If we had it on the trait object pointer, the type would include the vtable name for the metadata.
So we could have `VTableMethodLookup` take the trait in addition to the method (which was once effectively the case with globally unique MethodNames).

spec/lang/syntax.md
Indeximal: So, you choose to go for the path of Oxford Commas 😂

spec/mem/pointer.md
Indeximal: Yes, size was ambiguous.

spec/lang/representation.md
RalfJung: If callers always have the `PointerMetaKind` available, then that seems cleaner, yes. Can you adjust this?

spec/lang/representation.md
```diff
@@ -440,6 +453,48 @@ fn ensure_else_ub(b: bool, msg: &str) -> Result<()> {
 }
 
 impl<M: Memory> Machine<M> {
+    /// Checks that a pointer is well-formed.
+    fn check_pointer(&self, ptr: Pointer<M::Provenance>, ptr_ty: PtrType) -> Result {
+        ensure_else_ub(ptr.thin_pointer.addr.in_bounds(Unsigned, M::T::PTR_SIZE), "Value::Ptr: pointer out-of-bounds")?;
+
+        match (ptr.metadata, ptr_ty.meta_kind()) {
```
RalfJung, line 460:
```suggestion
        // Ensure the metadata makes sense.
        match (ptr.metadata, ptr_ty.meta_kind()) {
```

spec/lang/step/operators.md
RalfJung: I think the fix is to make `PtrType::VTabePtr` carry the trait name. But that can be a future PR.

tooling/miniutil/src/build/mod.rs
RalfJung: In that case I'd phrase it more like
```suggestion
        // Currently, we do not check that this vtable has all the methods required by the trait.
```

tooling/miniutil/src/build/ty.rs
RalfJung: These flags say, e.g., whether the type is `Freeze` which indicates that there is no `UnsafeCell` in this type. Minimize computes this from Rust types, but for minitest we need to set this ourselves.

I think this is fine for test use but the name should indicate that. Maybe `ref_mut_ty_default_markers_for`? And then the doc comment should explain the rest. (`Freeze` and `Unpin` are marker traits, hence "default markers". I am open to other names.)

spec/lang/well-formed.md
RalfJung: I think a comment can easily explain that confusion.

Having some central place where all (dynamically invokeable) traits are "declared" makes a lot of sense IMO.

spec/lang/representation.md
Indeximal: Ok, I understood correctly. Currently `self.compute_size == self.size_computer()` (using partial application). Thats what I meant.

tooling/miniutil/src/build/ty.rs
Indeximal: Okay. Are these flags given for a trait object?

spec/lang/well-formed.md
```diff
@@ -759,6 +776,15 @@ impl Program {
             }
         }
 
+        // Check vtables: All vtables for the same trait must have the same method names defined.
+        for (_name, vtable) in self.vtables {
+            let Some(trait_methods) = self.traits.get(vtable.trait_name) else {
+                throw_ill_formed!("Program: vtable for unknown trait");
+            };
+            let methods = vtable.methods.keys().collect::<Set<_>>();
+            ensure_wf(methods == trait_methods, "Program: vtables which defines invalid methods")?;
```
RalfJung, line 785:
```suggestion
            ensure_wf(methods == trait_methods, "Program: vtable has not the right set of methods")?;
```
Also, is there a test hitting this case?

tooling/miniutil/src/build/mod.rs
```diff
@@ -73,6 +87,52 @@ impl ProgramBuilder {
...
+    }
+
+    #[track_caller]
+    pub fn finish_vtable(&mut self, v: VTableBuilder) -> VTableName {
+        let name = v.name();
+        let vtable = v.finish_vtable();
+        // Check that this vtable has all the methods required by the trait.
+        assert_eq!(
+            self.traits.get(vtable.trait_name).expect("TraitName must have been declared"),
+            vtable.methods.keys().collect()
```
RalfJung, line 117:
```suggestion
            vtable.methods.keys().collect(),
            "vtable does not declare the right set of methods",
```

spec/lang/representation.md
```diff
@@ -440,6 +461,55 @@ fn ensure_else_ub(b: bool, msg: &str) -> Result<()> {
...
+        };
+
+        Ok(())
+    }
+
+    /// Checks that a pointer is well-formed.
+    fn check_pointer(&self, ptr: Pointer<M::Provenance>, ptr_ty: PtrType) -> Result {
+        ensure_else_ub(ptr.thin_pointer.addr.in_bounds(Unsigned, M::T::PTR_SIZE), "Value::Ptr: pointer out-of-bounds")?;
+
+        self.check_pointer_metadata(ptr.metadata, ptr_ty.meta_kind())?;
```
RalfJung, line 488:
```suggestion
        // This has to be checked first, to ensure we can e.g. compute size/align below.
        self.check_pointer_metadata(ptr.metadata, ptr_ty.meta_kind())?;
```

spec/lang/well-formed.md
Indeximal: No, this PR contains neither UB nor Ill-formedness tests. I should probably add this though.

RalfJung: Yes please do :)

tooling/minitest/src/tests/trait_object.rs
```diff
@@ -17,7 +17,7 @@ use crate::*;
 /// assert_eq!(x == y.foo());
 /// ```
 #[test]
-fn dynamic_dispatch() {
+fn example() {
```
RalfJung, line 20: The old name seemed better?

tooling/minitest/src/tests/trait_object.rs
```diff
@@ -76,3 +73,375 @@ fn dynamic_dispatch() {
     dump_program(p);
     assert_stop::<BasicMem>(p);
 }
+
+/// Tests that assigning a vtable defined for a different type, but same trait is fine.
+///
+/// This is not good code, but not found as UB, not even in MIRI, because it gets near impossible to track.
```
RalfJung, line 79:
```suggestion
/// This is not good code, but not defined as UB, not even in Miri.
```

tooling/minitest/src/tests/trait_object.rs
```diff
@@ -76,3 +73,375 @@ fn dynamic_dispatch() {
...
+
+        f.exit();
+        p.finish_function(f)
+    };
+
+    let p = p.finish_program(main);
+    assert_ub::<BasicMem>(p, "Value::Ptr: invalid vtable in metadata");
+}
+
+// Ill-formed tests
```
RalfJung, line 320: These tests could use comments as well.

RalfJung:
Looks great! Please squash this a little (doesn't have to be a single commit, but fewer of these "fix typo" commits ;).  Please use the `--keep-base` flag when squashing so that the force-push diff is easier to review.

Indeximal:
Okay, I have squashed into some more explicit commits, and had to do another force-push to fix nightly-bump-induced formatting changes.

RalfJung:
Awesome, let's land this. :)

Issue #191, opened 2024-06-14
Add support for 'char'


RalfJung:
This is basically a 4-byte integer type, but with a particular validity invariant: in must be in the range `0..=0x10FFFF`, and *not* in the range `0xD800..=0xDFFF`.

Issue #175, opened 2024-04-19
Build our own standard library


RalfJung:
Currently we use the standard library that is shipped with rustc. This means if a function is already codegen'd, we fail as we cannot access its MIR.

Instead we need to work more like Miri, and build our own standard library.

RalfJung:
> Instead we need to work more like Miri, and build our own standard library.

Or maybe we should use `-Zbuild-std`. That could be a lot less work...

PR #240, opened 2024-09-20
Better slice support

Indeximal:
This PR contains multiple changes to make various minimized code using slices compile properly.

This includes:
- Adding various bounds check failure functions to the `is_panic_fn` hack. This is was sufficient to allow safe subslicing.
- Add the `IntUnOp::CountOnes` operation to support the `ctpop` intrinsic. It was used by `slice::from_raw_parts`.
- Convert the `str` type to a `[u8]`. It was used by `panic_fmt` in `slice::from_raw_parts`. Now this can be called aswell.
- Allow minimize to handle unsized const pointers. This allows for literal strings.
- Some tests that the same things also work for strings.

There are still several things outstanding that hinder widespread use of slices and strings:
- #243
- #234 
- #191 
- #175 
- the `compare_bytes` intrinsic is not implemented, which is used for `[u8]` and `str`.
- closures aren't implemented.

### DIFF
--- a/spec/lang/step/operators.md
+++ b/spec/lang/step/operators.md
@@ -15,22 +15,57 @@ impl<M: Memory> Machine<M> {
 
 ```rust
 impl<M: Memory> Machine<M> {
-    fn eval_int_un_op(op: IntUnOp, operand: Int) -> Result<Int> {
+    /// Perform the operation on the mathematical integer `operand`,
+    /// but correcting for non-pure effects dependent of the `operand_ty`.
+    fn eval_int_un_op(op: IntUnOp, operand: Int, operand_ty: IntType) -> Result<Int> {
         use IntUnOp::*;
         ret(match op {
-            Neg => -operand,
-            BitNot => !operand,
+            // Put the result into the right range (in case of overflow).
+            Neg => operand_ty.bring_in_bounds(-operand),
+            // Put the result into the right range (in case of a unsigned numbers, which `!`
+            // makes negative by inverting all the leading zeros).
+            BitNot => operand_ty.bring_in_bounds(!operand),
+            // This can never overflow, as the total number of bits is below `u32::MAX`.
+            CountOnes => Self::eval_count_ones(operand, operand_ty),
         })
     }
     fn eval_un_op(&self, UnOp::Int(op): UnOp, (operand, op_ty): (Value<M>, Type)) -> Result<(Value<M>, Type)> {
         let Type::Int(int_ty) = op_ty else { panic!("non-integer input to integer operation") };
         let Value::Int(operand) = operand else { panic!("non-integer input to integer operation") };
 
-        // Perform the operation.
-        let result = Self::eval_int_un_op(op, operand)?;
-        // Put the result into the right range (in case of overflow).
-        let result = int_ty.bring_in_bounds(result);
-        ret((Value::Int(result), Type::Int(int_ty)))
+        let ret_ty = match op {
+            IntUnOp::CountOnes => IntType { signed: Unsigned, size: Size::from_bytes(4).unwrap() },
+            _ => int_ty,
+        };
+
+        let result = Value::Int(Self::eval_int_un_op(op, operand, int_ty)?);
+
+        // Sanity-check that the result of `eval_int_un_op` is in-bounds.
+        result.check_wf(Type::Int(ret_ty))
+            .expect("sanity check: result of UnOp::Int does not fit in the return type");
+        ret((result, Type::Int(ret_ty)))
+    }
+}
+```
+
+`CountOnes` aka `ctpop` is a rust intrinsic on integer types that always returns a `u32`.
+This is not a pure function on mathematical integers,
+since the bit representation for non-negative values has infinite zeros,
+and infinite ones for negatives values.
+Therefore, the bit width is important, and iterating until `remaining_bits = 0` is not correct.
+
+```rust
+impl<M: Memory> Machine<M> {
+    fn eval_count_ones(operand: Int, int_ty: IntType) -> Int {
+        let mut ones = Int::ZERO;
+        let mut remaining_bits = operand;
+        // Iterate once per bit in the bit width, afterwards the remaining bits are leading bits only.
+        for _ in Int::ZERO..int_ty.size.bits() {
+            // Extract the least significant bit and update `remaining_bits` to remove that bit.
+            ones += remaining_bits & Int::ONE;
+            remaining_bits >>= 1;
+        }
+        ones
     }
 }
 ```
--- a/spec/lang/step/operators.md
+++ b/spec/lang/step/operators.md
@@ -48,7 +48,7 @@ impl<M: Memory> Machine<M> {
 }
 ```
 
-`CountOnes` aka `ctpop` is a rust intrinsic on integer types that always returns a `u32`.
+`CountOnes` aka `ctpop` is a Rust intrinsic on integer types that always returns a `u32`.
 This is not a pure function on mathematical integers,
 since the bit representation for non-negative values has infinite zeros,
 and infinite ones for negatives values.
--- a/tooling/minimize/src/bb.rs
+++ b/tooling/minimize/src/bb.rs
@@ -330,6 +330,16 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
                 let terminator = Terminator::Goto(self.bb_name_map[&target.unwrap()]);
                 return TerminatorResult { stmts: list!(stmt), terminator };
             }
+            rs::sym::ctpop => {
+                let v = self.translate_operand(&args[0].node, span);
+                let destination = self.translate_place(&destination, span);
+
+                let val = build::count_ones(v);
+                let stmt = Statement::Assign { destination, source: val };
+
+                let terminator = Terminator::Goto(self.bb_name_map[&target.unwrap()]);
+                return TerminatorResult { stmts: list!(stmt), terminator };
+            }
             rs::sym::exact_div => {
                 let l = self.translate_operand(&args[0].node, span);
                 let r = self.translate_operand(&args[1].node, span);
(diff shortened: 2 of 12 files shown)


spec/lang/step/operators.md
```diff
@@ -15,22 +15,36 @@ impl<M: Memory> Machine<M> {
...
+            _ => int_ty,
+        };
+
         // Perform the operation.
-        let result = Self::eval_int_un_op(op, operand)?;
+        let result = Self::eval_int_un_op(op, operand, int_ty)?;
         // Put the result into the right range (in case of overflow).
-        let result = int_ty.bring_in_bounds(result);
-        ret((Value::Int(result), Type::Int(int_ty)))
+        let result = ret_ty.bring_in_bounds(result);
```
RalfJung, line 46: I don't think we want to call `bring_in_bounds` on the result of `CountOnes`. That should be a bug if we ever lose precision there.

spec/lang/step/operators.md
```diff
@@ -15,22 +15,36 @@ impl<M: Memory> Machine<M> {
...
-    fn eval_int_un_op(op: IntUnOp, operand: Int) -> Result<Int> {
+    fn eval_int_un_op(op: IntUnOp, operand: Int, int_ty: IntType) -> Result<Int> {
         use IntUnOp::*;
         ret(match op {
             Neg => -operand,
             BitNot => !operand,
+            CountOnes => {
+                let mut ones = Int::ZERO;
+                let mut remaining_bits = operand;
+                for _ in Int::ZERO..int_ty.size.bits() {
```
RalfJung, line 26: Some comments here would be good.

Please move the actual counting of ones into a helper function.

Would `while remaining_bits != Int::ZERO` also work?

Will this work correctly for negative numbers?

tooling/minimize/src/bb.rs
```diff
@@ -443,5 +453,13 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
 
 // HACK to skip translating some functions we can't handle yet.
```
RalfJung, line 454:
```suggestion
// HACK to skip translating some functions we can't handle yet.
// These always panic so we just turn them into the panic intrinsic.
```

tooling/minimize/src/ty.rs
```diff
@@ -106,6 +111,13 @@ impl<'tcx> Ctxt<'tcx> {
                 let elem = GcCow::new(self.translate_ty(*ty, span));
                 Type::Slice { elem }
             }
+            rs::TyKind::Str => {
```
RalfJung, line 114: Can you add some tests for `str`? Currently it seems only indirectly exercised.

tooling/minimize/src/ty.rs
```diff
@@ -106,6 +111,13 @@ impl<'tcx> Ctxt<'tcx> {
...
+                let elem = GcCow::new(Type::Int(IntType {
```
RalfJung, line 115:
```suggestion
                // Treat `str` like `[u8]`.
                let elem = GcCow::new(Type::Int(IntType {
```

tooling/minimize/src/ty.rs
```diff
@@ -23,6 +23,11 @@ impl<'tcx> Ctxt<'tcx> {
 
                 PointeeInfo { size, align, inhabited, freeze, unpin }
             }
+            &rs::TyKind::Str => {
+                let size = SizeStrategy::Slice(Size::from_bytes_const(1));
```
RalfJung, line 27:
```suggestion
                // Treat `str` like `[u8]`.
                let size = SizeStrategy::Slice(Size::from_bytes_const(1));
```

tooling/minimize/src/ty.rs
Indeximal: Can't, they get const promoted and unsized const pointers don't work currently.

RalfJung: What is `str` even needed for then? panic_fmt is marked as a panic_fn so it should not need `str`.

Indeximal: Idk maybe it isn't used by panic_fmt, but something else, but it doesn't build with `panic_fmt` marked, but no str support

Indeximal:
I fixed some things, and added some str tests, but they are rather limited due to various todos, including #234, #175 and #191.

tooling/minimize/src/ty.rs
RalfJung: Then it should be possible to write a test.

RalfJung: Ah, you do have tests now, never mind. :)

tooling/minimize/src/rvalue.rs
```diff
@@ -413,14 +413,19 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
                         let root = GcCow::new(expr);
                         PlaceExpr::Index { root, index: i }
                     }
+                    stable_mir::mir::ProjectionElem::ConstantIndex {
+                        offset,
+                        from_end: false,
+                        ..
```
RalfJung, line 419: Please don't ignore `min_length`, that seems dangerous.

We should raise UB if the length is smaller than `min_length`.

RalfJung (commented):
First round of comments, didn't get to the tests yet.

tooling/minimize/tests/pass/slice.rs
```diff
@@ -22,21 +25,45 @@ fn main() {
...
+    assert!(sub_slice[0] == -40);
+
+    let sub_slice = &slice[..4];
+    assert!(sub_slice.len() == 4);
+    assert!(sub_slice[0] == 50);
+
+    // Check equality
+    assert!(&slice[1..4] == &[-40, 30, -20]);
+    assert!(slice[1..4] == [-40, 30, -20]);
+    // This would fails, since it specializes to `compare_bytes`
```
RalfJung, line 66:
```suggestion
    // This fails, since it uses the `compare_bytes` intrinsic.
```

tooling/minimize/src/rvalue.rs
Indeximal: Makes sense, but this is actually rather a convoluted change, so I'd first like to know the purpose & be sure it is dangerous. Because we do generate UB on invalid indices anyway, so this just seems like additional UB? Or is there a case where we want UB, but it is only encoded by the `min_length`, but the index is valid? For arrays it is always exactly the size.
And the implementation is convoluted, since this now means, the folding cannot just output a `PlaceExpr`, there needs to be code generated to raise UB, which must include code to get the length of the indexable, which then depends on whether it is an array or slice.

spec/lang/step/operators.md
```diff
@@ -15,22 +15,53 @@ impl<M: Memory> Machine<M> {
 
 ```rust
 impl<M: Memory> Machine<M> {
-    fn eval_int_un_op(op: IntUnOp, operand: Int) -> Result<Int> {
+    fn eval_int_un_op(op: IntUnOp, operand: Int, int_ty: IntType) -> Result<Int> {
         use IntUnOp::*;
         ret(match op {
-            Neg => -operand,
-            BitNot => !operand,
+            // Put the result into the right range (in case of overflow).
```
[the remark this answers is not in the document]
Indeximal: No, actually BitNot also needs this (as some tests have caught). The reason for this is that the mathematical BitNot function will flip all infinite bits, causing the a positive number to become negative. If this number was meant to be unsigned, then this is not the current semantics, instead, the bits above the bit width stay zero and the number should stay positive.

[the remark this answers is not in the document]
RalfJung: Ah, good catch!

spec/lang/step/operators.md
```diff
@@ -15,22 +15,56 @@ impl<M: Memory> Machine<M> {
...
+    /// Perform the operation on the mathematical integer `operand`,
+    /// but correcting for non-pure effects dependent of the `operand_ty`.
+    fn eval_int_un_op(op: IntUnOp, operand: Int, operand_ty: IntType) -> Result<Int> {
         use IntUnOp::*;
         ret(match op {
-            Neg => -operand,
-            BitNot => !operand,
+            // Put the result into the right range (in case of overflow).
+            Neg => operand_ty.bring_in_bounds(-operand),
+            // Put the result into the right range (in case of a unsigned numbers).
```
RalfJung, line 25:
```suggestion
            // Put the result into the right range (in case of a unsigned numbers, which `!`
            // makes negative by inverting all the leading zeros).
```

tooling/minimize/src/rvalue.rs
RalfJung: I am honestly not entirely sure what the point of the `min_length` field is, but Miri uses this for a dedicated UB check, and that seems better than just ignoring it. I wrote https://github.com/rust-lang/rust/pull/131285 to clarify this in the rustc docs.

> which must include code to get the length of the indexable, which then depends on whether it is an array or slice.

You can share that code with `Rvalue::Len`.

Indeximal:
I have instead reverted the partial implementation of ConstantIndex, as it didn't seem to break current tests. Still, I have opened an issue and also collected other TODOs in this PRs comment.
With that, this PR is ready imo.

RalfJung:
The commit message still talks about "const slices"?

spec/lang/step/operators.md
```diff
@@ -15,22 +15,57 @@ impl<M: Memory> Machine<M> {
...
+
+        // Sanity-check that the result of `eval_int_un_op` is in-bounds.
+        result.check_wf(Type::Int(ret_ty))
+            .expect("sanity check: result of UnOp::Int does not fit in the return type");
+        ret((result, Type::Int(ret_ty)))
+    }
+}
+```
+
+`CountOnes` aka `ctpop` is a rust intrinsic on integer types that always returns a `u32`.
```
RalfJung, line 51:
```suggestion
`CountOnes` aka `ctpop` is a Rust intrinsic on integer types that always returns a `u32`.
```

Indeximal:
> The commit message still talks about "const slices"?

This is in reference to the changes in `tooling/minimize/src/constant.rs`, allowing wide pointers in constants.

RalfJung:
Ah, makes sense, thanks. :)

PR #241, opened 2024-10-04
Refactor size and alignment into layoutstrategy

Indeximal:
Since for structs with a trait object tail, the size cannot be known without knowing the align, having  separate Align and Size strategies does not work.
Therefore this PR prepares for trait objects by allowing the alignment to be dynamic and does this while merging them into a new struct `LayoutStrategy`.

### DIFF
--- a/spec/lang/types.md
+++ b/spec/lang/types.md
@@ -129,7 +129,7 @@ They *do* have a mutability since that is (or will be) relevant for the memory m
 
 ## Layout of a type
 
-Here we define how to compute the size and other layout properties of a type.
+Here we define the size and other layout properties of a type.
 
 ```rust
 impl IntType {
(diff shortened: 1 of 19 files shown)


spec/mem/pointer.md
```diff
@@ -172,18 +189,19 @@ impl PtrType {
...
+        if let Some(pointee) = self.safe_pointee() {
             // Safe addresses need to be non-null, aligned, and not point to an uninhabited type.
             // (Think: uninhabited types have impossible alignment.)
-            addr != 0 && layout.align.is_aligned(addr) && layout.inhabited
+            // FIXME(UnsizedTypes): Compute the alignment, don't assume it is sized.
+            addr != 0 && pointee.layout.expect_align("FIXME").is_aligned(addr) && pointee.inhabited
         } else {
             true
         }
     }
```
Indeximal, line 200: Do you agree that this function should check the dynamic alignment and therefore needs the full pointer?

spec/mem/pointer.md
```diff
@@ -129,35 +129,52 @@ pub enum PtrType {
...
 
     /// Computes the dynamic size, but the caller must provide compatible metadata.
-    pub fn compute(self, meta: Option<PointerMeta>) -> Size {
+    pub fn compute_size(self, meta: Option<PointerMeta>) -> Size {
         match (self, meta) {
-            (SizeStrategy::Sized(size), None) => size,
+            (LayoutStrategy::Sized(size, _), None) => size,
             // FIXME(UnsizedTypes): We need to assert that the resulting size isn't too big.
```
RalfJung, line 157: Not "assert", we need to raise UB if it is too big. This will require access to a target machine instance.

spec/mem/pointer.md
RalfJung: Ah, good question... yes we probably want that. But I am not sure if it is better to pass in the metadata as an extra argument, or pass in the full pointer. You can check out how this function is called, that should help with this.

RalfJung:
Looks reasonable!

Indeximal:
The problem of computing the align in dereferenceablity checks has been resolved by lifting them to the machine in #244.

So this is ready, but weirdly stacked on #244. The merge [a271499](https://github.com/minirust/minirust/pull/241/commits/a2714996bca926db75816e6b8c25949e74ef5b72) (commits f19f5c2..97db5fb) as well as the following commit [79fc14c](https://github.com/minirust/minirust/pull/241/commits/79fc14cb9910371a506090609949b566eb3e467b) are from this PR.

spec/lang/types.md
```diff
@@ -143,38 +143,83 @@ impl IntType {
...
-            Slice { elem } => SizeStrategy::Slice(elem.size::<T>().expect_sized("WF ensures slice element is sized")),
         }
     }
 
-    pub fn align<T: Target>(self) -> Align {
-        use Type::*;
+    /// Returns the metadata kind when this type is used as a pointee.
+    pub fn meta_kind(self) -> PointerMetaKind {
+        // The metadata kind does not rely on the target, therefore we can give any target.
+        self.layout::<x86_64>().meta_kind()
```
RalfJung, line 170: We shouldn't hard code a target here...

Let's keep the old direct impl, and assert in WF that it matches `layout().meta_kind()`.

spec/lang/types.md
```diff
@@ -143,38 +143,83 @@ impl IntType {
...
+    }
+}
+```
+
+And we also define how to compute the actual size and alignment.
+
+```rust
+impl LayoutStrategy {
+    pub fn is_sized(self) -> bool {
+        matches!(self, LayoutStrategy::Sized(_, _))
```
RalfJung, line 180:
```suggestion
        matches!(self, LayoutStrategy::Sized(..))
```

tooling/miniutil/src/build/ty_conv.rs
```diff
@@ -8,11 +8,8 @@ pub trait TypeConv {
     fn get_type() -> Type;
 
     // Convenience methods, these should not be overridden.
-    fn get_size() -> SizeStrategy {
-        Self::get_type().size::<DefaultTarget>()
-    }
-    fn get_align() -> Align {
-        Self::get_type().align::<DefaultTarget>()
+    fn get_layout() -> LayoutStrategy {
+        Self::get_type().layout::<DefaultTarget>()
```
RalfJung, line 12: Might be worth keeping the `get_size/align` methods (with a `Self: Sized` bound)

PR #244, opened 2024-10-24
Refactor value validity as an addtional decode step

Indeximal:
This PR contains the necessary changes for lifting typed copies to `Machine` and introducing an explicit `check_value` function to check the validity invariant.
I have also rewritten the affected literary sections and reordered some sections, to feel more natural to me. This is of course a proposal.


Indeximal:
The changes in the PR have been completed, and I look forward to your review.

spec/lang/representation.md
```diff
@@ -405,22 +395,174 @@ Unsized types do not have values and thus there is no representation relation.
...
+                    // (Think: uninhabited types have impossible alignment.)
+                    ensure_ub(ptr.thin_pointer.addr != 0, "Value::Ptr: null safe pointer")?;
+                    ensure_ub(layout.align.is_aligned(ptr.thin_pointer.addr), "Value::Ptr: unaligned safe pointer")?;
+                    ensure_ub(layout.inhabited, "Value::Ptr: safe pointer to uninhabited type")?;
+                    ensure_ub(
+                        self.mem.dereferenceable(ptr.thin_pointer, size).is_ok(),
+                        "Value::Ptr: non-dereferenceable safe pointer"
+                    )?;
+
+                    // However, is it not UB, if the validity invariant of the pointee is broken.
```
RalfJung, line 461:
```suggestion
                    // However, we do not care about the data stored wherever this pointer points to.
```

spec/lang/representation.md
```diff
@@ -405,22 +395,174 @@ Unsized types do not have values and thus there is no representation relation.
...
+            }
+            (Value::Union(chunk_data), Type::Union { chunks, .. }) => {
+                ensure_ub(chunk_data.len() == chunks.len(), "Value::Union: invalid chunk size")?;
+                for (data, (_, size)) in chunk_data.zip(chunks) {
+                    ensure_ub(data.len() == size.bytes(), "Value::Union: invalid chunk data")?;
+                }
+            }
+            (Value::Variant { discriminant, data }, Type::Enum { variants, .. }) => {
+                let Some(variant) = variants.get(discriminant) else {
+                    throw_ub!("Value::Variant: invalid discrimant type");
```
RalfJung, line 490:
```suggestion
                    throw_ub!("Value::Variant: invalid discrimant");
```

spec/lang/representation.md
```diff
@@ -405,22 +395,174 @@ Unsized types do not have values and thus there is no representation relation.
...
+                val
+            }
+            None => throw_ub!("load at type {ty:?} but the data in memory violates the validity invariant"), // FIXME use Display instead of Debug for `ty`
+        })
+    }
+}
+
+impl<M: Memory> ConcurrentMemory<M> {
+    /// Find all pointers in this value, ensure they are valid, and retag them.
+    fn retag_val(&mut self, frame_extra: &mut M::FrameExtra, val: Value<M>, ty: Type, fn_entry: bool) -> Result<Value<M>> {
```
RalfJung, line 535: This should probably be moved to the `Validating a value` section in `statements.md`.

spec/lang/representation.md
```diff
@@ -3,11 +3,17 @@
...
+
+`encode` and `decode` must satisfy some [properties](#generic-properties), which the specification must ensure, i.e. violating these properties is a spec bug.
+However, they can also make some assumptions:
+The types in the representation relation are all [well-formed](well-formed.md#well-formed-layouts-and-types) and sized, as unsized types cannot be represented as values.
+`encode` can also assume the value is [valid][valid-value] for the type,
+and `decode` can assume the size of the byte list matches the size of the type.
 
 [representation relation]: https://github.com/rust-lang/unsafe-code-guidelines/blob/master/reference/src/glossary.md#representation-relation
-[well-formed-value]: well-formed.md#well-formed-values
+[valid-value]: #Value-validity
```
RalfJung, line 16: I am not sure we want to rename "well-formed" to "valid"... I am trying to move away from the term "validity invariant" (see https://github.com/rust-lang/unsafe-code-guidelines/issues/539), so it seems a step backwards for MiniRust to start using this term now.

RalfJung (commented):
Thanks, overall this looks great!

spec/lang/representation.md
```diff
@@ -405,22 +395,174 @@ Unsized types do not have values and thus there is no representation relation.
...
+        panic!("decode of Type::Slice")
     }
     fn encode<M: Memory>(Type::Slice { .. }: Self, val: Value<M>) -> List<AbstractByte<M::Provenance>> {
-        panic!("tried to endoce a slice")
+        panic!("encode of Type::Slice")
+    }
+}
+```
+
+## Value validity
```
RalfJung, line 406:
```suggestion
## Language invariant for valid representations

The language specification enforces (by penalty of Undefined Behavior) that when a range of bytes is "used at a given type", it satisfies an invariant -- the *language invariant* (or "validity invariant"). [...]
```

spec/lang/representation.md
```diff
@@ -3,11 +3,17 @@
 The main purpose of [types](types.md) is to define how [values](values.md) are (de)serialized into/from memory.
 This is the *[representation relation]*, which is defined in the following.
 `decode` converts a list of bytes into a value; this operation can fail if the byte list is not a valid encoding for the given type.
-`encode` inverts `decode`; it will always work when the value is [well-formed][well-formed-value] for the given type (which the specification must ensure, i.e. violating this property is a spec bug).
-Unsized types cannot be represented as values and thus a call to `encode` or `decode` for such types can never occur in a well-formed program.
+Some types also have additional runtime constraints to be [valid][valid-value], which is always checked after decoding.
```
RalfJung, line 6:
```suggestion
Some types also have additional runtime constraints to be [well-formed][well-formed-value], which is always checked after decoding.
```

spec/lang/step/expressions.md
```diff
@@ -174,6 +167,9 @@ impl<M: Memory> Machine<M> {
             panic!("AddrOf generated an incompatible pointer type for the metadata");
```
RalfJung, line 167: Shouldn't this check be in `check_value`?

Indeximal:
I've given it another go, restructuring a bit more than before.

spec/lang/step/statements.md
```diff
@@ -89,24 +87,49 @@ impl<M: Memory> Machine<M> {
...
+To do this, we lift retagging from pointers to compound values.
 
 ```rust
 impl<M: Memory> Machine<M> {
     fn eval_statement(&mut self, Statement::Validate { place, fn_entry }: Statement) -> NdResult {
         let (place, ty) = self.eval_place(place)?;
 
-        // WF ensures all validate expressions are sized.
-        let val = self.mem.place_load(place, ty)?;
+        // WF ensures all validate expressions are sized, so we can invoke the load.
```
RalfJung, line 99:
```suggestion
        // WF ensures all valid expressions are sized, so we can invoke the load.
```

spec/lang/step/statements.md
```diff
@@ -106,6 +107,29 @@ impl<M: Memory> Machine<M> {
         ret(())
     }
 }
+
+impl<M: Memory> ConcurrentMemory<M> {
+    /// Find all pointers in this value, ensure they are valid, and retag them.
+    fn retag_val(&mut self, frame_extra: &mut M::FrameExtra, val: Value<M>, ty: Type, fn_entry: bool) -> Result<Value<M>> {
```
RalfJung, line 113: Please move this above the place where it is used.

spec/lang/representation.md
```diff
@@ -1,52 +1,57 @@
 # MiniRust representation relation
 
-The main purpose of [types](types.md) is to define how [values](values.md) are (de)serialized into/from memory.
-This is the *[representation relation]*, which is defined in the following.
-`decode` converts a list of bytes into a value; this operation can fail if the byte list is not a valid encoding for the given type.
-`encode` inverts `decode`; it will always work when the value is [well-formed][well-formed-value] for the given type (which the specification must ensure, i.e. violating this property is a spec bug).
-Unsized types cannot be represented as values and thus a call to `encode` or `decode` for such types can never occur in a well-formed program.
+The language specification enforces (by penalty of Undefined Behavior) that when a range of bytes is "used at a given type",
+it satisfies an invariant -- the *language invariant* (or "validity invariant").
+This is the main purpose of [types](types.md): to define how [values](values.md) are (de)serialized into/from memory.
```
RalfJung, line 5:
```suggestion
The main purpose of [types](types.md) is to define how [values](values.md) are (de)serialized into/from memory.
This is the *[representation relation]*, which is defined in the following.
We also use this file to define the *language invariant* (or "validity invariant"): a predicate on byte ranges that must always hold when a range of bytes is "used at a given type".
For most types, the language invariant is defined by "these bytes represent some value", but for pointers, the language invariant imposes extra requirements.
```

spec/lang/representation.md
```diff
@@ -1,52 +1,57 @@
...
+This *[representation relation]* for a type is defined in the following.
+`decode` converts a list of bytes into a (potentially not well-formed) value; this operation can already fail if the byte list is not a valid encoding for the given type.
+The values are always checked after decoding to be [well-formed][well-formed-value] by `check_value`, since some types also have additional runtime constraints.
+`encode` inverts `decode`; it will always succeed for [well-formed][well-formed-value] inputs.
+
+`encode`, `decode` and `check_value` satisfy some [properties](#generic-properties), which the specification must ensure, i.e. violating these properties is a spec bug.
+However, they can also make some assumptions which the specification ensures:
+The types these are called with/on are all [well-formed](well-formed.md#well-formed-layouts-and-types) and sized, as unsized types cannot be represented as values.
+`encode` can also assume the value is [well-formed][well-formed-value] for the type,
+and `decode` can assume the size of the byte list matches the size of the type.
```
RalfJung, line 16:
```suggestion
The representation relation is defined by two functions, which are duals to each other: `decode` and `encode`.
`decode` converts a list of bytes into a (potentially not well-formed) value; this operation can fail if the byte list is not a valid representation for the given type.
The decoded values must still be checked to be [well-formed][well-formed-value] by `check_value`, since some types also have additional runtime constraints.
`encode` inverts `decode`; it will always succeed for [well-formed][well-formed-value] inputs.

`encode`, `decode` and `check_value` satisfy some [properties](#generic-properties), which the specification must ensure, i.e. violating these properties is a spec bug.
However, they can also make some assumptions which the specification ensures:
The types these are called with/on are all [well-formed](well-formed.md#well-formed-layouts-and-types) and sized, as unsized types cannot be represented as values.
`encode` can also assume the value is [well-formed][well-formed-value] for the type,
and `decode` can assume the length of the byte list matches the size of the type.
```

spec/lang/representation.md
```diff
@@ -405,22 +402,157 @@ Unsized types do not have values and thus there is no representation relation.
...
-        panic!("tried to endoce a slice")
+        panic!("encode of Type::Slice")
+    }
+}
+```
+
+## Well-formed values
+
+We call a value `val` *well-formed* for a type `ty` if `machine.check_value(val, ty).is_ok()`.
+The specification ensures (statically or dynamically) that all values in MiniRust are always well-formed, the only exception being values returned from `decode`, which are sanitized by [typed memory accesses](#Typed-memory-accesses).
```
RalfJung, line 416:
```suggestion
The specification ensures (statically or dynamically) that all values in MiniRust are always well-formed, the only exception being values returned from `decode`, which  must be passed to `check_value` before being passed to any other part of the specification.
```

spec/lang/representation.md
```diff
@@ -405,22 +402,157 @@ Unsized types do not have values and thus there is no representation relation.
...
+Violating this when running a well-formed program is a spec bug.
+Therefore `encode` and other expressions can assume well-formed inputs as well.
```
RalfJung, line 418:
```suggestion
Therefore, `encode` and other expressions can assume well-formed inputs as well.
```

spec/lang/representation.md
```diff
@@ -405,22 +402,157 @@ Unsized types do not have values and thus there is no representation relation.
...
+
+In particular must a load instruction cause UB, when the sequence of bytes does not correspond to any representation of a well-formed value.
```
RalfJung, line 420:
```suggestion
In particular, a load instruction causes UB when the sequence of bytes does not correspond to a representation of a well-formed value.
```

spec/lang/representation.md
```diff
@@ -405,22 +402,157 @@ Unsized types do not have values and thus there is no representation relation.
...
+Generally, this is already dealt with by `decode`: if decoding succeeds, the representation is well-formed.
+However, pointer types have further runtime constraints: specifically, [safe pointers][ptr_type] require the address to be aligned, dereferenceable and non-null, and they require the pointee type to be inhabited.
+This is separated from `decode`, because checking alignment or the well-formedness of a trait object pointer needs vtable information, and dereferenceablity need the allocations.
```
RalfJung, line 423:
```suggestion
This is separated from `decode`, because checking alignment or the well-formedness of a trait object pointer needs vtable information, and dereferenceablity need information about the current state of memory.
```

spec/lang/representation.md
```diff
@@ -405,22 +402,157 @@ Unsized types do not have values and thus there is no representation relation.
...
+Dropping dereferenceablity as a requirement to unify these steps was discussed, see [this discussion](https://github.com/rust-lang/unsafe-code-guidelines/issues/77), but for trait objects this separation is desireable.
+Therefore usages for `decode` will either follow up by calling `check_value` to raise UB or reason about why this is statically ensured already. 
+
+So types like `&!`, have many byte lists which decode without problems, but none are *well-formed*:
+when the pointee type is uninhabited (as given by the pointee type), there exists no valid reference to that type.
```
RalfJung, line 428:
```suggestion
when the pointee type is uninhabited, there exists no valid reference to that type.
```

spec/lang/representation.md
```diff
@@ -405,22 +402,157 @@ Unsized types do not have values and thus there is no representation relation.
...
+            }
+            (Value::Bool(_), Type::Bool) => {},
+            (Value::Ptr(ptr), Type::Ptr(ptr_ty)) => {
+                ensure_else_ub(ptr_ty.meta_kind().matches(ptr.metadata), "Value::Ptr: invalid metadata")?;
+                ensure_else_ub(ptr.thin_pointer.addr.in_bounds(Unsigned, M::T::PTR_SIZE), "Value::Ptr: pointer out-of-bounds")?;
+
+                // Safe pointer, i.e. references, boxes
+                if let Some(layout) = ptr_ty.safe_pointee() {
+                    let size = layout.size.compute(ptr.metadata);
+                    // The total size of slices must be at most `isize::MAX`.
```
RalfJung, line 454:
```suggestion
                    // The total size must be at most `isize::MAX`.
```

spec/lang/representation.md
```diff
@@ -405,22 +402,157 @@ Unsized types do not have values and thus there is no representation relation.
...
+                    ensure_else_ub(data.len() == size.bytes(), "Value::Union: invalid chunk data")?;
+                }
+            }
+            (Value::Variant { discriminant, data }, Type::Enum { variants, .. }) => {
+                let Some(variant) = variants.get(discriminant) else {
+                    throw_ub!("Value::Variant: invalid discrimant");
+                };
+                self.check_value(data, variant.ty)?;
+            }
+            (_, Type::Slice { .. }) => throw_ub!("Value: slices cannot be represented as values"),
```
RalfJung, line 500: Shouldn't this be `panic` since such a value is not supposed to even get here?

spec/lang/representation.md
```diff
@@ -405,22 +402,157 @@ Unsized types do not have values and thus there is no representation relation.
...
+
+        ret(())
+    }
+}
+```
+
+- TODO: Do we really want to special case references to uninhabited types? Do we somehow want to require more, like pointing to a valid instance of the pointee type?
+  (The latter would not even be possible with the current pointee information in MiniRust.)
+  Also see [this discussion](https://github.com/rust-lang/unsafe-code-guidelines/issues/77).
+- TODO: Inhabitedness and non-nullness are (the only two) conditions that could be checked in `decode` itself. Should we?
```
RalfJung, line 512:
```suggestion
- TODO: Inhabitedness, alignment, and non-nullness could be checked in `decode` itself. Should we?
```

Indeximal: I removed alignment there, because it can only be checked for sized pointees.

RalfJung:
Okay, please squash the commits and then this can land. :)

PR #245, opened 2024-11-01
ComputeSize and ComputeAlign UnOps

Indeximal:
This PR replaces #238 to instead contain a `ComputeSize<ty>(metadata)`, which can also be used to model `size_of_val_raw`, in addition to `size_of_val`, and actually `size_of` with unit metadata as well.
Analogously `ComputeAlign<ty>(metadata)` has been defined, but is fairly uninteresting until trait objects land.

This PR is stacked on #241 and only the last commit (08c3021db44b45be90b3f3b70e284d42a0315e9b) is new.


RalfJung:
The last commit looks good. :)

Indeximal:
After #241 this is ready for review and merge as well.

spec/lang/well-formed.md
```diff
@@ -367,12 +367,12 @@ impl ValueExpr {
...
-                        // TODO(UnsizedTypes): Support raw pointers as well.
-                        ensure_wf(
-                            matches!(operand, Type::Ptr(PtrType::Ref { .. })),
-                            "UnOp::SizeOfVal: invalid operand: not a reference"
-                        )?;
+                    ComputeSize(ty) | ComputeAlign(ty) => {
+                        // A thin pointer can also be the target type, with unit metadata.
+                        let meta_ty = ty.meta_kind().ty::<T>().unwrap_or_else(unit_type);
+                        if operand != meta_ty {
+                            throw_ill_formed!("UnOp::ComputeSize|ComputeAlign: invalid right type: not metadata of type");
```
RalfJung, line 374: This is an un op, so "right" makes little sense I think? "invalid operand type" or something like that should work.

tooling/minitest/src/tests/compute_align.rs
```diff
@@ -0,0 +1,198 @@
...
+        );
+        f.assume(eq(
+            compute_align(<[u32]>::get_type(), get_metadata(slice_ptr)),
+            const_int(4_usize),
+        ));
+
+        let arr = f.declare_local::<[u8; 312]>();
+        f.storage_live(arr);
+        let slice_ptr = construct_wide_pointer(
+            addr_of(arr, <&()>::get_type()),
```
RalfJung, line 143: Why `&()`? Isn't this a `&[u8; 312]`?

tooling/minitest/src/tests/compute_align.rs
```diff
@@ -0,0 +1,198 @@
+use crate::*;
+
+/// Helper which builds code to assert that compute_align with given type gives the expected align.
```
RalfJung, line 3: The comment should say that this is for *sized* types only. (same in `compute_size.rs`)

tooling/minitest/src/tests/compute_align.rs
Indeximal: Fair, it would be more accurate, but we're casting again immediately afterwards, therefore this just makes it more verbose. Also this notion is already in use in slice tests.

RalfJung: In just find the wrong type makes the test harder to read. So if it's not too hard, would be good to use more accurate types.

RalfJung:
This is still not rebased on current master it seems? The [diff](https://github.com/minirust/minirust/pull/245/files) is too big. Please rebase.

Indeximal:
Will do, I waited for you review, so that GitHub can still nicely show the newest changes.

PR #246, opened 2024-11-05
Unsized structs

Indeximal:
discussion playground for unsized structs/tuples.


spec/lang/types.md
```diff
@@ -30,13 +30,14 @@ pub enum Type {
...
+        // (FIXME: this does not respect `repr(packed)` and `repr(align)` and I don't see an easy way to incorporate this).
         fields: Fields,
         /// The total size of the tuple can indicate trailing padding.
         /// Must be large enough to contain all fields.
-        size: Size,
-        /// Total alignment of the tuple. Due to `repr(packed)` and `repr(align)`,
+        /// The total alignment of the tuple. Due to `repr(packed)` and `repr(align)`,
         /// this is independent of the fields' alignment.
-        align: Align,
+        layout: LayoutStrategy,
```
RalfJung, line 40: Let's avoid the redundancy (and O(n²) complexity) here by just storing what is needed for "this layer" of the type.

Indeximal:
I have now reworked some of the definitions and could benefit from some high-level feedback before continuing the refactoring.

spec/lang/types.md
```diff
@@ -30,13 +30,12 @@ pub enum Type {
...
-        fields: Fields,
-        /// The total size of the tuple can indicate trailing padding.
-        /// Must be large enough to contain all fields.
-        size: Size,
-        /// Total alignment of the tuple. Due to `repr(packed)` and `repr(align)`,
-        /// this is independent of the fields' alignment.
-        align: Align,
+        sized_fields: Fields,
+        /// A last field (in terms of offset) may contain an unsized type,
+        /// then its offset is given by rounding the `head_end` of `sized_head_layout` up to its nearest alignment.
```
RalfJung, line 35:
```suggestion
        /// then its offset is given by rounding the `head_end` of `sized_head_layout` up to the alignment
        /// of this type.
```

spec/lang/types.md
```diff
@@ -178,6 +185,19 @@ impl Type {
 And we also define how to compute the actual size and alignment.
 
 ```rust
+impl TupleHeadLayout {
+    // This is one function to avoid quadratic computations.
```
RalfJung, line 189: This comment isn't very clear.

spec/lang/types.md
```diff
@@ -178,6 +185,19 @@ impl Type {
...
+    pub fn compute_size_align(self, tail_size: Size, tail_align: Align) -> (Size, Align) {
+        // TODO: support packed self
+        let align = tail_align.max(self.min_align);
```
RalfJung, line 192: Isn't it a bug if tail_align < min_align?

spec/mem/pointer.md
```diff
@@ -99,13 +99,30 @@ pub struct PointeeInfo {
     pub unpin: bool,
 }
 
+pub struct TupleHeadLayout {
+    /// The offset where the tail starts.
+    /// Also determines the size together with the tail size and padding to satisfy alignment.
+    head_end: Offset,
+
+    /// The minimal alignment of the head.
+    min_align: Align,
```
RalfJung, line 108: `min_` seems redundant, alignment is always a lower bound (it's always okay to align something more).

RalfJung (commented):
The structure looks good to me. Do you have any questions for any specific part of it?

spec/lang/types.md
Indeximal: Actually I think I noticed a further-reaching problem, that since `compute_size` calls `tail.compute_size` and `tail.compute_align`, I think computing just the size might lead to exponentially (in the number of layers) many calls to the `compute_align` of the last tail.
So I will introduce a `LayoutStrategy::compute_size_align -> (Size, Align)` and have the `compute_size` be simply `compute_size_align.0`.

spec/lang/types.md
Indeximal: Why? A struct can contain 8 aligned fields, but the tail might only need to be 1 aligned.

RalfJung: Ah of course.

spec/mem/pointer.md
RalfJung: Maybe call it `head_align`?

spec/lang/types.md
RalfJung: Let's call it `compute_size_and_align`, that's also how this is called in rustc.

Indeximal:
Is there no function for rounding an integer to the multiple of an alignment yet?

RalfJung:
You probably want the *next* multiple?

Do you mean this? https://doc.rust-lang.org/nightly/nightly-rustc/rustc_abi/struct.Align.html#method.restrict_for_offset
If we don't have this on our `Align` type, we should add it.

Indeximal:
Yes, the "next", I meant `next_multiple_of` and no, it appears that our `Int` doesn't have this implemented.
I have not quite understood what `restict_for_offset` would be useful for.

spec/lang/step/terminators.md
```diff
@@ -85,14 +85,17 @@ fn check_abi_compatibility(
         (Type::Ptr(_), Type::Ptr(_)) =>
             // The kind of pointer and pointee details do not matter for ABI.
             true,
-        (Type::Tuple { fields: caller_fields, size: caller_size, align: caller_align },
-         Type::Tuple { fields: callee_fields, size: callee_size, align: callee_align }) =>
+        (Type::Tuple { sized_fields: caller_fields, sized_head_layout: caller_head_layout, unsized_field: None },
+         Type::Tuple { sized_fields: callee_fields, sized_head_layout: callee_head_layout, unsized_field: None }) => {
```
RalfJung, line 89: This means we entirely reject unsized tuples here, right?
That's fine but the comment on the `_ =>` arm needs to be updated.

spec/lang/well-formed.md
```diff
@@ -523,12 +523,17 @@ impl PlaceExpr {
             }
             Field { root, field } => {
                 let root = root.check_wf::<T>(locals, prog)?;
-                let (_offset, field_ty) = match root {
-                    Type::Tuple { sized_fields, unsized_field: Some(unsized_field), .. } if field == sized_fields.len() => unsized_field,
+                let field_ty = match root {
+                    Type::Tuple { sized_fields, unsized_field, .. } if field == sized_fields.len() => {
```
RalfJung, line 535: Please add a FIXME saying we'd like to match `unsized_field` on `Some` here but specr currently does not support that.

tooling/miniutil/src/fmt/ty.rs
```diff
@@ -65,24 +65,39 @@ fn fmt_meta_kind(kind: PointerMetaKind) -> String {
...
         LayoutStrategy::Sized(size, align) =>
             format!("size={}, align={}", size.bytes(), align.bytes()),
         LayoutStrategy::Slice(size, align) =>
             format!("size={}*len, align={}", size.bytes(), align.bytes()),
         LayoutStrategy::TraitObject(..) => "size,align={unknown}".into(),
-    };
+        LayoutStrategy::Tuple { head, tail } =>
+            format!(
+                "head=(end={}, align={}), tail=({}))",
+                head.end.bytes(),
```
RalfJung, line 78: This has more closing parentheses than opening ones?

tooling/miniutil/src/fmt/ty.rs
```diff
@@ -89,8 +89,8 @@ fn fmt_pointee_info(pointee: PointeeInfo) -> String {
         false => ", uninhabited",
     };
     let freeze_str = match pointee.freeze {
-        true => ", freeze",
-        false => "",
+        true => "",
+        false => ", unfreeze",
```
RalfJung, line 93: If you think always printing `freeze` is too verbose, please use `!Freeze` (and `!Unpin`) to make things slightly more standard-looking.

spec/lang/types.md
```diff
@@ -197,18 +197,30 @@ And we also define how to compute the actual size and alignment.
...
+        }
+    }
+
+    /// Where the tail starts, given the aligment of the tail type.
     pub fn tail_offset(self, tail_align: Align) -> Offset {
-        // TODO: support packed self
-        let tail_offset = Size::from_bytes(self.end.bytes().next_multiple_of(tail_align.bytes())).unwrap();
+        let capped_tail_align = self.capped_tail_align(tail_align);
+        // `self.end.align_to(capped_tail_align)`.
+        let tail_offset = Size::from_bytes(self.end.bytes().next_multiple_of(capped_tail_align.bytes())).unwrap();
```
RalfJung, line 212: Might be worth actually adding `align_to` to our `Align` type?

RalfJung (commented):
This is looking better and better. :) What is the current status?

spec/lang/types.md
Indeximal: To the `Size` type, right?

RalfJung: Ah right. With an `Align` argument.

RalfJung:
There are conflicts with master, please rebase to resolve those.

spec/lang/types.md
```diff
@@ -182,6 +196,40 @@ impl Type {
...
+            Some(packed_align) => tail_align.min(packed_align),
+            None => tail_align,
+        }
+    }
+
+    /// Where the tail starts, given the aligment of the tail type.
+    pub fn tail_offset(self, tail_align: Align) -> Offset {
+        let capped_tail_align = self.capped_tail_align(tail_align);
+        // `self.end.align_to(capped_tail_align)`.
+        let tail_offset = Size::from_bytes(self.end.bytes().next_multiple_of(capped_tail_align.bytes())).unwrap();
```
RalfJung, line 212: This should use `Size::align_to` now

spec/lang/types.md
```diff
@@ -182,6 +196,40 @@ impl Type {
...
+        let align = capped_tail_align.max(self.align);
+        let tail_offset = self.tail_offset(tail_align);
+        let end = tail_offset + tail_size;
+        let size = Size::from_bytes(end.bytes().next_multiple_of(align.bytes())).unwrap();
+        (size, align)
+    }
+
+    /// Returns the size and alignment of the tuple when there is no tail.
+    pub fn head_size_and_align(self) -> (Size, Align) {
+        self.full_size_and_align(Size::ZERO, Align::ONE)
```
RalfJung, line 229: Isn't this just `(self.end, self.align)`?

spec/lang/well-formed.md
```diff
@@ -47,6 +62,8 @@ impl LayoutStrategy {
                 ensure_wf(size.bytes() % align.bytes() == 0, "check_aligned: element size not a multiple of alignment")?;
             }
             LayoutStrategy::TraitObject(..) => (),
+            // The size and align computation aligns the size.
+            LayoutStrategy::Tuple {..} => (),
```
RalfJung, line 66: Shouldn't this recurse into the tail?

spec/mem/pointer.md
```diff
@@ -94,6 +94,19 @@ pub struct PointeeInfo {
...
+pub struct TupleHeadLayout {
+    /// The offset where the head ends and tail starts.
+    /// Also determines the size together with the padding to satisfy alignment and tail size.
+    pub end: Offset,
+
+    /// The alignment of the head. This is the maximal alignment of any sized field and capped to the packed alignment.
+    pub align: Align,
+
+    /// If this is `Some(a)` the alignment of the tail will be capped at this value.
+    pub packed_align: Option<Align>,
```
RalfJung, line 107:
```suggestion
    /// Must be `None` for sized tuples.
    pub packed_align: Option<Align>,
```

spec/mem/pointer.md
```diff
@@ -94,6 +94,19 @@ pub struct PointeeInfo {
     pub unpin: bool,
 }
 
+/// Describes what is needed to define the layout of the sized head of a tuple `(head, tail)`.
```
RalfJung, line 97:
```suggestion
/// Describes what is needed to define the layout of the sized head of a tuple `(head.., tail)`.
```

spec/lang/well-formed.md
```diff
@@ -267,9 +293,10 @@ impl ValueExpr {
                 t.check_wf::<T>()?;
 
                 match t {
-                    Type::Tuple { fields, .. } => {
-                        ensure_wf(exprs.len() == fields.len(), "ValueExpr::Tuple: invalid number of tuple fields")?;
-                        for (e, (_offset, ty)) in exprs.zip(fields) {
+                    Type::Tuple { sized_fields, unsized_field, .. } => {
+                        ensure_wf(unsized_field.is_none(), "ValueExpr::Tuple: aggregating an unsized tuple")?;
```
RalfJung, line 297:
```suggestion
                        ensure_wf(unsized_field.is_none(), "ValueExpr::Tuple: constructing an unsized tuple value")?;
```

spec/lang/types.md
```diff
@@ -154,7 +154,17 @@ impl Type {
             Bool => Sized(Size::from_bytes_const(1), Align::ONE),
             Ptr(p) if p.meta_kind() == PointerMetaKind::None => Sized(T::PTR_SIZE, T::PTR_ALIGN),
             Ptr(_) => Sized(libspecr::Int::from(2) * T::PTR_SIZE, T::PTR_ALIGN),
-            Tuple { size, align, .. } | Union { size, align, .. } | Enum { size, align, .. } => Sized(size, align),
+            Union { size, align, .. } | Enum { size, align, .. } => Sized(size, align),
+            Tuple { sized_head_layout, unsized_field, .. } => match unsized_field {
+                None => {
+                    let (size, align) = sized_head_layout.full_size_and_align(Size::ZERO, Align::ONE);
```
RalfJung, line 160: This could use `head_size_and_align`, I think?

spec/lang/step/terminators.md
```diff
@@ -85,14 +85,19 @@ fn check_abi_compatibility(
...
             // The kind of pointer and pointee details do not matter for ABI.
             true,
-        (Type::Tuple { fields: caller_fields, size: caller_size, align: caller_align },
-         Type::Tuple { fields: callee_fields, size: callee_size, align: callee_align }) =>
+        (Type::Tuple { sized_fields: caller_fields, sized_head_layout: caller_head_layout, unsized_field: caller_unsized_field },
+         Type::Tuple { sized_fields: callee_fields, sized_head_layout: callee_head_layout, unsized_field: callee_unsized_field }) => {
+            let (caller_size, caller_align) = caller_head_layout.head_size_and_align();
+            let (callee_size, callee_align) = callee_head_layout.head_size_and_align();
+            caller_unsized_field.is_none() &&
+            callee_unsized_field.is_none() &&
```
RalfJung, line 93: These could be assertions, right? Unsized types cannot currently be passed as arguments.

spec/lang/step/expressions.md
```diff
@@ -271,16 +271,26 @@ impl<M: Memory> Machine<M> {
...
+        let ((offset, field_ty), retain_metadata) = match ty {
+            Type::Tuple { sized_fields, .. } if field < sized_fields.len() => (sized_fields[field], false),
+            Type::Tuple { sized_fields, unsized_field, sized_head_layout } if field == sized_fields.len() => {
+                let tail_ty = unsized_field.expect("field projection to non-existing tail");
+                let tail_align = self.compute_align(tail_ty.layout::<M::T>(), root.ptr.metadata);
+                let offset = sized_head_layout.tail_offset(tail_align);
+                ((offset, tail_ty), true)
+            }
+            Type::Union { fields, .. } => (fields[field], false),
             _ => panic!("field projection on non-projectable type"),
```
RalfJung, line 283: This panic can now also be triggered when the index is too big, I think?

Please have only a single `Type::Tuple` case above and handle all the indexing logic there.

spec/lang/well-formed.md
```diff
@@ -503,11 +531,18 @@ impl PlaceExpr {
             }
             Field { root, field } => {
                 let root = root.check_wf::<T>(locals, prog)?;
-                let (_offset, field_ty) = match root {
-                    Type::Tuple { fields, .. } | Type::Union { fields, .. } => {
+                let field_ty = match root {
+                    Type::Tuple { sized_fields, unsized_field, .. } if field == sized_fields.len() => {
```
RalfJung, line 535: This logic is duplicated with the one for actually executing a field projection, I think.

Seems better to have a helper method like `get_field` or so that implements this logic just once. You might want to add a `struct TupleType` holding the fields of `Type::Tuple` to have some natural place to put this method.

spec/lang/types.md
Indeximal: No, because we align the end first. It is `self.end.align_to(self.align) though if you consider this nicer.

Indeximal:
Rebased and addressed issues (only the last commit is new)

spec/lang/step/expressions.md
```diff
@@ -272,15 +272,31 @@ impl<M: Memory> Machine<M> {
     fn eval_place(&mut self, PlaceExpr::Field { root, field }: PlaceExpr) -> Result<(Place<M>, Type)> {
         let (root, ty) = self.eval_place(root)?;
         let (offset, field_ty) = match ty {
-            Type::Tuple { fields, .. } => fields[field],
+            Type::Tuple { sized_fields, unsized_field, sized_head_layout } => {
+                if field < sized_fields.len() {
+                    sized_fields[field]
+                } else if field == sized_fields.len() {
+                    let tail_ty = unsized_field.expect("field projection to non-existing tail");
```
RalfJung, line 279:
```suggestion
                    let tail_ty = unsized_field.expect("field projection to non-existing unsized tail");
```

spec/lang/step/terminators.md
```diff
@@ -90,8 +90,9 @@ fn check_abi_compatibility(
          Type::Tuple { sized_fields: callee_fields, sized_head_layout: callee_head_layout, unsized_field: callee_unsized_field }) => {
             let (caller_size, caller_align) = caller_head_layout.head_size_and_align();
             let (callee_size, callee_align) = callee_head_layout.head_size_and_align();
-            caller_unsized_field.is_none() &&
-            callee_unsized_field.is_none() &&
+            // FIXME: Matching on `unsized_field: None` would be nicer, but specr currently does not support that.
```
RalfJung, line 93: I don't agree with this, having a single block handling al `Tuple` cases seems quite nice to me.

spec/lang/types.md
```diff
@@ -220,13 +219,13 @@ impl TupleHeadLayout {
...
         let end = tail_offset + tail_size;
-        let size = Size::from_bytes(end.bytes().next_multiple_of(align.bytes())).unwrap();
+        let size = end.align_to(align);
         (size, align)
     }
 
     /// Returns the size and alignment of the tuple when there is no tail.
     pub fn head_size_and_align(self) -> (Size, Align) {
-        self.full_size_and_align(Size::ZERO, Align::ONE)
+        (self.end.align_to(self.align), self.align)
```
RalfJung, line 228: Please add a comment saying this is the same as `full_size_and_align(ZERO, ONE)`.

spec/lang/well-formed.md
```diff
@@ -532,14 +532,18 @@ impl PlaceExpr {
...
+                            };
+                            unsized_ty
+                        } else {
                             throw_ill_formed!("PlaceExpr::Field: invalid field");
-                        };
-                        unsized_ty
-                    }
-                    Type::Tuple { sized_fields: fields, .. } | Type::Union { fields, .. } => {
+                        }
+                    } Type::Union { fields, .. } => {
```
RalfJung, line 546: Please add a newline for the next match arm

spec/mem/pointer.md
```diff
@@ -94,6 +94,20 @@ pub struct PointeeInfo {
     pub unpin: bool,
 }
 
+/// Describes what is needed to define the layout of the sized head of a tuple `(head.., tail)`.
+pub struct TupleHeadLayout {
+    /// The offset where the head ends and tail starts.
+    /// Also determines the size together with the padding to satisfy alignment and tail size.
```
RalfJung, line 100:
```suggestion
    /// This is the end of the last sized field; it is *not* necessarily aligned to `align`.
```

spec/lang/step/terminators.md
Indeximal: Fair, I was searching for a place where it would be nicer, but if we always want to match only once on tuples, this comment doesn't have a good place anymore.

RalfJung:
Looks good. :) Please squash.

PR #248, opened 2024-12-17
Trait Object support for minimize

Indeximal:
This PR adds dynamic dispatch support to minimize. Compared to rustc's const eval, there are less special cases handled, and the receiver type is not smartly adjusted, but it doesn't seem to be a problem.


Indeximal:
I believe this PR is ready for a review. Particularly interesting would be suggestions for special cases to write more interesting tests.

tooling/minimize/src/program.rs
```diff
@@ -75,12 +83,19 @@ impl<'tcx> Ctxt<'tcx> {
         let start = FnName(Name::from_internal(number_of_fns as _));
         self.functions.insert(start, mk_start_fn(0));
 
+        // since all vtables for the same trait name use the same trait, they also have the same methods defined.
```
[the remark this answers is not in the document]
Indeximal: The vtable map will have multiple entries for the same trait, but only one of them can end up in the traits map. Since they were all generated from the same trait, the also have the same methods defined, overriding is a No-Op.

tooling/minimize/src/vtable.rs
```diff
@@ -0,0 +1,61 @@
...
+            let fresh_name = VTableName(Name::from_internal(self.vtable_map.len() as _));
+            let vtable = self.generate_vtable(ty, trait_);
+            self.vtable_map.insert((ty, trait_), (fresh_name, vtable));
+            fresh_name
+        }
+    }
+
+    /// Generates a vtable for the given type and trait.
+    fn generate_vtable(&mut self, ty: rs::Ty<'tcx>, trait_name: TraitName) -> VTable {
+        // This is using linear search because we only have a one-way mapping.
```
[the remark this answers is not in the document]
Indeximal: generate_vtables cannot easily take a trait predicate list as then the unsizing coercion cannot match on MiniRust types anymore, making that part uglier instead.
Additionally, I am unsure how I could get the trait indices without providing a self type, which would be necessary to fill the traits map.

tooling/minimize/src/bb.rs
```diff
@@ -464,8 +465,25 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
                 })
                 .collect();
 
+            // Distinguish direct function calls or dynamic dispatch on a trait object.
+            let callee = if let rs::InstanceKind::Virtual(_trait, method) = instance.def {
+                // FIXME: This does not implement all receivers as allowed by `std::ops::DispatchFromDyn`,
+                // it also doesn't properly adjust the type, but instead uses the raw pointer given by `GetThinPointer`.
+                // However, this doesn't seem to be an issue.
```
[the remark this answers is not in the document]
Indeximal: If I understand this right, the adjustment has to recurse into the arbitrary caller type, extract the thin pointer there and then build tuples again, such that the caller type is abi-compatible with the trait receiver?
I have for now adjusted the comment and will most likely leave this for further work.

tooling/minimize/tests/pass/trait_object.rs
```diff
@@ -0,0 +1,27 @@
+trait A {
+    fn foo(&self) -> usize;
```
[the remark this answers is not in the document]
Indeximal: `core[d68a]::alloc::layout::{impl#0}::is_size_align_valid) does not have a "optimized_mir"` or `alloc[66d4]::alloc::{extern#0}::__rust_dealloc) does not have a "optimized_mir"` or various other #175 related issues prevent such a test.

tooling/minimize/src/program.rs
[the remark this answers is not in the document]
RalfJung: The traits map shouldn't be generated from the vtable map... this is related to the the structure proposed [here](https://github.com/minirust/minirust/pull/248#discussion_r1903884436).

tooling/minimize/src/vtable.rs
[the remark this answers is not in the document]
RalfJung: > the unsizing coercion cannot match on MiniRust types anymore, making that part uglier instead.

It probably should match on rustc types.

> Additionally, I am unsure how I could get the trait indices without providing a self type, which would be necessary to fill the traits map.

I don't understand that sentence.

tooling/minimize/src/bb.rs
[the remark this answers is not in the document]
RalfJung: The caller type must already be ABI-compatible with the pointer, so we don't re-build any tuples, but we do have to recurse down until we find the pointer yes.

tooling/minimize/tests/pass/trait_object.rs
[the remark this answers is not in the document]
RalfJung: Ah right, we can't have `Box` yet since we'll have to figure out first how we wire up the allocator.

tooling/minimize/src/vtable.rs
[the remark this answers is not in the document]
Indeximal: Currently, I generate the `Set<TraitMethodName>` by looking at `VtblEntry`. But does this work without first setting `trait_.with_self_ty(self.tcx, ty)`?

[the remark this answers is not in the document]
RalfJung: Hm... it should be possible to use the same logic as [here](https://github.com/rust-lang/rust/blob/9a1c5eb5b385adb3cd04af0049cbf5c225cefdc3/compiler/rustc_ty_utils/src/instance.rs#L239-L248) to get the list of valid method indices without having a concrete self type...
`own_existential_vtable_entries` is the easy one, you just need a DefId for that which you can fairly easily get from `principal()`. `first_method_vtable_slot` is trickier, it seems you need to construct a `TraitRef` where `Self` is `dyn Trait`... I'll ask the t-types folks for advice.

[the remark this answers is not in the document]
RalfJung: We got an answer there. :D  This code snippet is the key:
```rust
let ty::Dynamic(data, _, _) = *ty.kind() else { panic!(); }
let existential_trait_ref = tcx.instantiate_bound_regions_with_erased(data.principal().unwrap());
let trait_ref = existential_trait_ref.with_self_ty(tcx, ty);
```
This `trait_ref` can then be passed to `first_method_vtable_slot`, and its `DefId` passed to `own_existential_vtable_entries`.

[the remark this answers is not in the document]
Indeximal: Okay, I have refactored the code now.

tooling/minimize/src/program.rs
```diff
@@ -6,6 +6,14 @@ pub struct Ctxt<'tcx> {
...
 
+    /// maps Rust traits to MiniRust TraitNames and those to sets of methods.
+    pub trait_map: HashMap<&'tcx rs::List<rs::PolyExistentialPredicate<'tcx>>, TraitName>,
+    pub traits: Map<TraitName, Set<TraitMethodName>>,
+
+    /// maps Rust vtables to MiniRust VTableNames and those to Vtables.
+    pub vtable_map:
+        HashMap<(rs::Ty<'tcx>, &'tcx rs::List<rs::PolyExistentialPredicate<'tcx>>), VTableName>,
+    pub vtables: Map<VTableName, VTable>,
     /// Stores which AllocId evaluates to which GlobalName.
```
RalfJung, line 17:
```suggestion

    /// Stores which AllocId evaluates to which GlobalName.
```

tooling/minimize/src/bb.rs
```diff
@@ -464,8 +465,26 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
                 })
                 .collect();
 
+            // Distinguish direct function calls or dynamic dispatch on a trait object.
+            let callee = if let rs::InstanceKind::Virtual(_trait, method) = instance.def {
+                // FIXME: This does not implement all receivers as allowed by `std::ops::DispatchFromDyn`.
+                // It doesn't (and also can't) properly adjust the type, however, MiniRust only requires the argument
```
RalfJung, line 471: Why can't it do that?

tooling/minimize/src/rvalue.rs
```diff
@@ -289,33 +289,53 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
+                            (
+                                rs::TyKind::Dynamic(data_a, _, rs::Dyn),
+                                rs::TyKind::Dynamic(data_b, _, rs::Dyn),
+                            ) => {
+                                // MIR building generates odd NOP casts, prevent them from causing unexpected trouble.
+                                // See <https://github.com/rust-lang/rust/issues/128880>.
+                                if data_a == data_b {
+                                    operand
+                                } else {
+                                    rs::span_bug!(span, "Unsupported super trait unsizing coercion")
```
RalfJung, line 323: Please open an issue to track our missing support for unsizing coercions

tooling/minimize/src/vtable.rs
```diff
@@ -0,0 +1,108 @@
+use crate::*;
+
+impl<'tcx> Ctxt<'tcx> {
+    /// Gets the vtable name for the given type and trait object or creates it if it doesn't exist yet.
+    /// `trait_obj_ty` must be of kind [`rs::TyKind::Dynamic`].
+    pub fn get_vtable(&mut self, ty: rs::Ty<'tcx>, trait_obj_ty: rs::Ty<'tcx>) -> VTableName {
+        let rs::TyKind::Dynamic(trait_, _, _) = *trait_obj_ty.kind() else {
```
RalfJung, line 7:
```suggestion
        let rs::TyKind::Dynamic(trait_, _, rs::DynKind::Dyn) = *trait_obj_ty.kind() else {
```

tooling/minimize/src/vtable.rs
```diff
@@ -0,0 +1,108 @@
...
+            let vtable = self.generate_vtable(ty, trait_obj_ty);
+            self.vtables.insert(fresh_name, vtable);
+            self.vtable_map.insert((ty, trait_), fresh_name);
+            fresh_name
+        }
+    }
+
+    /// Generates a vtable for the given type and trait object.
+    fn generate_vtable(&mut self, ty: rs::Ty<'tcx>, trait_obj_ty: rs::Ty<'tcx>) -> VTable {
+        let rs::TyKind::Dynamic(trait_, _, _) = *trait_obj_ty.kind() else {
```
RalfJung, line 24:
```suggestion
        let rs::TyKind::Dynamic(trait_, _, rs::DynKind::Dyn) = *trait_obj_ty.kind() else {
```

tooling/minimize/src/vtable.rs
```diff
@@ -0,0 +1,108 @@
...
+            let trait_ref = self.tcx.erase_regions(trait_ref);
+            let entries = self.tcx.vtable_entries(trait_ref);
+
+            entries
+                .iter()
+                .enumerate()
+                .filter_map(|(idx, entry)| self.vtable_entry_to_trait_method(idx as _, entry))
+                .collect()
+        } else {
+            Map::new()
```
RalfJung, line 45: Please add a comment (github is broken and doesn't let me add a suggestion...)
```
// This dyn type has no principal trait, and therefore no methods.
```

RalfJung, line 45: Please add a comment (github is broken and doesn't let me add a suggestion...)
```
// This dyn type has no principal trait, and therefore no methods.
```

tooling/minimize/src/vtable.rs
```diff
@@ -0,0 +1,108 @@
...
+        };
+        let trait_name = self.get_trait_name(trait_obj_ty);
+
+        VTable { trait_name, size, align, methods }
+    }
+
+    /// Generate an method implementation for a vtable entry if it is a method.
+    fn vtable_entry_to_trait_method(
```
RalfJung, line 53: This function is called only once -- I think the code might become a bit clearer even if you just inline it at its only call site.

tooling/minimize/src/vtable.rs
```diff
@@ -0,0 +1,108 @@
...
+            return Set::new();
+        };
+
+        let existential_trait_ref =
+            self.tcx.instantiate_bound_regions_with_erased(trait_.principal().unwrap());
+        // A trait ref with `dyn Trait` as the self type.
+        let trait_ref = existential_trait_ref.with_self_ty(self.tcx, trait_obj_ty);
+        let vtable_base = self.tcx.first_method_vtable_slot(trait_ref);
+
+        // The method names are given by an base offset and the index.
```
RalfJung, line 100:
```suggestion
        // The method names are given by the index into the vtable, which is the base offset
        // plus the index into the "own_existential_vtable_entries" list.
```

tooling/minimize/src/vtable.rs
```diff
@@ -0,0 +1,108 @@
...
+        };
+
+        let Some(princial_def_id) = trait_.principal_def_id() else {
+            // no principal trait means no methods
+            return Set::new();
+        };
+
+        let existential_trait_ref =
+            self.tcx.instantiate_bound_regions_with_erased(trait_.principal().unwrap());
+        // A trait ref with `dyn Trait` as the self type.
```
RalfJung, line 96:
```suggestion
        // A trait ref with `dyn Trait` as the self type.
        // This corresponds to `dyn Trait: Trait`, which matches how the Rust compiler
        // handles a method call on `dyn Trait`.
```

tooling/minimize/src/bb.rs
Indeximal: The proper Type for `&self` is a pointer with the size and alignement of 8. MiniRust can't know this at compile time.

RalfJung: This doesn't quite make sense to me. Where does the 8 come from?

The Rust codegen backend also does this all at compile time so clearly it is possible

tooling/minimize/tests/pass/drop.rs
```diff
@@ -9,6 +9,17 @@ impl Drop for Bomb {
...
+
 fn main() {
+    // Drop once normally
     let _b = Bomb;
+
+    // Then drop as trait object for Shell.
+    let mut bomb = Bomb;
+    unsafe {
+        std::ptr::drop_in_place(&mut bomb as &mut dyn Shell);
+    }; // drop at dyn type
```
RalfJung, line 23:
```suggestion
    // drop at dyn type
    unsafe {
        std::ptr::drop_in_place(&mut bomb as &mut dyn Shell);
    }
```

tooling/minimize/src/bb.rs
```diff
@@ -208,12 +208,39 @@ impl<'cx, 'tcx> FnCtxt<'cx, 'tcx> {
...
+                let (drop_fn, ptr_to_drop) = match ty.kind() {
+                    // For trait objects we must first fetch the drop function dynamically
+                    rs::TyKind::Dynamic(..) => {
+                        let Type::TraitObject(trait_name) = self.translate_ty(ty, span) else {
+                            rs::span_bug!(
+                                span,
+                                "translate_ty for TyKind::Dynamic didn't give a Type::TraitObject"
+                            );
+                        };
+                        let ptr = build::addr_of(
```
RalfJung, line 221:
```suggestion
                        // Compute the wide pointer pointing to `dyn Trait`.
                        let ptr = build::addr_of(
```

RalfJung:
Seems like tests don't work any more when merged with latest master, could you rebase and investigate?

Indeximal:
Was a small fix

PR #249, opened 2024-12-18
Add minimize support for size_of_val & align_of_val

Indeximal:
Based on #234, but now with the operations from #245, added support for the size and align intrinsics of minimize.


RalfJung:
This looks great, thanks. :)

PR #252, opened 2024-12-21
Fix #251: check VTablePtr

### LINKED ISSUE #251: spec bug when looking up vtable from a ptr not pointing to a vtable
The following test panics the interpreter:
```rust
#[test]
fn ub_method_lookup() {
    let mut p = ProgramBuilder::new();

    let mut t_builder = p.declare_trait();
    let meth1 = t_builder.declare_method();
    let _trait_name = p.finish_trait(t_builder);

    let main = {
        let mut f = p.declare_function();
        let x = f.declare_local::<i32>();
        let y = f.declare_local_with_ty(Type::Ptr(PtrType::FnPtr));
        f.storage_live(x);
        f.storage_live(y);
        f.assign(y, vtable_lookup(addr_of(x, Type::Ptr(PtrType::VTablePtr)), meth1));
        f.exit();
        p.finish_function(f)
    };

    let p = p.finish_program(main);
    assert_ub::<BasicMem>(p, "invalid pointer for vtable lookup");
}
```

The problem is that a pointer of type `VTablePtr` doesn't necessarily point to a vtable. This is fine; the bug is in the part of the code that assumes all pointers of this type must point to vtables.

Cc @Indeximal

Indeximal:
My fix for #251 was to do as the comment said, and make sure all VTablePtr must point to an allocated vtable in the language invariant check. Previously this was only done for the metadata, not the pointers themselves.
Additionally, this PR adds trait information for vtablePtr pointer types.

### DIFF
--- a/spec/lang/representation.md
+++ b/spec/lang/representation.md
@@ -136,7 +136,7 @@ impl PointerMetaKind {
         match self {
             PointerMetaKind::None => None,
             PointerMetaKind::ElementCount => Some(Type::Int(IntType::usize_ty::<T>())),
-            PointerMetaKind::VTablePointer(..) => Some(Type::Ptr(PtrType::VTablePtr)),
+            PointerMetaKind::VTablePointer(trait_name) => Some(Type::Ptr(PtrType::VTablePtr(trait_name))),
         }
     }
 }
@@ -162,7 +162,7 @@ impl PtrType {
 impl PointerMetaKind {
     /// Decodes a value to metadata.
     /// The spec will only call this with values which are well formed for `self.ty()`,
-    /// but this may return ill-formed metadata (as defined by `Machine::check_pointer_metadata`), thus needs to be checked.
+    /// but this may return ill-formed metadata (as defined by `Machine::check_ptr_metadata`), thus needs to be checked.
     fn decode_value<M: Memory>(self, value: Value<M>) -> Option<PointerMeta<M::Provenance>> {
         match (self, value) {
             (PointerMetaKind::ElementCount, Value::Int(count)) => Some(PointerMeta::ElementCount(count)),
@@ -173,7 +173,7 @@ impl PointerMetaKind {
     }
 
     /// Encodes metadata as a value.
-    /// The spec ensures this is only called with well-formed metadata (as defined by `Machine::check_pointer_metadata`).
+    /// The spec ensures this is only called with well-formed metadata (as defined by `Machine::check_ptr_metadata`).
     fn encode_as_value<M: Memory>(self, meta: Option<PointerMeta<M::Provenance>>) -> Value<M> {
         match (self, meta) {
             (PointerMetaKind::ElementCount, Some(PointerMeta::ElementCount(count))) => Value::Int(count),
@@ -471,15 +471,13 @@ fn ensure_else_ub(b: bool, msg: &str) -> Result<()> {
 
 impl<M: Memory> Machine<M> {
     /// Defines well-formedness for pointer metadata for a given kind.
-    fn check_pointer_metadata(&self, meta: Option<PointerMeta<M::Provenance>>, kind: PointerMetaKind) -> Result {
+    fn check_ptr_metadata(&self, meta: Option<PointerMeta<M::Provenance>>, kind: PointerMetaKind) -> Result {
         match (meta, kind) {
             (None, PointerMetaKind::None) => {}
             (Some(PointerMeta::ElementCount(num)), PointerMetaKind::ElementCount) =>
                 self.check_value(Value::Int(num), Type::Int(IntType::usize_ty::<M::T>()))?,
             (Some(PointerMeta::VTablePointer(ptr)), PointerMetaKind::VTablePointer(trait_name)) => {
-                // Check that the vtable exists and is for the correct trait
-                let vtable = self.vtable_from_ptr(ptr)?;
-                ensure_else_ub(vtable.trait_name == trait_name, "Value::Ptr: invalid vtable in metadata")?;
+                self.check_ptr(ptr.widen(None), PtrType::VTablePtr(trait_name))?;
             }
             _ => throw_ub!("Value::Ptr: invalid metadata"),
         };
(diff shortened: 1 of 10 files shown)


tooling/minitest/src/tests/trait_object.rs
```diff
@@ -318,6 +318,29 @@ fn ub_wrong_vtable_ty() {
     assert_ub::<BasicMem>(p, "Value::Ptr: invalid vtable in metadata");
 }
 
+#[test]
```
RalfJung, line 260: Please add a doc comment:
```
/// Ensure we correctly report UB when calling vtable_method_lookup on a pointer
/// that does not point to a vtable.
```

spec/lang/representation.md
```diff
@@ -503,6 +503,9 @@ impl<M: Memory> Machine<M> {
             )?;
 
             // However, we do not care about the data stored wherever this pointer points to.
+        } else if let PtrType::VTablePtr = ptr_ty {
+            // Vtable pointers must always point to an allocated vtable
+            self.vtable_from_ptr(ptr.thin_pointer)?;
```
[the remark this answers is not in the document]
RalfJung: Oh I see, this isn't even about the metadata, it is about the "standalone vtbale pointers" which don't exist in Rust.

spec/lang/representation.md
```diff
@@ -503,6 +501,10 @@ impl<M: Memory> Machine<M> {
             )?;
 
             // However, we do not care about the data stored wherever this pointer points to.
+        } else if let PtrType::VTablePtr(trait_name) = ptr_ty {
+            // Vtable pointers must always point to an allocated vtable for the correct trait
```
RalfJung, line 505:
```suggestion
            // This is a "stand-alone" vtable pointer, something that does not exist
            // in surface Rust. Ensure that it points to an allocated vtable for the correct trait.
```

spec/lang/representation.md
```diff
@@ -468,9 +468,7 @@ impl<M: Memory> Machine<M> {
             (Some(PointerMeta::ElementCount(num)), PointerMetaKind::ElementCount) =>
                 self.check_value(Value::Int(num), Type::Int(IntType::usize_ty::<M::T>()))?,
             (Some(PointerMeta::VTablePointer(ptr)), PointerMetaKind::VTablePointer(trait_name)) => {
-                // Check that the vtable exists and is for the correct trait
-                let vtable = self.vtable_from_ptr(ptr)?;
-                ensure_else_ub(vtable.trait_name == trait_name, "Value::Ptr: invalid vtable in metadata")?;
+                self.check_pointer(ptr.widen(None), PtrType::VTablePtr(trait_name))?;
```
RalfJung, line 471: We are pretty inconsistent with "Ptr" vs "Pointer" it seems...

spec/lang/step/operators.md
```diff
@@ -156,12 +156,8 @@ impl<M: Memory> Machine<M> {
         };
         // It is checked in check_value that the vtable is always valid.
         let vtable = self.vtable_lookup()(ptr.thin_pointer);
-        let Some(fn_name) = vtable.methods.get(method) else {
-            // This would be a type error, but since we don't store the trait in the VTablePtr type,
-            // we do not type check this.
-            // FIXME: Store the TraitName in VTablePtr and have this be a program well-formedness requirement.
-            throw_ub!("the referenced vtable does not have an entry for the invoked method");
-        };
+        // Wellformedness ensures the method is declared
```
RalfJung, line 159:
```suggestion
        // Wellformedness ensures `ptr` points to a vtable of the right trait,
        // and hence the method exists.
```

spec/lang/syntax.md
```diff
@@ -431,7 +431,7 @@ Finally, the general structure of programs and functions:
 /// The internal representations of these types do not matter.
 pub struct FnName(pub libspecr::Name);
 pub struct GlobalName(pub libspecr::Name);
-pub struct VTableName(pub libspecr::Name);
+pub struct VTableName(pub TraitName, pub libspecr::Name);
```
RalfJung, line 434: This is very odd. A name shouldn't usually contain another name. What is happening here?

spec/lang/well-formed.md
```diff
@@ -382,13 +384,16 @@ impl ValueExpr {
...
-                        let Type::Ptr(PtrType::VTablePtr) = operand else {
+                    VTableMethodLookup(method) => {
+                        let Type::Ptr(PtrType::VTablePtr(trait_name)) = operand else {
                             throw_ill_formed!("UnOp::VTableMethodLookup: invalid operand: not a vtable pointer");
                         };
 
-                        // We do not statically have enough information to check that the method name exists in the right vtable,
-                        // but this is only a type safety problem, which we don't catch anyways.
+                        let Some(trait_methods) = prog.traits.get(trait_name) else {
+                            throw_ill_formed!("UnOp::VTableMethodLookup: invalid operand: trait of the vtable pointer doesn't exist");
```
RalfJung, line 393: Is this even reachable? I would expect `operand` itself to not be well-formed rather than having a nonsense type.

spec/lang/well-formed.md
```diff
@@ -748,6 +753,16 @@ impl Relocation {
 
 impl Program {
     fn check_wf<T: Target>(self) -> Result<()> {
+        // Check vtables: All vtables for the same trait must have the same method names defined.
```
RalfJung, line 756:
```suggestion
        // Check vtables: All vtables for the same trait must have all trait methods defined.
```

spec/lang/syntax.md
Indeximal: The reason is that in the builder API the constant both needs the vtable name and trait name, which is redundant.
But yes, it is weird.

RalfJung: The builder API should never affect what happens in `spec/`.

spec/lang/well-formed.md
Indeximal: It was reachable because `PtrType::check_wf` and thus `Type::check_wf` didn't take the program as an argument and thus couldn't check this. Fixed now.

Indeximal:
I have reverted the traitname in the vtable name and incorporated other feedback. In particular have I extended `Type::check_wf` to also take the program, to check in both the `TraitObject(trait_name) ` and `PtrType::VTablePtr(trait_name)` that the trait name exists.

Indeximal:
Resolved conflicts

tooling/minitest/src/tests/trait_object.rs
```diff
@@ -402,6 +386,37 @@ fn ill_lookup_wrong_ty() {
...
+    let mut t_builder = p.declare_trait();
+    let meth2 = t_builder.declare_method();
+    let _trait2_name = p.finish_trait(t_builder);
+
+    let f = {
+        let mut f = p.declare_function();
+        let y = f.declare_local_with_ty(Type::Ptr(PtrType::FnPtr));
+        f.storage_live(y);
+        // Ill formed since the vtable doesn't declare the meth2.
+        // (Actually it might still work, since both miniutil and minimize use low entropy method indices).
```
RalfJung, line 407: What do you mean by this?

RalfJung (commented):
Looks great, thanks! Just some minor comments.

tooling/minitest/src/tests/trait_object.rs
Indeximal: It is only checked that some method with the same name exists. However, there are a bunch of homonyms of the method because miniutil uses monotonically increasing numbers and minimize indices.

RalfJung: But we control the entire testcase and it's not very big so I fail to see how that's relevant...? Also the test does report "illformed" so how does it make sense to say "it might still work" while at the same time asserting that it does not work?

RalfJung:
Thanks! Please squash this a little (in particular, "Revert including trait name in vtable name" should not exist, the right representation should be used form the start).

Indeximal:
I've fully squashed, makes little sense to spend time untangling the revert specifically.

PR #253, opened 2025-01-03
Add comparasion of wide pointers

Indeximal:
The TODO was left open to support RelOp on wide pointers. This PR introduces this by using lexicographic ordering and ignoring provenance.
For vtable this uses the address in the pointer metadata, meaning also wide pointers with identical underlying vtables, but differing allocations will not be considered equal.
As far as I know this is the current state in rustc.

### DIFF
--- a/spec/lang/step/operators.md
+++ b/spec/lang/step/operators.md
@@ -321,52 +321,59 @@ impl<M: Memory> Machine<M> {
 
 ```rust
 impl<M: Memory> Machine<M> {
-    fn eval_rel_op(rel: RelOp, left: Int, right: Int) -> (Value<M>, Type) {
+    /// Turns the ordering from the comparasion result into a value, depending on the operation.
+    fn eval_rel_op(rel: RelOp, ord: std::cmp::Ordering) -> (Value<M>, Type) {
         use RelOp::*;
-        let b = match rel {
-            Lt => left < right,
-            Gt => left > right,
-            Le => left <= right,
-            Ge => left >= right,
-            Eq => left == right,
-            Ne => left != right,
+        match rel {
+            Lt => (Value::Bool(ord.is_lt()), Type::Bool),
+            Gt => (Value::Bool(ord.is_gt()), Type::Bool),
+            Le => (Value::Bool(ord.is_le()), Type::Bool),
+            Ge => (Value::Bool(ord.is_ge()), Type::Bool),
+            Eq => (Value::Bool(ord.is_eq()), Type::Bool),
+            Ne => (Value::Bool(ord.is_ne()), Type::Bool),
             Cmp => {
-                // 3-way comparison.
-                let result = if left < right {
-                    Int::from(-1)
-                } else if left == right {
-                    Int::from(0)
-                } else {
-                    Int::from(1)
+                let val = match ord {
+                    std::cmp::Ordering::Less => -1,
+                    std::cmp::Ordering::Equal => 0,
+                    std::cmp::Ordering::Greater => 1,
                 };
-
-                // FIXME avoid using early return
-                return (Value::Int(result), Type::Int(IntType::I8));
+                (Value::Int(Int::from(val)), Type::Int(IntType::I8))
             }
-        };
-        (Value::Bool(b), Type::Bool)
-    }
-    /// Prepare a value for being compared with a `RelOp`. This involves turning it into an `Int`.
-    fn prepare_for_rel_op(val: Value<M>) -> Int {
-        match val {
-            Value::Int(i) => i,
-            Value::Bool(b) => if b { Int::from(1) } else { Int::from(0) },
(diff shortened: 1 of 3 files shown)


spec/lang/step/operators.md
```diff
@@ -346,14 +346,15 @@ impl<M: Memory> Machine<M> {
...
+    /// Prepare a value for being compared with a `RelOp`. This involves turning it into a pair of `Int`s.
+    /// This pair is compared lexicographically, and is only important for wide pointers.
+    fn prepare_for_rel_op(val: Value<M>) -> (Int, Int) {
         match val {
-            Value::Int(i) => i,
-            Value::Bool(b) => if b { Int::from(1) } else { Int::from(0) },
-            Value::Ptr(p) if p.metadata.is_none() => p.thin_pointer.addr,
-            // TODO(UnsizedTypes): Prepare wide pointer for rel op: `addr * (MAX_META_VALUE+1) + meta_value`
-            Value::Ptr(_) => unimplemented!("comparing pointer with metadata"),
+            Value::Int(i) => (i, Int::ZERO),
```
RalfJung, line 353: I think this is a bit too hacky... we shouldn't compare regular integers via pairs. Also the lexocographic part should be explicit in our sources, rather than relying implicitly on the order of `(Int, Int)`.

Instead maybe rename `eval_rel_op` to `eval_rel_op_int`, and then use that in `eval_bin_op` n a type-dependent way.

Also please match on `Value::Ptr` only once and then have a nested match on `metadata`.

Indeximal:
I have given another proposal, but it is not very consitent right now, because I'd first like to know your opinion on the structuring. Bascially, I use the std::cmp::Ordering struct and use PartialOrd for the thin pointer and the metadata, and then do an explicit lexicographic operation. I don't know whether using the standard library is preferable in this context.
If not, I'd write out the same structure but with direct i8's and without using stdlib functions.
Then I'd refactor the the integer operations as well to make them consitent.

spec/lang/step/operators.md
```diff
@@ -346,28 +346,67 @@ impl<M: Memory> Machine<M> {
...
     fn eval_bin_op(
         &self,
         BinOp::Rel(rel_op): BinOp,
         (left, l_ty): (Value<M>, Type),
         (right, _r_ty): (Value<M>, Type)
     ) -> Result<(Value<M>, Type)> {
-        let left = Self::prepare_for_rel_op(left);
-        let right = Self::prepare_for_rel_op(right);
+        match l_ty {
+            Type::Int(_) | Type::Bool => {
```
RalfJung, line 397: I don't think it's worth having these in a shared match arm; `prepare_for_int_rel_op` isn't carrying its weight any more IMO.

spec/lang/step/operators.md
```diff
@@ -346,28 +346,67 @@ impl<M: Memory> Machine<M> {
...
-            _ => panic!("invalid value for relational operator"),
+            Value::Int(i) => i,
+            Value::Bool(b) => if b { Int::from(1) } else { Int::from(0) },
+            _ => panic!("invalid value for integer relational operator"),
         }
     }
+
+    fn eval_ptr_rel_op(rel: RelOp, left: Pointer<M::Provenance>, right: Pointer<M::Provenance>) -> (Value<M>, Type) {
+        use RelOp::*;
+        let cmp = Self::compare_ptr(left, right);
```
RalfJung, line 360: Using `cmp` as name for something of type `Ordering` is confusing IMO, let's use `ord`.

spec/lang/step/operators.md
```diff
@@ -346,28 +346,67 @@ impl<M: Memory> Machine<M> {
...
+        match rel {
+            Lt => (Value::Bool(cmp.is_lt()), Type::Bool),
+            Gt => (Value::Bool(cmp.is_gt()), Type::Bool),
+            Le => (Value::Bool(cmp.is_le()), Type::Bool),
+            Ge => (Value::Bool(cmp.is_ge()), Type::Bool),
+            Eq => (Value::Bool(cmp.is_eq()), Type::Bool),
+            Ne => (Value::Bool(cmp.is_ne()), Type::Bool),
```
RalfJung, line 367: I like the idea of using `std::cmp::Ordering`; it should be possible to use that to avoid having to match on `rel` twice. Basically, in `eval_bin_op`, we do something like
```rust
let ord = match l_ty { ... };
match rel {
   // Convert `ord` to ` Value::Bool`/`Value::Int` according to `rel`.
}
```
Then `rel` should not be used anywhere else.

tooling/Cargo.lock
RalfJung, line 1: Looks like this needs a rebase, there shouldn't be a lockfile diff in this PR.

RalfJung (commented):
This looks great, thanks. :) Please clean up the history, then we can land this.

Indeximal:
Alright, first did a rebase and then a squash. Should be good now

PR #254, opened 2025-01-03
Fix wide pointer ABI compatibility

Indeximal:
Fixes a bug introduced in #225, where wide and thin pointers are still considered ABI compatible.
Now the `PointerMetaKind` must match.

### DIFF
--- a/tooling/minitest/src/tests/wide_ptr.rs
+++ b/tooling/minitest/src/tests/wide_ptr.rs
@@ -15,7 +15,8 @@ fn ub_wide_thin_abi_incompatible() {
         let mut f = p.declare_function();
         let x = f.declare_local::<u32>();
         f.storage_live(x);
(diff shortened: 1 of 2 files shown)


tooling/minitest/src/tests/wide_ptr.rs
```diff
@@ -1,5 +1,30 @@
...
+        let _arg = f.declare_arg::<*const [u32]>();
+        f.exit();
+        p.finish_function(f)
+    };
+
+    let main = {
+        let mut f = p.declare_function();
+        let x = f.declare_local::<u32>();
+        f.storage_live(x);
+        // UB: &u32 and &[u32] are different size and thus certainly not abi compatible.
```
RalfJung, line 18:
```suggestion
        // UB: `*const u32` and `*const [u32]` are different size and thus certainly
        //  not ABI compatible.
```

RalfJung:
Good catch, thanks!

PR #255, opened 2025-01-03
Finalize unsized types implementation

Indeximal:
I might also split this into multiple PRs if that is easier. Contains a bunch of more or less standalone commits at the moment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ However, without LaTeX this is a pain, and it also involves a lot of jargon whic
 That's why the MiniRust spec is written as an *interpreter*, so the spec itself is code.
 That begs the question, which language do we write that code in?
 We are using a kind of "Rust-like pseudo-code" called *specr lang*:
-imagine Rust without all the restrictions about sizendess and pointer indirections for recursive types (we could implicitly insert `Arc` where needed).
+imagine Rust without all the restrictions about sizedness and pointer indirections for recursive types (we could implicitly insert `Arc` where needed).
 We use generic type names like `List`, `Map`, `Set` rather than concrete implementations like `Vec`, `HashMap`, `HashSet`, since the implementation details do not matter.
 We also assume some "obvious" language extensions -- basically, it should always be clear what is meant to anyone with some Rust experience, even if this is not actually legal Rust.
 
@@ -83,8 +83,7 @@ So just to be clear, there are *two* Rust dialects at play here:
 ## Status
 
 MiniRust is extremely incomplete!
-Many features are entirely missing (e.g. floats, unsized types) or only partially sketched (enum layouts).
-Many types have missing representation relations.
+Some features are entirely missing (e.g. floats).
 There are lots of TODOs.
 The language syntax is also missing many of the Rust operators and casts.
 I hope to slowly chip away at all this over time.
@@ -175,8 +174,8 @@ The intention is that MiniRust has no more UB than that, but it *does* have less
 
 - It is *not* UB to dereference a null, unaligned, or dangling raw pointer. In other words, `addr_of!(*ptr)` is always defined.
   However, if the `*ptr` place expression is being offset, that still needs to happen in-bounds, and actual loads/stores need to be sufficiently aligned.
-- It is *not* always UB to create a reference or `Box` to an invalid value, or one that is dangling.
-  However, it *is* UB to create a reference or `Box` to an *uninhabited type*, or one that is null or unaligned.
+- It is *not* always UB to create a reference or `Box` to an invalid value.
+  However, it *is* UB to create a reference or `Box` to an *uninhabited type*, or one that is dangling or unaligned.
   Moreover, when evaluating an `&[mut]` value expression, dangling references are UB.
   (Dangling references nested in fields will likely also become UB when the aliasing model is added.)
 
--- a/spec/lang/representation.md
+++ b/spec/lang/representation.md
@@ -170,9 +170,9 @@ impl PointerMetaKind {
     /// but this may return ill-formed metadata (as defined by `Machine::check_ptr_metadata`), thus needs to be checked.
     fn decode_value<M: Memory>(self, value: Value<M>) -> Option<PointerMeta<M::Provenance>> {
         match (self, value) {
+            (PointerMetaKind::None, Value::Tuple(fields)) if fields.is_empty() => None,
             (PointerMetaKind::ElementCount, Value::Int(count)) => Some(PointerMeta::ElementCount(count)),
             (PointerMetaKind::VTablePointer(_), Value::Ptr(ptr)) if ptr.metadata.is_none() => Some(PointerMeta::VTablePointer(ptr.thin_pointer)),
-            (PointerMetaKind::None, Value::Tuple(fields)) if fields.is_empty() => None,
             _ => panic!("PointerMeta::decode_value called with invalid value"),
         }
     }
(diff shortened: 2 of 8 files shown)


spec/lang/representation.md
```diff
@@ -131,25 +131,28 @@ fn encode_ptr<M: Memory>(ptr: ThinPointer<M::Provenance>) -> List<AbstractByte<M
 }
 
 impl PointerMetaKind {
-    /// Returns the representable type of the metadata if there is any.
-    pub fn ty<T: Target>(self) -> Option<Type> {
+    /// Returns the representable type of the metadata.
```
RalfJung, line 134: What does "representable" mean here?

spec/lang/representation.md
```diff
@@ -131,25 +131,28 @@ fn encode_ptr<M: Memory>(ptr: ThinPointer<M::Provenance>) -> List<AbstractByte<M
...
+        if self.meta_kind() == PointerMetaKind::None {
+            return None;
+        }
+        let meta_ty = self.meta_kind().ty::<T>();
         let thin_pointer_field = (Offset::ZERO, Type::Ptr(PtrType::Raw { meta_kind: PointerMetaKind::None }));
         let metadata_field = (T::PTR_SIZE, meta_ty);
         ret(Type::Tuple {
             fields: list![thin_pointer_field, metadata_field],
-            size: Int::from(2) * T::PTR_SIZE,
+            size: T::PTR_SIZE + meta_ty.layout::<T>().expect_size("metadata is always sized"),
```
RalfJung, line 155: This is still making some subtle assumptions about alignment. So probably we should just assert that the metadata has the same size and alignment as a pointer, and then `2 * PTR_SIZE` makes perfect sense here.

spec/lang/representation.md
```diff
@@ -131,25 +131,28 @@ fn encode_ptr<M: Memory>(ptr: ThinPointer<M::Provenance>) -> List<AbstractByte<M
...
     }
 }
 
 impl PtrType {
     /// Returns a pair type representing this wide pointer or `None` if it is thin.
     pub fn as_wide_pair<T: Target>(self) -> Option<Type> {
-        let meta_ty = self.meta_kind().ty::<T>()?;
+        if self.meta_kind() == PointerMetaKind::None {
+            return None;
+        }
```
RalfJung, line 149: Hm, this is not pretty... where is this function used?

tooling/minitest/src/tests/slice.rs
```diff
@@ -177,6 +177,13 @@ fn index_with_transmuted() {
...
 
+/// Corresponds to
+/// ```rust
+/// let x: [u32; 3] = [42, 43, 44];
+/// let y = &x as &[u32];
+/// let z = *unsafe { y.get_unchecked(1) };
+/// assert!(z == 43);
+/// ```
 #[test]
 fn index_with_constructed() {
```
RalfJung, line 188: I don't think I understand the name of this test...

spec/lang/representation.md
Indeximal: How so? Previously it was that `ty()` returned None for thin pointers, now the as_wide_pair does this, which is used to encode wide pointers as tuples (but not thin pointers, as this would lead to infinite recursion).

Indeximal:
I have rebased this to split the new tests into #260 and fix the conflict.

spec/lang/representation.md
RalfJung: Previously we didn't need an `if` like this, did we? The `?` handled this implicitly which is nicer.

Given all the `.unwrap_or_else(unit_type)` I think the new way is still better. I also can't find a nicer way to organize things in or around `as_wide_pair` here so let's just keep this.

spec/lang/representation.md
```diff
@@ -131,20 +131,25 @@ fn encode_ptr<M: Memory>(ptr: ThinPointer<M::Provenance>) -> List<AbstractByte<M
...
+        let meta_ty = self.meta_kind().ty::<T>();
+        assert_eq!(meta_ty.layout::<T>().expect_size("metadata is always sized"), T::PTR_SIZE, "metadata is assumed to be pointer sized");
+        assert_eq!(meta_ty.layout::<T>().expect_align("metadata is always sized"), T::PTR_ALIGN, "metadata is assumed to be pointer aligned");
```
RalfJung, line 152:
```suggestion
        assert_eq!(meta_ty.layout::<T>().expect_size("metadata is always sized"), T::PTR_SIZE, "metadata is assumed to be pointer-sized");
        assert_eq!(meta_ty.layout::<T>().expect_align("metadata is always sized"), T::PTR_ALIGN, "metadata is assumed to be pointer-aligned");
```

RalfJung (commented):
Just a tiny nit :) Also please clean up the git history a little.

README.md
```diff
@@ -175,8 +175,8 @@ The intention is that MiniRust has no more UB than that, but it *does* have less
 
 - It is *not* UB to dereference a null, unaligned, or dangling raw pointer. In other words, `addr_of!(*ptr)` is always defined.
   However, if the `*ptr` place expression is being offset, that still needs to happen in-bounds, and actual loads/stores need to be sufficiently aligned.
-- It is *not* always UB to create a reference or `Box` to an invalid value, or one that is dangling.
-  However, it *is* UB to create a reference or `Box` to an *uninhabited type*, or one that is null or unaligned.
+- It is *not* always UB to create a reference or `Box` to an invalid value.
+  However, it *is* UB to create a reference or `Box` to an *uninhabited type*, or one that is dangling or unaligned.
   Moreover, when evaluating an `&[mut]` value expression, dangling references are UB.
   (Dangling references nested in fields will likely also become UB when the aliasing model is added.)
```
Indeximal, line 181: We changed this in #244, but I am somewhat unsure about how to update this text.

Indeximal:
Okay, so I think now these are the last planned changes, do look over them and then I'll clean the git history.

README.md
```diff
@@ -83,7 +83,7 @@ So just to be clear, there are *two* Rust dialects at play here:
 ## Status
 
 MiniRust is extremely incomplete!
-Many features are entirely missing (e.g. floats, unsized types) or only partially sketched (enum layouts).
+Many features are entirely missing (e.g. floats) or only partially sketched (enum layouts).
 Many types have missing representation relations.
```
RalfJung, line 87:
```suggestion
Some features are entirely missing (e.g. floats).
```

README.md
```diff
@@ -129,7 +129,7 @@ The obvious answer is that Niko is much better at picking names. ;)
 
 On a more serious note, these projects have very a different scope: MiniRust is *only* about the operational semantics.
 a-mir-formality is a lot more ambitious; as the [inaugural blog post](https://nikomatsakis.github.io/a-mir-formality/blog/2022/05/12/) explains, it aims to also formalize traits, type checking, and borrow checking -- all of which I consider out-of-scope for MiniRust.
-a-mir-formality is machine-readable and written in PLT redex; MiniRust uses pseudo-code that is not currently machine-readable (but I have ideas :).
+a-mir-formality is machine-readable and written in PLT redex; MiniRust uses pseudo-code that is not currently machine-readable (but I have ideas :) ).
```
RalfJung, line 132: This was deliberate, the `:)` also serves as closing parentheses :-)

RalfJung:
There's also a conflict that needs resolving.

RalfJung:
Should be ready for history cleanup then. :)

PR #261, opened 2025-03-16
Fix typo in documentation of LayoutStrategy::Slice

ia0:
The documentation does not match the implementation of `LayoutStrategy::compute_size_and_align()`. I'm assuming the implementation is correct because it matches my understanding of Rust.


RalfJung:
Good catch, thanks!

PR #262, opened 2025-03-24
Add support for unwinding

DominikDammann:
- Add control-flow edges for unwinding
- Add StartUnwind and ResumeUnwind
- Add support for unwinding in the builder


spec/lang/syntax.md
```diff
@@ -360,9 +360,16 @@ pub enum Terminator {
         /// The block to jump to when this call returns.
         /// If `None`, UB will be raised when the function returns.
         next_block: Option<BbName>,
+        /// The block to jump to when this call unwinds.
+        /// If `None`, UB will be raised when the function unwinds.
+        unwind_block: Option<BbName>
     },
     /// Return from the current function.
     Return,
+    /// Starts unwinding, jump to the first cleanup block.
```
RalfJung, line 369:
```suggestion
    /// Starts unwinding, jump to the indicated cleanup block.
```

spec/lang/syntax.md
```diff
@@ -360,9 +360,16 @@ pub enum Terminator {
         /// The block to jump to when this call returns.
         /// If `None`, UB will be raised when the function returns.
         next_block: Option<BbName>,
+        /// The block to jump to when this call unwinds.
+        /// If `None`, UB will be raised when the function unwinds.
```
RalfJung, line 364:
```suggestion
        /// If `None`, UB will be raised when the function unwinds.
        /// This comes with a well-formedness requirement: if the current block is a regular block,
        /// `unwind_block` must be a cleanup block; otherwise, `unwind_block` must be a terminating block.
```

spec/lang/syntax.md
```diff
@@ -472,6 +480,14 @@ pub struct Function {
...
     pub statements: List<Statement>,
     pub terminator: Terminator,
+    pub kind: BbKind,
+}
+
+/// The kind of a basic block in the CFG.
+pub enum BbKind {
+    Regular,
+    Cleanup,
+    Terminate,
```
RalfJung, line 490:
```suggestion
    /// Regular blocks may use `Return` and `StartUnwind` but not `ResumeUnwind`.
    Regular,
    /// Cleanup blocks may use `ResumeUnwind` but not `Return` or `StartUnwind`.
    Cleanup,
    /// `Terminate` blocks may use neither `Return` nor `ResumeUnwind` nor `StartUnwind`.
    Terminate,
```

spec/lang/well-formed.md
```diff
@@ -670,13 +670,15 @@ fn is_atomic_binop(op: IntBinOp) -> bool {
 impl Terminator {
     fn check_wf<T: Target>(
         self,
+        block: BasicBlock,
```
RalfJung, line 673: We only need the block kind, right? So please only pass that. Also add a comment to the function explaining the argument.

spec/lang/well-formed.md
```diff
@@ -670,13 +670,15 @@ fn is_atomic_binop(op: IntBinOp) -> bool {
...
         func: Function,
         prog: Program,
     ) -> Result<()> {
         use Terminator::*;
         match self {
             Goto(block_name) => {
                 ensure_wf(func.blocks.contains_key(block_name), "Terminator::Goto: next block does not exist")?;
+                ensure_wf(block.kind == func.get_block_kind(block_name), "Terminator::Goto: next block has the wrong block kind")?;
```
RalfJung, line 681: You are looking up the block twice here. Instead, please do something like:
```rust
let Some(next_block) = func.blocks.get(block_name) else { throw_ill_formed!("...") };
ensure_wf(next_block.kind == block_kind, "...");
```

This of course applies throughout this function.

Maybe you'll want to define a little helper function, `check_next_block`, locally in this file, to make this less repetetive.

spec/lang/well-formed.md
```diff
@@ -734,9 +739,32 @@ impl Terminator {
 
                 if let Some(next_block) = next_block {
                     ensure_wf(func.blocks.contains_key(next_block), "Terminator::Call: next block does not exist")?;
+                    ensure_wf(block.kind == func.get_block_kind(next_block), "Terminator::Call: next block has the wrong block kind")?;
+                }
+
+                if let Some(unwind_block) = unwind_block {
+                    ensure_wf(func.blocks.contains_key(unwind_block), "Terminator::Call: unwind block does not exist")?;
+                    match block.kind{
```
RalfJung, line 747:
```suggestion
                    match block.kind {
```
Please ensure you follow standard formatting practices.

spec/lang/well-formed.md
```diff
@@ -770,11 +798,15 @@ impl Function {
...
                 statement.check_wf::<T>(self, prog)?;
             }
-            block.terminator.check_wf::<T>(self, prog)?;
+            block.terminator.check_wf::<T>(block, self, prog)?;
         }
 
         ret(())
     }
+
+    fn get_block_kind(self, block_name: BbName) -> BbKind {
```
RalfJung, line 807: This function should probably disappear once you resolved my other comments.

spec/lang/step/intrinsics.md
```diff
@@ -117,6 +117,21 @@ impl<M: Memory> Machine<M> {
 }
 ```
 
+`Abort` stopts the machine immediately.
+
+```rust
+impl<M: Memory> Machine<M> {
+    fn eval_intrinsic(
+        &mut self,
+        IntrinsicOp::Abort: IntrinsicOp,
```
RalfJung, line 126: We should probably remove the `Panic` intrinsic while we are at it. You can just replace all its existing uses with `Abort`.

spec/lang/machine.md
```diff
@@ -62,6 +62,10 @@ struct StackFrame<M: Memory> {
     /// Expresses what happens after the callee (this function) returns.
     return_action: ReturnAction<M>,
 
+    /// The basic block to jump to when the callee calls ResumeUnwind.
+    /// If `None`, UB will be raised when the callee calls ResumeUnwind.
```
RalfJung, line 66: What is the "callee" here? I don't understand this comment.

Shouldn't this be a new field in `ReturnAction::ReturnToCaller`?

spec/lang/machine.md
```diff
@@ -302,6 +306,7 @@ impl<M: Memory> Machine<M> {
         let init_frame = self.create_frame(
             func,
             ReturnAction::BottomOfStack,
+            None,
```
RalfJung, line 309:
```suggestion
            None, // No unwind block
```

spec/lang/step/terminators.md
```diff
@@ -325,7 +328,7 @@ impl<M: Memory> Machine<M> {
             ReturnAction::BottomOfStack => {
                 // Only the bottom frame in a stack has no caller.
                 // Therefore the thread must terminate now.
-                self.terminate_active_thread()?;
+                self.terminate_active_thread("return")?;
```
RalfJung, line 331: I'm not a fan of passing around `&str` arguments.

I think it's fine here to just have the error say "the start function must not return nor unwind".

spec/lang/step/terminators.md
```diff
@@ -350,6 +353,51 @@ impl<M: Memory> Machine<M> {
 Note that the caller has no guarantee at all about the value that it finds in its return place.
 It should probably do a `Validate` as the next step to encode that it would be UB for the callee to return an invalid value.
 
+## Start Unwind
+
+```rust
```
RalfJung, line 358:
```suggestion
## Starting unwinding

Initiating unwinding is as simple as jumping to a cleanup block.
This will then eventually invoke `ResumeUnwind` and thus propagate upwards through the stack.

```rust
```

spec/lang/step/terminators.md
```diff
@@ -350,6 +353,51 @@ impl<M: Memory> Machine<M> {
...
+impl<M: Memory> Machine<M> {
+    fn eval_terminator(&mut self, Terminator::StartUnwind(block_name): Terminator) -> NdResult {
+        self.jump_to_block(block_name)?;
+        ret(())
+    }
+}
+```
+
+## Resume Unwind
```
RalfJung, line 367:
```suggestion
## Resuming unwinding in the caller
```

spec/lang/step/terminators.md
```diff
@@ -350,6 +353,51 @@ impl<M: Memory> Machine<M> {
...
+            frame.storage_dead(&mut self.mem, local)?;
+        }
+
+        // Inform the memory model that this call has ended.
+        self.mem.end_call(frame.extra)?;
+
+        // Terminate current thread, if the bottom of the stack is reached.
+        if self.active_thread().stack.len() == 0{
+                self.terminate_active_thread("resume")?;
+        }
```
RalfJung, line 387: Having to check the stack lengths here is suspect; this should look more like the logic in `Return`.

spec/lang/step/terminators.md
```diff
@@ -350,6 +353,51 @@ impl<M: Memory> Machine<M> {
...
+        // Jump to the unwind_block of the caller. 
+        else{
+            if let Some(unwind_block) = frame.unwind_block {
+                    self.jump_to_block(unwind_block)?;
+                } else {
+                    throw_ub!("unwinding from a function where caller did not specify unwind_block");
```
RalfJung, line 393:
```suggestion
                    throw_ub!("unwinding from a function where caller did not specify an unwind block");
```

RalfJung (commented):
Thanks!

I have looked only at `spec` so far; all comments are fairly minor.

tooling/miniutil/src/build/mod.rs
```diff
@@ -261,6 +261,44 @@ impl FunctionBuilder {
...
+    {
+        let mut cur_block = self.cur_block.take();
+        let cleanup_block = self.declare_block();
+        self.set_cur_block(cleanup_block, BbKind::Cleanup);
+        cleanup_builder(self);
+
+        if self.cur_block.is_some() {
+            panic!("The cleanup block is unfinished. The block needs to end with a Terminator.");
+        }
+        self.cur_block = cur_block.take();
```
RalfJung, line 277: Why is there a `take` here?

tooling/miniutil/src/build/mod.rs
```diff
@@ -261,6 +261,44 @@ impl FunctionBuilder {
         self.args.push(name);
         local_by_name(name)
     }
+
+    pub fn cleanup<F>(&mut self, cleanup_builder: F) -> BbName
+    where
+        F: Fn(&mut Self),
+    {
+        let mut cur_block = self.cur_block.take();
```
RalfJung, line 269: Maybe call this `old_cur_block` or so.

tooling/miniutil/src/build/mod.rs
```diff
@@ -261,6 +261,44 @@ impl FunctionBuilder {
         self.args.push(name);
         local_by_name(name)
     }
+
+    pub fn cleanup<F>(&mut self, cleanup_builder: F) -> BbName
```
RalfJung, line 265: Please add a doc comment.

tooling/miniutil/src/build/mod.rs
```diff
@@ -261,6 +261,44 @@ impl FunctionBuilder {
...
+        cleanup_block
+    }
+
+    pub fn cleanup_resume(&mut self) -> BbName {
+        self.cleanup(|f| {
+            f.resume_unwind();
+        })
+    }
+
+    pub fn terminate<F>(&mut self, terminat_builer: F) -> BbName
```
RalfJung, line 287: `b.terminate(...)` looks confusing. Maybe call it `terminating_block`?

And then call the other one `cleanup_block` for consistency.

Also please add a doc comment here.

tooling/miniutil/src/build/terminator.rs
```diff
@@ -13,117 +13,167 @@ impl FunctionBuilder {
...
         self.finish_block(Terminator::Return);
     }
 
+    pub fn resume_unwind(&mut self) {
+        self.finish_block(Terminator::ResumeUnwind);
+    }
+
     pub fn panic(&mut self) {
         self.finish_block(panic());
     }
```
RalfJung, line 38: This will probably also disappear when the corresponding terminator disappears.

tooling/miniutil/src/build/terminator.rs
```diff
@@ -13,117 +13,167 @@ impl FunctionBuilder {
...
 
-    /// Call a function that does not return.
+    // Call terminators
```
RalfJung, line 40:
```suggestion
    // Call terminators

```

tooling/miniutil/src/build/terminator.rs
```diff
@@ -13,117 +13,167 @@ impl FunctionBuilder {
...
+            arguments: args.iter().copied().collect(),
+            ret,
+            next_block: Some(next_block),
+            unwind_block: Some(unwind_block),
         });
-        self.set_cur_block(next_block);
+        self.set_cur_block(next_block, next_block_kind);
+    }
+
+    // terminator with 0 following blocks and 1 clean up block
```
RalfJung, line 103:
```suggestion
    /// terminator with 0 following blocks and 1 clean up block
```
Also please make this a proper sentence.

tooling/miniutil/src/build/terminator.rs
```diff
@@ -13,117 +13,167 @@ impl FunctionBuilder {
...
+    pub fn start_unwind(&mut self, clean_up: BbName) {
+        self.finish_block(start_unwind(clean_up));
     }
 
+    // terminators with 1 following block
```
RalfJung, line 108: Same here re: doc comment

tooling/miniutil/src/build/terminator.rs
```diff
@@ -146,44 +197,50 @@ impl FunctionBuilder {
...
-        self.set_cur_block(next_block)
+        self.set_cur_block(next_block, next_block_kind);
     }
 
     pub fn lock_create(&mut self, ret: PlaceExpr) {
         let next_block = self.declare_block();
+        let next_block_kind = self.cur_block().kind;
         self.finish_block(lock_create(ret, bbname_into_u32(next_block)));
-        self.set_cur_block(next_block)
+        self.set_cur_block(next_block, next_block_kind);
```
RalfJung, line 229: This could be made less repetitive by writing it like
```rust
self.finish_with_next_block(|next_block| lock_acquire(lock_id, bbname_into_u32(next_block)));
```

tooling/minitest/src/tests/builder_api.rs
```diff
@@ -171,3 +171,33 @@ fn no_exit() {
     let p = p.finish_program(f);
     assert_stop::<BasicMem>(p);
 }
+
+#[test]
+fn reach_terminate_block() {
```
RalfJung, line 176: This should be in `unwinding.rs`.

tooling/minitest/src/tests/call.rs
```diff
@@ -126,3 +131,43 @@ fn ret_mismatch() {
     dump_program(p);
     assert_ub::<BasicMem>(p, "call ABI violation: return types are not compatible");
 }
+
+#[test]
```
RalfJung, line 135: Every test should have a (doc) comment explaining what it tests.

tooling/minitest/src/tests/ill_formed.rs
RalfJung, line 1: Let's have a file `blocks.rs` that has all tests for "block does not exist" and "block has wrong kind".

tooling/minitest/src/tests/ill_formed.rs
```diff
@@ -36,3 +36,178 @@ fn type_mismatch() {
...
+    let f0 = function(Ret::No, 0, &[], &[bb0, bb1, bb2]);
+
+    let f1 = function(Ret::No, 0, &[], &[block!(return_())]);
+    let p = program(&[f0, f1]);
+    dump_program(p);
+    assert_ill_formed::<BasicMem>(p, "Terminator::Call: unwind block has the wrong block kind");
+}
+
+#[test]
+fn return_in_cleanup() {
```
RalfJung, line 148: These and the following should be in `unwinding.rs`.

tooling/minitest/src/tests/panic.rs
RalfJung, line 1: There should be a test for `abort` instead.

tooling/minitest/src/tests/unwinding.rs
```diff
@@ -0,0 +1,204 @@
...
+}
+
+#[test]
+fn start_unwind_in_main() {
+    let mut p = ProgramBuilder::new();
+    let mut f = p.declare_function();
+
+    let cleanup = f.cleanup(|f| {
+        f.print(const_int(42));
+        f.exit(); //Call exit() instead of abort, because there is currently no way to access the standard output if the program has aborted.
```
RalfJung, line 26:
```suggestion
        // Call exit() instead of abort, because there is currently no way to access the standard output if the program has aborted.
        f.exit();
```

tooling/minitest/src/tests/unwinding.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    let f = {
+        let mut f = p.declare_function();
+        let arg = f.declare_arg::<i32>();
+        let var = f.declare_local::<i32>();
+        let ret = f.declare_ret::<i32>();
+
+        let cleanup_resume = f.cleanup_resume();
+        let cleanup_print = f.cleanup(|f| {
+            f.print(load(arg));
```
RalfJung, line 47: I think it'd be fun to print both at the top of the function and in cleanup, so we can see the nesting.

Maybe make the number smaller than 10 to make the output shorter. ;)

tooling/miniutil/src/build/terminator.rs
DominikDammann: In this file, before my changes, the Terminator was divided into three groups.
 

> // terminators with 0 following blocks

> // terminators with exactly 1 following block

> // terminators with 2 or more following blocks

I intended to create a new such group for `StartUnwind` with this comment. 
It might be better to put `StartUnwind` in the  `// terminators with exactly 1 following block`  group.

Therefore this comment was not meant to be a doc comment. 

So far only two terminators have a doc comment. Should I still add a doc comment to `StartUnwind` and `ResumeUnwind`?

spec/lang/machine.md
DominikDammann: I removed this argument of `create_frame`. It is no longer needed since the unwind block is now part of `StackPopAction`.

DominikDammann:
I have implemented the suggested changes from the previous review. 

Please see also the reply I made for the `/builder/terminator.rs` file.

tooling/miniutil/src/build/terminator.rs
RalfJung: > Therefore this comment was not meant to be a doc comment.

Then there should be an empty line between the comment and the function. Otherwise it looks like it refers only to that function.

But yeah I also think this should be grouped with the "exactly 1 following block" terminators.

spec/lang/step/terminators.md
```diff
@@ -279,11 +277,11 @@ The callee should probably start with a bunch of `Validate` statements to ensure
 
 ```rust
 impl<M: Memory> Machine<M> {
-    fn terminate_active_thread(&mut self, action: &str) -> NdResult {
+    fn terminate_active_thread(&mut self) -> NdResult {
         let active = self.active_thread;
         // The main thread may not terminate, it must call the `Exit` intrinsic.
         if active == 0 {
-            throw_ub!("the start function must not {action}");
+            throw_ub!("the start function must not return nor resume");
```
RalfJung, line 284:
```suggestion
            throw_ub!("the start function must not return nor unwind");
```

spec/lang/step/terminators.md
```diff
@@ -381,17 +382,21 @@ impl<M: Memory> Machine<M> {
...
-        }
-        // Jump to the unwind_block of the caller. 
-        else{
-            if let Some(unwind_block) = frame.unwind_block {
+        // Perform the stack pop action.
+        match frame.stack_pop_action {
+            StackPopAction::BottomOfStack => {
+                // Only the bottom frame in a stack has no caller.
+                // Therefore the thread must terminate now.
+                self.terminate_active_thread()?;
```
RalfJung, line 390: It is UB to unwind out of the bottom of the stack.

DominikDammann:
I’ve made the changes. I hope I understood correctly: it’s UB when the function at the bottom of the stack unwinds -regardless of whether it’s the start function or the function at the bottom of the stack in a new thread.

RalfJung:
> I’ve made the changes. I hope I understood correctly: it’s UB when the function at the bottom of the stack unwinds -regardless of whether it’s the start function or the function at the bottom of the stack in a new thread.

Yes, that is correct. Both of these cases should have a test.

DominikDammann:
I added the missing test.

RalfJung (commented):
Oh great, I had a comment sitting here which github never showed to you. Great UI... sorry for this.

RalfJung:
Looks good, thanks. :)

Please squash the commits so we don't have a bunch of "fix this or that".

DominikDammann:
I squashed the commits.

PR #266, opened 2025-05-19
fix miniutil tests and add them to mini test

### LINKED ISSUE #265: `miniutil` is bit-rotting
MiniUtil is currently bit-rotting.

To start with, the test `basic_works` for `mock_write` has been broken since [fall 2023](https://github.com/minirust/minirust/commit/7811cda671d261148d6f7c12791dab90c5308be9), but no one cared to run the tests:

```
$ cargo test
   Compiling miniutil v0.1.0 (/tmp/minirust/minirust/tooling/miniutil)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.20s
     Running unittests src/lib.rs (/tmp/minirust/minirust/tooling/target/debug/deps/miniutil-1bc5d73ba3d15074)

running 1 test
test mock_write::tests::basic_works ... FAILED

failures:

---- mock_write::tests::basic_works stdout ----
thread 'mock_write::tests::basic_works' panicked at miniutil/src/mock_write.rs:57:9:
assertion `left == right` failed
  left: 1
 right: 2
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Even when that test is fixed, other tests fail to compile because the doctests in `src/build` are not being kept up-to-date with the actual macros in there.

The fix would be to add a `cargo test` run to CI?

JoJoDeveloping:
Fix #265

### DIFF
--- a/mini
+++ b/mini
@@ -25,6 +25,7 @@ COMMAND="$1"
 shift
 case "$COMMAND" in
     test)
+        cargo test --manifest-path=tooling/miniutil/Cargo.toml $CARGOFLAGS "$@"
         cargo test --manifest-path=tooling/minitest/Cargo.toml $CARGOFLAGS "$@"
         cargo test --manifest-path=tooling/minimize/Cargo.toml $CARGOFLAGS "$@"
         ;;
--- a/tooling/miniutil/src/build/mod.rs
+++ b/tooling/miniutil/src/build/mod.rs
@@ -3,8 +3,9 @@
 //! Example:
 //!
 //! ```rust
+//! use miniutil::build::*;
 //! // Our main function has one local of type `usize`.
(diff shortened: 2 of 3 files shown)


mini
```diff
@@ -27,6 +27,7 @@ case "$COMMAND" in
     test)
         cargo test --manifest-path=tooling/minitest/Cargo.toml $CARGOFLAGS "$@"
         cargo test --manifest-path=tooling/minimize/Cargo.toml $CARGOFLAGS "$@"
+        cargo test --manifest-path=tooling/miniutil/Cargo.toml $CARGOFLAGS "$@"
```
RalfJung, line 30: Please move this up so it is tested first -- this is the most fundamental of the crates, after all.

JoJoDeveloping: done

PR #273, opened 2025-09-11
fix edge-cases around TB protectors

JoJoDeveloping:
This implements the behavior of https://github.com/rust-lang/miri/pull/4580, and similarly allows deallocation when the protector is entirely on `Cell`s, [similar to miri](https://github.com/rust-lang/miri/blob/14dc656410abc0c9bf018e2e263297b591c58c4f/src/borrow_tracker/tree_borrows/tree.rs#L758).

It also makes sure that there are no protector end accesses to `Cell` bytes, which was also yet missing but [is done in Miri](https://github.com/rust-lang/miri/blob/fc9c59b2e5d0e745d862852df30a20ea0f7699e9/src/borrow_tracker/tree_borrows/perms.rs#L311).


RalfJung:
Looks good, thanks :)

PR #289, opened 2026-04-28
Add support for json-(de)serializing specr programs

JoJoDeveloping:
See https://github.com/minirust/specr/pull/15

This PR adds cool new features to `minimize`:
* `--minimize-dump=json` prints the minimized program to standard output as a JSON, instead of running it. This JSON is mechanically derived from the MiniRust source code and thus interoperable with other clients/consumers of MiniRust programs.
* `--minimize-src=json` makes minimize not invoke `rustc` to turn a Rust program into a MiniRust program, but it instead loads a MiniRust program previously exported using `--minimize-dump=json`.

A debug assertion asserts that all programs we invoke minimize on are successfully dumpable. This should presumably be removed again at some point.

This PR requires that the corresponding `specr` PR is merged first. Until then, it is WIP and CI will fail since the new `specr-transpile` version has not been released.


JoJoDeveloping:
CC @k-vh @N1ark

tooling/minimize/src/main.rs
```diff
@@ -161,23 +164,33 @@ fn main() {
...
     }
-    let (minimize_args, rustc_args) = split_args(all_args);
+    let (minimize_args, mut args_for_rustc) = split_args(user_args);
+    // if set, dumps the program in a pretty-printed format.
     let dump = minimize_args.iter().any(|x| x == "--minimize-dump");
+    // if set, dumps the program as JSON. This JSON is derived from the specr definitions and designed to be reusable.
+    let dump_json = minimize_args.iter().any(|x| x == "--minimize-dump=json");
+    // if set, expects the program in the shape of a JSON dump instead of a rust source file.
+    // We can then entirely skip the Rust Compiler interaction.
+    let source_is_json = minimize_args.iter().any(|x| x == "--minimize-src=json");
```
RalfJung: I think it makes more sense to make this a separate little program (`minirun`?) than to put it into minimize.

JoJoDeveloping: Should the general pattern then e.g. be `minimize rustfile.rs | minirun --minirun-tree-borrows`? Or should `minimize` have a "run immediately" mode that avoids round-tripping?

RalfJung: I think we can keep the run-immediate mode for now, yeah. It's not a lot of code.

JoJoDeveloping:
`@ralfbot` ready

tooling/minimize/src/main.rs
```diff
@@ -37,6 +37,7 @@ mod rs {
         InterpCx<'tcx, rustc_const_eval::const_eval::CompileTimeMachine<'tcx>>;
 }
 
+use miniutil::json;
```
RalfJung: Please find a reasonable place to put this import. This here looks random.

tooling/minimize/src/main.rs
```diff
@@ -147,10 +138,8 @@ pub fn be_rustc(mut args: Vec<String>) {
 }
 
 fn main() {
-    // Compute the rustc flags we will use. Start by adding our default flags before the
-    // user-defined ones.
-    let mut all_args: Vec<String> = env::args().collect();
-    all_args.splice(1..1, DEFAULT_ARGS.iter().map(ToString::to_string));
+    let mut rustc_additional_args: Vec<_> = DEFAULT_ARGS.iter().map(ToString::to_string).collect();
+    let mut user_args: Vec<String> = env::args().collect();
```
RalfJung: Why did you refactor this code? The old logic maybe wasn't great, but the new one is still a total mess and if you change it that means I have to review it all over again.

Please leave the logic as unchanged as you can. Or if you rewrite it, then make sure the result is clean and easy to review.

tooling/minimize/src/main.rs
```diff
@@ -161,60 +150,41 @@ fn main() {
...
+        return be_rustc(user_args);
     }
-    let (minimize_args, rustc_args) = split_args(all_args);
-    let dump = minimize_args.iter().any(|x| x == "--minimize-dump");
-
-    get_mini(rustc_args, |_tcx, prog| {
-        if dump {
+    let (minimize_args, mut args_for_rustc) = split_args(user_args);
+    // if set, dumps the program in a pretty-printed format.
+    let dump_readable = minimize_args.iter().any(|x| x == "--minimize-dump=human-readable");
```
RalfJung: What is going on here? The PR description says that there's also `--minimize-dump=json` but I see no evidence of that in the code.

Also, this is too messy of an arg parser even by my standards. If you want to have a flag with a value, you should extract the flag's value and match on it, not search for `--flag=value`.

tooling/minimize/tests/ui.rs
```diff
@@ -16,6 +17,7 @@ enum Mode {
 fn cfg(path: &str, mode: Mode) -> Config {
     let mut program = CommandBuilder::rustc();
     program.program = PathBuf::from(env!("CARGO_BIN_EXE_minimize"));
+    program.args.push(OsString::from("--minimize-and-run"));
```
RalfJung: That should be the default mode for minimize, it should not require any flag.

tooling/minirun/src/main.rs
```diff
@@ -0,0 +1,23 @@
+use std::{env, ffi::OsString, fs::File, path::PathBuf};
+
+use miniutil::{json, run::RunConfig};
+
+fn main() {
+    let user_args: Vec<String> = env::args().skip(1).collect();
+    let config = RunConfig::parse_args(&user_args);
+    let mut source: Option<PathBuf> = None;
+    for arg in user_args {
+        if !arg.starts_with("--minimize-") {
```
RalfJung: Why does `minirun` take any "minimize" arguments...? We use that prefix for minimize because that also takes arbitrary rustc flags. `minirun` doesn't have to be like that.

tooling/miniutil/src/json.rs
```diff
@@ -0,0 +1,34 @@
+use std::io::{BufReader, Read};
+
+use minirust_rs::lang::Program;
+
+/// Print a program to stdout, as JSON
+pub fn dump_program_json(prog: Program) {
+    serde_json::to_writer(std::io::stdout(), &prog).expect("serialization to work")
```
RalfJung: This panic message will look quite confusing when it actually triggers.

tooling/miniutil/src/run.rs
```diff
@@ -49,3 +49,47 @@ fn run<M: Memory>(
...
+        TerminationInfo::Abort => show_error!("program aborted"),
+        TerminationInfo::Ub(err) => show_error!("UB: {}", err.get_internal()),
+        TerminationInfo::Deadlock => show_error!("program dead-locked"),
+        TerminationInfo::MemoryLeak => show_error!("program leaked memory"),
+    }
+}
+
+impl RunConfig {
+    pub fn parse_args(args: &[String]) -> RunConfig {
+        RunConfig { tree_borrows: args.iter().any(|x| x == "--minimize-tree-borrows") }
```
RalfJung: Please no arg parsing in `miniutil`.

tooling/miniutil/src/run.rs
```diff
@@ -49,3 +49,47 @@ fn run<M: Memory>(
...
+macro_rules! show_error {
+    ($($tt:tt)*) => {$crate::run::show_error_fn(&format_args!($($tt)*)) };
+}
+
+fn run_program_with_config(prog: Program, c: RunConfig) -> TerminationInfo {
+    if c.tree_borrows { run_program::<TreeBorrowMem>(prog) } else { run_program::<BasicMem>(prog) }
+}
+
+/// Runs the program. If the program fails to run, we print an error and exit with code 101.
+pub fn run_program_interactively(p: Program, c: RunConfig) {
```
RalfJung: What is "interactive" about this...?

tooling/miniutil/src/run.rs
```diff
@@ -49,3 +49,47 @@ fn run<M: Memory>(
...
+
+pub fn show_error_fn(msg: &impl std::fmt::Display) -> ! {
+    eprintln!("fatal error: {msg}");
+    std::process::exit(101) // exit code needed to make ui_test happy
+}
+
+#[macro_export]
+macro_rules! show_error {
+    ($($tt:tt)*) => {$crate::run::show_error_fn(&format_args!($($tt)*)) };
+}
```
RalfJung: This has nothing to do with "running" the program, please find a better spot for it.

RalfJung (commented):
I'm afraid I am confused by the majority of this PR. This will need some refactoring/cleanup, or avoid changing existing code as much as it does.

tooling/miniutil/src/run.rs
RalfJung: For just a single flag I don't think it's worth sharing the arg parsing logic.

It's easier to have a simple parser like Miri's for both minimize and miniutil:
```rust
for arg in args {
  if arg == "--flag" { ... }
  else if let Some(val) = arg.strip_prefix("--flag-with-val=") { ... }
  else { /* error */ }
}
```

JoJoDeveloping:
I refactored everything, please have a clean look again unburdened by what has been. I recommend you look through it commit-by-commit.

tooling/miniutil/src/json.rs
```diff
@@ -0,0 +1,21 @@
+use std::{io::{BufReader, Read}, ops::Deref};
+
+use minirust_rs::lang::Program;
+
+pub fn dump_program(prog: &Program) {
+    serde_json::to_writer(std::io::stdout(), &prog).expect("Failed to format program as JSON!")
+}
+
+pub fn does_roundtrip(prog: &Program) -> bool {
```
RalfJung: If this is only used for assertions, let's make this `assert_roundtrip` without a return value.

tooling/miniutil/src/tui.rs
RalfJung, line 1: I think this is more of a CLI than a TUI.

tooling/miniutil/src/tui.rs
```diff
@@ -0,0 +1,77 @@
...
+}
+
+impl MinirustMachineConfig {
+    fn consume_arg(&mut self, arg: &str) {
+        //TODO remove this legacy hack
+        if arg == "--minimize-tree-borrows" {
+            self.tree_borrows = true;
+            return;
+        }
+        let Some(arg) = arg.strip_prefix("--minirun-") else {
```
RalfJung: Maybe let's call this "minirust"?

tooling/minimize/src/main.rs
```diff
@@ -183,44 +175,28 @@ fn main() {
...
-            }
+            tui::run_program_with_flags_and_print_errors(prog, &minimize_args);
         }
     });
 }
 
-/// split arguments into arguments for minimize and rustc
-fn split_args(args: Vec<String>) -> (Vec<String>, Vec<String>) {
+/// split arguments into arguments for minimize, minirun, and rustc
+fn split_args(args: Vec<String>) -> (Vec<String>, Vec<String>, Vec<String>) {
```
RalfJung: Nah, sorry, that's too much.^^

If we refactor this, turn it into a loop that does one pass over the arguments and handles them one-by-one (some get parsed, the rest is put on the lits of args passed to rustc).

tooling/miniutil/src/tui.rs
```diff
@@ -0,0 +1,72 @@
...
+
+fn run_prog(prog: Program, args: &Vec<String>) -> TerminationInfo {
+    parse_minirun_args(args).run_prog(prog)
+}
+
+/// Runs a MiniRust program, while controlling some finer details of the used Machine
+/// using the given `flags`. These must start with `--minirun`.
+/// This function returns if the program terminates cleanly, otherwise it prints a
+/// human-readable error message to stderr and exists the process (!) with error code 101.
+pub fn run_program_with_flags_and_print_errors(prog: Program, flags: &Vec<String>) {
```
RalfJung: Might be cleaner to take a `MinirustMachineConfig` here?

tooling/minimize/src/main.rs
```diff
@@ -175,6 +175,7 @@ fn main() {
                 x => show_error!("Unknown dump format {x}"),
             }
         } else {
+            debug_assert!(json::does_roundtrip(&prog));
```
RalfJung: Given that these are all derived, I am not sure this is worth it...

RalfJung (commented):
Here we go :)

tooling/miniutil/src/cli.rs
```diff
@@ -0,0 +1,68 @@
+// This module provides utilities for the terminal user-interface.
+// That is, it bundles common command line argument parsing and error message rendering.
```
RalfJung: These should be doc comments.

tooling/miniutil/src/cli.rs
```diff
@@ -0,0 +1,68 @@
...
+            return false;
+        };
+        match arg {
+            "tree-borrows" => self.tree_borrows = true,
+            _ => show_error!("Unknown argument --minirun-{arg}!"),
+        }
+        true
+    }
+
+    pub fn run_prog(&self, prog: Program) -> TerminationInfo {
```
RalfJung, line 42: Maybe add a doc comment now that this is public API?

tooling/miniutil/src/cli.rs
```diff
@@ -0,0 +1,68 @@
...
+            TerminationInfo::MachineStop => { /* silent exit. */ }
+            TerminationInfo::Abort => show_error!("program aborted"),
+            TerminationInfo::Ub(err) => show_error!("UB: {}", err.get_internal()),
+            TerminationInfo::Deadlock => show_error!("program dead-locked"),
+            TerminationInfo::MemoryLeak => show_error!("program leaked memory"),
+        }
+    }
+}
+
+impl Default for MinirustMachineConfig {
```
RalfJung: Please put this next to the struct definition.

tooling/miniutil/src/cli.rs
```diff
@@ -41,7 +41,7 @@ impl MinirustMachineConfig {
     }
 
     /// Runs a MiniRust program, while controlling some finer details of the used Machine
-    /// using the given `flags`. These must start with `--minirun`.
+    /// using the given `flags`. These must start with `--minirust`.
```
RalfJung: Which `flags`?

tooling/miniutil/src/cli.rs
```diff
@@ -0,0 +1,68 @@
...
+            return false;
+        };
+        match arg {
+            "tree-borrows" => self.tree_borrows = true,
+            _ => show_error!("Unknown argument --minirust-{arg}!"),
+        }
+        true
+    }
+
+    /// Runs the program using [run_program]. The memory model/machine is constructed according to this config.
```
RalfJung:
```suggestion
    /// Runs the program using [`run_program`]. The memory model/machine is constructed according to this config.
```

tooling/miniutil/src/cli.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub fn run_prog(&self, prog: Program) -> TerminationInfo {
+        if self.tree_borrows {
+            run_program::<TreeBorrowMem>(prog)
+        } else {
+            run_program::<BasicMem>(prog)
+        }
+    }
+
+    /// Runs a MiniRust program, see [run_prog].
```
RalfJung:
```suggestion
    /// Runs a MiniRust program, see [`Self::run_prog`].
```

RalfJung (commented):
Please also squash the commits.

PR #290, opened 2026-05-21
tree borrows: implicit writes

quiode:
This PR adds support for implicit writes in the tree borrows memory model. It mirrors the changes done in Miri here: https://github.com/rust-lang/miri/pull/4947.

Implicit writes are disabled by default, but can be enabled using a new argument: `--minirust-tree-borrows-implicit-writes`, same as in Miri.
For this some functions had to be changed to enable the passing of parameters.
Most tests for implicit writes were also copied from Miri and modified to work for minirust.

### DIFF
--- a/spec/mem/interface.md
+++ b/spec/mem/interface.md
@@ -51,7 +51,9 @@ impl<Provenance> AbstractByte<Provenance> {
 
 ## Memory interface
 
-The MiniRust memory interface is described by the following (not-yet-complete) trait definition:
+The MiniRust memory interface is described by the following (not-yet-complete) trait definition.
+A memory model can be configured through parameters that toggle specific or experimental behavior, such as enabling implicit writes in the Tree Borrows memory model.
+These parameters are bundled into the `Params` associated type and passed to the `new` function at construction time:
 
 ```rust
 /// The "kind" of an allocation is used to distinguish, for instance, stack from heap memory.
@@ -84,7 +86,11 @@ pub trait Memory {
     /// Extra information for each stack frame.
     type FrameExtra;
 
-    fn new() -> Self;
+    /// Parameters controlling memory model behavior, passed at construction time.
+    type Params: Default;
+
+    /// Create a new instance of the memory model with the given parameters.
+    fn new(params: Self::Params) -> Self;
 
     /// Create a new allocation.
     /// The initial contents of the allocation are `AbstractByte::Uninit`.
--- a/spec/mem/tree_borrows/memory.md
+++ b/spec/mem/tree_borrows/memory.md
@@ -48,7 +48,17 @@ type TreeBorrowsProvenance = (AllocId, Path);
 The memory itself largely reuses the basic memory infrastructure, with the tree as extra state.
 
 ```rust
+/// Global parameters controlling Tree Borrows behavior.
+#[derive(Default)]
+pub struct TreeBorrowsParams {
+    /// Whether implicit writes are enabled or not.
+    pub implicit_writes: bool,
+}
+```
+
+```rust
 pub struct TreeBorrowsMemory<T: Target> {
+    params: TreeBorrowsParams,
     mem: BasicMemory<T, Path, TreeBorrowsAllocationExtra>,
 }
 
@@ -113,7 +123,7 @@ impl<T: Target> TreeBorrowsMemory<T> {
             // Add the new node to the tree
             let child_path = allocation.extra.root.add_node(parent_path, child_node);
 
-            // Perform a read access on all bytes inside the pointee whose permission requires such an access.
+            // Perform a read or write access on all bytes inside the pointee whose permission requires such an access.
             for (idx, perm) in settings.inside.iter().enumerate() {
                 let idx = Int::from(idx); // FIXME: we need a version of `enumerate` that yields `Int`s.
                 let idx = Size::from_bytes(idx).unwrap();
@@ -173,9 +183,10 @@ impl<T: Target> Memory for TreeBorrowsMemory<T> {
     type Provenance = TreeBorrowsProvenance;
     type FrameExtra = TreeBorrowsFrameExtra;
     type T = T;
+    type Params = TreeBorrowsParams;
 
-    fn new() -> Self {
-        Self { mem: BasicMemory::new() }
+    fn new(params: TreeBorrowsParams) -> Self {
+        Self { mem: BasicMemory::new(), params }
     }
 
     fn allocate(&mut self, kind: AllocationKind, size: Size, align: Align) -> NdResult<ThinPointer<Self::Provenance>>  {
@@ -233,7 +244,7 @@ impl<T: Target> Memory for TreeBorrowsMemory<T> {
         fn_entry: bool,
         vtable_lookup: impl Fn(ThinPointer<Self::Provenance>) -> crate::lang::VTable + 'static,
     ) -> Result<Pointer<Self::Provenance>> {
-        ret(if let Some(perms) = ReborrowSettings::new(ptr, ptr_type, fn_entry, vtable_lookup) {
+        ret(if let Some(perms) = ReborrowSettings::new(ptr, ptr_type, fn_entry, self.params, vtable_lookup) {
             self.reborrow(ptr.thin_pointer, perms, frame_extra)?.widen(ptr.metadata)
         } else {
             ptr
--- a/spec/mem/tree_borrows/reborrow_settings.md
+++ b/spec/mem/tree_borrows/reborrow_settings.md
@@ -93,8 +93,10 @@ impl ReborrowSettings {
         ptr: Pointer<TreeBorrowsProvenance>,
         ptr_type: PtrType,
         fn_entry: bool,
+        params: TreeBorrowsParams,
         vtable_lookup: impl Fn(ThinPointer<TreeBorrowsProvenance>) -> crate::lang::VTable + 'static,
     ) -> Option<Self> {
+        // Returns None for `Raw`, `FnPtr` and `VTablePtr`, so `ptr_type` can only be `Ref` and `Box` from here.
         let Some(pointee_info) = ptr_type.safe_pointee() else {
             return None;
         };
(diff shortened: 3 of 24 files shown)


spec/mem/tree_borrows/reborrow_settings.md
```diff
@@ -115,26 +117,55 @@ impl ReborrowSettings {
...
+                PtrType::Ref { mutbl: Mutability::Mutable, .. } =>
+                    if implicit_writes_enabled && inside {
+                        // We cannot use `Unique` for the outside part.
+                        mk_perm(PermissionUnprot::Unique, PermissionProt::Unique)
+                    } else if protected.yes() || frozen {
+                        // We also use this for protected `&mut UnsafeCell` as otherwise adding `noalias` would not be sound.
+                        mk_perm(PermissionUnprot::Reserved, PermissionProt::Reserved { had_local_read: false, had_foreign_read: false })
+                    } else {
+                        // Unprotected mutable references and boxes start in `ReservedIm`, but if they are protected we ignore the `Im`
+                        mk_perm(PermissionUnprot::ReservedIm, PermissionProt::Reserved { had_local_read: false, had_foreign_read: false })
```
JoJoDeveloping:
```suggestion
                    } else {
                        // Unprotected interior-mutable references and boxes start in `ReservedIm`, but if they are protected we ignore the `Im`
                        mk_perm(
                          if frozen { PermissionUnprot::Reserved } else {  PermissionUnprot::ReservedIm },
                          PermissionProt::Reserved { had_local_read: false, had_foreign_read: false })
```

JoJoDeveloping (approved):
LGTM. Should we merge this now and add the handling of `#[rustc_no_writable]` later or in this PR (or not at all)?

quiode:
> LGTM. Should we merge this now and add the handling of `#[rustc_no_writable]` later or in this PR (or not at all)?

Regarding the attribute: Correct me if I'm wrong, but I think there is currently no way that attributes of functions get passed, so I would also have to implement that, probably first only for `RustcNoWritable` with `bitflags` or something passed down to `ReborrowSettings::new`.
I think the project is also still on `nightly-2026-04-01`, so the version would have to be bumped for my attribute to appear (it was merged on April 17th). I tried to to that myself but there are some changes in nightly that make it non-trivial for me.

JoJoDeveloping:
Bumping `rustc` is no big hurdle. But indeed it seems like stable mir does not expose attributes very well. So let us ask back and we'll address this in a later MR, if at all.

spec/mem/tree_borrows/memory.md
```diff
@@ -47,9 +47,19 @@ type TreeBorrowsProvenance = (AllocId, Path);
...
+pub struct TreeBorrowsParams {
+    /// Whether implicit writes are enabled or not.
+    pub implicit_writes: bool,
+}
+```
+
 ```rust
 pub struct TreeBorrowsMemory<T: Target> {
     mem: BasicMemory<T, Path, TreeBorrowsAllocationExtra>,
+    params: TreeBorrowsParams,
```
RalfJung:
```suggestion
    params: TreeBorrowsParams,
    mem: BasicMemory<T, Path, TreeBorrowsAllocationExtra>,
```

spec/mem/interface.md
```diff
@@ -84,7 +84,10 @@ pub trait Memory {
     /// Extra information for each stack frame.
     type FrameExtra;
 
-    fn new() -> Self;
+    /// Parameters controlling memory model behavior, passed at construction time.
+    type Params: Default;
+
+    fn new(params: Self::Params) -> Self;
```
RalfJung, line 93: This function should also have a doc comment.

spec/mem/interface.md
```diff
@@ -84,7 +84,10 @@ pub trait Memory {
     /// Extra information for each stack frame.
     type FrameExtra;
 
-    fn new() -> Self;
+    /// Parameters controlling memory model behavior, passed at construction time.
+    type Params: Default;
```
RalfJung, line 90: The text above (outside the code block) should explain this `Params` business.

spec/mem/tree_borrows/reborrow_settings.md
```diff
@@ -115,26 +117,43 @@ impl ReborrowSettings {
...
+                PtrType::Ref { mutbl: Mutability::Immutable, .. } =>
+                    if frozen {
+                        mk_perm(PermissionUnprot::Frozen, PermissionProt::Frozen { had_local_read: false })
+                    } else {
+                        mk_perm(PermissionUnprot::Cell, PermissionProt::Cell)
+                    },
+                // Mutable references and Boxes (Boxes are treated the same as mutable references)
+                PtrType::Ref { mutbl: Mutability::Mutable, .. } | PtrType::Box { .. } =>
+                    if implicit_writes_enabled && inside {
+                        // We cannot use `Unique` for the outside part.
```
RalfJung:
```suggestion
                        // The implicit write only happens on the inside.
```

spec/mem/tree_borrows/reborrow_settings.md
```diff
@@ -115,26 +117,43 @@ impl ReborrowSettings {
...
+                        )
+                    },
+                // Cannot occur: `safe_pointee()` returns `None` for these variants.
+                PtrType::Raw { .. } | PtrType::FnPtr | PtrType::VTablePtr(_) => unreachable!("safe_pointee() returns None for Raw, FnPtr, and VTablePtr; should have returned early on function entry"),
+            }
         };
+
         let inside = pointee_info.unsafe_cells.freeze_mask(pointee_info.layout, ptr.metadata, vtable_lookup).map(|freeze|
-            if freeze { no_cell_perm } else { cell_perm }
+            perm(true, freeze)
```
RalfJung:
```suggestion
            perm(/* inside */ true, freeze)
```
same below

spec/mem/tree_borrows/state_machine.md
```diff
@@ -77,3 +70,19 @@ impl Permission {
     }
 }
 ```
+
+When a new node is created, it causes an implicit access, which is either a Read or a Write access.
+This is so the optimizer can insert reads / writes as soon as a reference is created.
+Note that the implicit operation is not universal, and depends on the permission.
```
RalfJung:
```suggestion
When a new node is created, it may cause an implicit (read or write) access.
This is so the optimizer can insert reads / writes as soon as a reference is created.
The presence and type of the access depends on the permission.
```

spec/mem/tree_borrows/reborrow_settings.md
```diff
@@ -115,26 +117,43 @@ impl ReborrowSettings {
...
+                    if frozen {
+                        mk_perm(PermissionUnprot::Frozen, PermissionProt::Frozen { had_local_read: false })
+                    } else {
+                        mk_perm(PermissionUnprot::Cell, PermissionProt::Cell)
+                    },
+                // Mutable references and Boxes (Boxes are treated the same as mutable references)
+                PtrType::Ref { mutbl: Mutability::Mutable, .. } | PtrType::Box { .. } =>
+                    if implicit_writes_enabled && inside {
+                        // We cannot use `Unique` for the outside part.
+                        mk_perm(PermissionUnprot::Unique, PermissionProt::Unique)
```
RalfJung: The "unprot" case here is dead code. Please inline mk_perm, and add an assertion that we're in a protected case.

spec/mem/tree_borrows/state_machine_unprotected.md
```diff
@@ -83,23 +83,14 @@ impl PermissionUnprot {
...
-When a new node is created, it causes an implicit access, usually an implicit read.
-This is so the optimizer can insert reads as soon as a reference is created.
-Note that the implicit read is not universal, and depends on the permission.
-The details are defined by the `init_access` function.
 
-```rust
-impl PermissionUnprot {
     fn init_access(self) -> Option<AccessKind> {
         // Everything except for `Cell` gets an initial read access.
+        // Write's are not possible for unprotected references.
```
RalfJung:
```suggestion
        // We never perform implicit writes for unprotected references.
```

tooling/miniutil/src/run.rs
```diff
@@ -1,25 +1,31 @@
...
-/// Run the program and return stdout as a `Vec<String>`  or a termination info
-/// if it did not terminate correctly. Stderr is just forwarded to the host.
+/// Run the program and return stdout as a `Vec<String>`, using default memory params,
+/// or a termination info if it did not terminate correctly. Stderr is just forwarded to the host.
 pub fn get_stdout<M: Memory>(prog: Program) -> Result<Vec<String>, TerminationInfo> {
     let out = MockWrite::new();
     let err = std::io::stderr();
 
-    let res = run::<M>(prog, out.clone(), err);
+    let res = run::<M>(prog, Default::default(), out.clone(), err);
```
RalfJung:
```suggestion
    let res = run::<M>(prog, M::Params::default(), out.clone(), err);
```
or so?

tooling/minimize/tests/pass/tree_borrows/no_implicit_writes.rs
```diff
@@ -0,0 +1,59 @@
...
+// Shows that without implicit writes, `ptr_write.rs`, `ptr_write_unsafe_cell.rs`,
+// `ptr_write_box.rs`, `as_mut_ptr.rs` would pass.
+
+use std::cell::UnsafeCell;
+
+fn main() {
+    normal();
+    unsafe_cell();
+    box_test();
+    as_mut_ptr_test()
```
RalfJung, line 13: So does every one of these 4 correspond to one of the implicit-write fail tests? If so, please add comments indicating the correspodence, since the function names don't make it clear.

RalfJung (commented):
Thanks! Looks good overall, just a bunch of nits.

quiode:
@RalfJung all comments should be resolved

tooling/minimize/tests/ub/tree_borrows/implicit_writes/as_mut_ptr.rs
```diff
@@ -0,0 +1,19 @@
+//@ compile-flags: --minirust-tree-borrows --minirust-tree-borrows-implicit-writes
+
+// Copied from the Miri test suite and modified for MiniRust.
+// This code no longer works using implicit writes in tree borrows. This code tests that.
```
RalfJung:
```suggestion
// This code no longer works using implicit writes in tree borrows.
```

tooling/minimize/tests/pass/tree_borrows/no_implicit_writes.rs
```diff
@@ -0,0 +1,59 @@
...
+// Shows that without implicit writes, `ptr_write.rs`, `ptr_write_unsafe_cell.rs`,
+// `ptr_write_box.rs`, `as_mut_ptr.rs` would pass.
+
+use std::cell::UnsafeCell;
+
+fn main() {
+    normal(); // passing version of: tooling/minimize/tests/ub/tree_borrows/implicit_writes/ptr_write.rs
+    unsafe_cell(); // passing version of: tooling/minimize/tests/ub/tree_borrows/implicit_writes/ptr_write_unsafe_cell.rs
+    box_test(); // passing version of: tooling/minimize/tests/ub/tree_borrows/implicit_writes/ptr_write_box.rs
+    as_mut_ptr_test() // passing version of: tooling/minimize/tests/ub/tree_borrows/implicit_writes/as_mut_ptr.rs
```
RalfJung: Please move these comments to the relevant functions.

RalfJung (commented):
With some minor final nits, LGTM. Please also squash the commits.

quiode:
@RalfJung done

RalfJung:
(In the future, please avoid force-pushing, as github then makes it impossible to actually see what changed since my last review. Sadly we have to work around the limitations of github manually here.)

RalfJung:
Thanks!

quiode:
> (In the future, please avoid force-pushing, as github then makes it impossible to actually see what changed since my last review. Sadly we have to work around the limitations of github manually here.)

I don't quite understand. How should I then do the squashing?

RalfJung:
Oh right I asked you to squash. But when you do, please use `git rebase --keep-base` so that the PR does not change the master commit it is based on.

PR #292, opened 2026-05-26
Tree borrows: niche claims proposal

digama0:
Based on the proposal: https://hackmd.io/2__yv8jLR1yiMhnxHvU4Ig
See also https://rust-lang.zulipchat.com/#narrow/channel/136281-t-opsem/topic/Proposal.3A.20Niche-preserving.20Tree.20Borrows/with/597882080.

Note: This PR was prepared with assistance from Claude, so while I approve of all of the content, it might be a bit overly verbose in places. AFAIK it didn't state any falsehoods though.

### Deviations from the proposal

- **Phantom reads start at `T_n`, not from the tree root.** The proposal's §3.3 phrases the phantom read as a "Local Read through `T_n`" walked from the root; this PR issues it as a `Node::access` call rooted at `T_n` itself, so it propagates only into `T_n`'s subtree. Justified by Lemma 2 + corollary (§6.6): propagation to siblings is provably a no-op, and propagation to ancestors is already settled by `T_n`'s implicit init Read at retag.
- **Per-byte writability gate at retag.** Proposal §3.2 attaches a claim when the niche region is "non-disjoint" from the new tag's pointee. The PR tightens this to "the niche region contains a byte inside the pointee whose `inside` permission is writable," which folds the shared-borrow case (claims silently dropped for `&T`) and the disjoint clause into one check.
- **Niche check fires on every TB-level access**, including reborrow init accesses and `storage_dead`/`deallocate`. The proposal text only talks about user load/store, but it specifies the trigger as "every TB-level Foreign access" — this PR honours that by hooking inside `Node::access`, the single function that handles all TB accesses.
- **`do_niche_check: bool` flag on `Node::access`** is used to break recursion on phantom reads (they pass `false`). Operationally equivalent to a separate state-only access function, but folds the two into one. (This should be viewed with suspicion. Happy to discuss this further.)

### Open questions from the proposal not addressed

- §7.1 — `&Cell<NonZeroU32>` (shared ref to interior-mutable niche bearer): claims are unconditionally dropped for `Mutability::Immutable` borrows.
- §7.6 — concurrency: phantom reads inherit whatever atomicity the surrounding load uses; not specially modelled.
- §7.7 — `!Unpin + niched`: `retag_ptr` early-returns for `&mut !Unpin`, so the niche claim never gets attached. Open in the proposal too.

### Test coverage

13 niche-extension tests in `tooling/minimize/tests/{pass,ub}/tree_borrows/niche_*.rs` covering §4.1–4.3, 4.6–4.9, 4.11, plus dealloc-as-Foreign-Write and descendant-read variants. `Option<NonZeroU32>` is rendered as `Option<bool>` (or `Option<&'static u32>` where a multi-byte niche is needed) since `NonZeroU32 = pattern_type!(u32 is 1..)` doesn't lower in minimize. Very unusual niche layouts with complex discrimination trees unfortunately can't be tested as rustc doesn't support them.


RalfJung:
Thanks! That should hopefully help clarify lingering questions with the proposal.

> Note: This PR was prepared with assistance from Claude

I'll mark it as draft then until a human gave it a good once-over.

digama0:
Just to clarify, I have read and approve all of this PR. Claude is just better at coming up with test cases than me. But of course you should also read it and become comfortable enough with this or any other code you want to merge.

RalfJung:
If it is "overly verbose in places", that sounds like it needs a bit more work before it is ready for review?

RalfJung:
> Very unusual niche layouts with complex discrimination trees unfortunately can't be tested as rustc doesn't support them.

FWIW this could be tested by writing a MiniRust test directly (rather than by writing Rust code that gets compiled to MiniRust).

digama0:
> If it is "overly verbose in places", that sounds like it needs a bit more work before it is ready for review?

Sure, in theory. I already deleted anything I thought of as beyond the pale, but it's always a judgment call and I'm not as familiar with the appropriate level of verbosity for this codebase, especially since it is explicitly didactic. It's slightly more commenting than I usually write, but it's also a setting where that is appropriate, so I trust your judgment. Better to err on the side of explaining everything until you and others are comfortable with the concepts in the proposal.

RalfJung:
Thinking about it a bit more, I think for the files in `spec` I'd like all changes to be written by a human. That part is just too critical to trust code that was not written with a human understanding and mind behind it. Reviewing code written by someone (or something) else is not the same thing as writing it.

digama0:
I'm telling you, I am cosigning every line of code in this PR and can explain any part of it. (It's not a lot of code to begin with.) I do not let AI do thinking for me. I'm not sure what I can do beyond this to satisfy your requirements.

PR #294, opened 2026-05-30
match miri behavior for tree borrows when encountering pinned boxes

### LINKED ISSUE #291: tree borrows: handling of !Unpin Box differs to Miri
While working on some changes, I saw that how minirust handles `!Unpin` `Box` differs to miri. The exact difference is here: https://github.com/rust-lang/miri/blob/55962eb26a4733e5db315da892df18c142d008cf/src/borrow_tracker/tree_borrows/mod.rs#L160 and here: https://github.com/minirust/minirust/blob/2d2a46b159d370a2506369f8d60eac107567b023/spec/mem/tree_borrows/reborrow_settings.md?plain=1#L102

While in Miri, Box to pinning type is a NOP, it's not one in minirust. I could show the difference using the tests here: https://github.com/quiode/rust/commit/2219f786de2d4b5eb2c015091fdc391585f954b1 and https://github.com/quiode/minirust/commit/2830a82c838636d4fedc165a7b7416d2b262cf1a. As sadly minirust support for `Box` is limited (https://github.com/minirust/minirust/issues/202), I had to change some things to make it work, but the effect should be the same, if you change either minirust or miri to have the same behavior, the tests will also have the same output (either both UB or both not), but currently, one has UB and the other doesn't.

I assume the intended way is that minirust behaves like miri, so then to make them behave the same, one could just add `Box` as another case in the `matches!`,

quiode:
This PR fixes https://github.com/minirust/minirust/issues/291.
It makes Tree Borrows in minirust behave the same as miri when encountering a pinned box. I also added a test which previously failed, meaning it had different behavior than miri, but now succeeds and thus matches miri's behavior.

### DIFF
--- a/spec/mem/tree_borrows/reborrow_settings.md
+++ b/spec/mem/tree_borrows/reborrow_settings.md
@@ -98,8 +98,8 @@ impl ReborrowSettings {
         let Some(pointee_info) = ptr_type.safe_pointee() else {
             return None;
         };
-        if matches!(ptr_type, PtrType::Ref { mutbl: Mutability::Mutable, pointee } if !pointee.unpin) {
-            // Mutable reference to pinning type: retagging is a NOP.
+        if matches!(ptr_type, PtrType::Ref { mutbl: Mutability::Mutable, pointee } | PtrType::Box { pointee } if !pointee.unpin) {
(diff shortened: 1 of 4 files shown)


RalfJung:
Thanks. :)

PR #296, opened 2026-07-01
use env bash for NixOS compatibility

quiode:
I modified all script files to use `#!/usr/bin/env bash` instead of `#!/bin/bash`. This has the same behaviour on most systems, but `#!/bin/bash` doesn't work on NixOS, while the new version does.


RalfJung:
NixOS causing trouble and extra work with their non-standard setup as usual.^^

Thanks!

PR #297, opened 2026-07-01
Bump rust

quiode:
This bumps the rust version used to `nightly-2026-05-01`.
This is needed as a future PR will use features that have only been introduced then.
For the version increase to work, the specr version also needed to be updated.


RalfJung:
LGTM, thanks!

